PR #2, opened 2015-08-17
Add support for authentication

DavidWittman:
Adds the `redis.password` flag, which allows for authenticating connections to the Redis server or servers. To simplify the function signatures, a `RedisHost` structure was implemented to contain all information about the connection to Redis.

There weren't any test instructions, so I wasn't entirely sure how to proceed there, but I updated the portions of the test which used the `NewRedisExporter` to use the new `RedisHost` style. If you have any guidance on that I'll be happy to implement some tests for the authentication portions.

### DIFF
--- a/redis_exporter.go
+++ b/redis_exporter.go
@@ -5,6 +5,7 @@ import (
 	//	"fmt"
 	"log"
 	"net/http"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the pull request!
I'm out of the country and away from my computer right now but will look into this as soon as I'm back.

redis_exporter.go
```diff
@@ -227,6 +233,12 @@ func (e *Exporter) scrape(scrapes chan<- scrapeResult) {
 			errorCount++
 			continue
 		}
+		if e.redis.password != "" {
+			if _, err := c.Do("AUTH", e.redis.password); err != nil {
+				log.Printf("redis err: %s", err)
+				continue
```
oliver006: Bump `errorCount` here before continuing?

oliver006:
Sorry, only got time to look at this now.
I like the approach of putting it all in the RedisHost structure, keeps it clean.

Two things I was wondering about
- this allows for only one password for all monitored redis hosts. I don't have a good, clean alternative how to pass several passwords if several hosts are configured. I don't think it's a big big problem but def a limitation.
- instead of passing the password via the command line we could use an environment variable so it doesn't pop up in the list of running tasks.

What do you think?

DavidWittman:
Hey @oliver006... I have amended my commit to improve this PR per your suggestions:

 * The `redis.password` flag will now default to `os.Getenv("REDIS_PASSWORD")`, which allows the user to securely include a password.
 * Bump error count after invalid authentication
 * Added documentation to the README

I agree the multi-password thing is a limitation, but we'd need a much more robust configuration to be able to handle that... This should suffice for now, but I'll be happy to take a crack at it later if it comes a common request.

oliver006:
:+1: 

Thanks for your patience and the pull request!

PR #4, opened 2015-10-23
Support older versions of redis which do not have key ttl per DB

wrouesnel:
This small change adds support for much older versions of redis which do not have TTL lines in the keys field. I've tested it with our old redis 2.4.15 installs and it works fine with this change.


oliver006:
Thanks for the PR

Issue #3, opened 2015-09-23
add gauge `redis_used_memory_percent`?


iwinux:
With `used_memory` in `INFO memory`, plus `CONFIG GET maxmemory`.

oliver006:
Interesting, let me look into that.

oliver006:
@iwinux I pushed up commit https://github.com/oliver006/redis_exporter/commit/e4cf9c925c03214ff727cc4a42559fabf33970d8 that adds a gauge for `redis_config_maxmemory` so you can create a metric that uses used_memory and maxmemory to calculate usage percentage.
Let me know if that works for you.

iwinux:
Thanks :)

I will try this out later and reply back soon.

iwinux:
Hi,

Sorry for the late reply.

`CONFIG GET maxmemory` would not work as expected, because command `CONFIG` of our Redis instances are renamed with secret prefixes (to prevent mis-operation).

Cas-pian:
Seems the `maxmemory` metric is missing from the newest version? The [commit](https://github.com/oliver006/redis_exporter/commit/e4cf9c925c03214ff727cc4a42559fabf33970d8) is not included?

oliver006:
Ah, interesting. Looks like this got removed in April 2017 via https://github.com/oliver006/redis_exporter/pull/64
and now the metric is available via `redis_memory_max_bytes` (but only Redis versions that expose that info via `CONFIG`).
There's an open issue for this here: https://github.com/oliver006/redis_exporter/issues/82

Let me see if I can find some time over the next few days/weeks to bring back support for the `config_` metrics but in the meantime, PRs are welcome.

Cas-pian:
En, seems AWS INFO GET doesnot have maxmemory , but for common redis it works. Seems it's better to retrive the data in different way?

PR #8, opened 2016-05-25
customizable `CONFIG` command

iwinux:
Hi,

To solve the problem I mentioned in #3, I've added a new CLI argument `-redis.config-cmd`, which allows command `CONFIG` to be customized.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,8 +2,17 @@ machine:
   services:
     - redis
   environment:
-    IMPORT_PATH: "github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME"
+    PKG: github.com/oliver006/redis_exporter
+    GO_PKG_PATH: $HOME/.go_workspace/src/$PKG
 
+dependencies:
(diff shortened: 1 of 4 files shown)


oliver006:
Thanks!

Interesting, I didn't know about comand renaming (more info here for future reference: https://github.com/antirez/redis/blob/unstable/redis.conf#L460-L477).

I'll have a more in-depth look later but two quick thoughts:
- I think the logic here is wrong https://github.com/oliver006/redis_exporter/pull/8/files#diff-0afc679325bd208c1d470d1dca17b6bbR168 this will only ever use the first `config` item of the split list (and I think it's also wrong for passwords)
- We should default to `CONFIG` in the `NewRedisExporter` method if the `configCmds` struct is empty, that'll make the change backwards compatible.

iwinux:
Thanks for review :)

I'll fix issues you mention and update tests (maybe this weekend).

iwinux:
get CI working with forked repo (finally!): https://circleci.com/gh/iwinux/redis_exporter/22

PR #10, opened 2016-06-15
redis_exporter dockerfile

tomzhang:


oliver006:
Hey there, can you provide a bit of context as to what this change is about? Thx!

PR #22, opened 2016-08-30
add support for metrics about the length of specified queues

jeremybz:
This PR adds basic support for metrics about the length of specified queues.
` -check.keys KEYNAME` as a command line option allows a comma-separated list of keys.
The exporter will expose them as gauges of the form:
```
# HELP redis_queuelength_KEYNAME 
# TYPE redis_queuelength_KEYNAME gauge
redis_queuelength_KEYNAME{addr="localhost:6379"} 0
```

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -26,6 +26,7 @@ type Exporter struct {
 	scrapeErrors prometheus.Gauge
 	totalScrapes prometheus.Counter
 	metrics      map[string]*prometheus.GaugeVec
+    checkKeys    string
 	sync.RWMutex
 }
 
@@ -116,7 +117,7 @@ func (e *Exporter) initGauges() {
 }
 
 // NewRedisExporter returns a new exporter of Redis metrics.
-func NewRedisExporter(redis RedisHost, namespace string) (*Exporter, error) {
+func NewRedisExporter(redis RedisHost, namespace string, checkKeys string) (*Exporter, error) {
 	for _, addr := range redis.Addrs {
 		parts := strings.Split(addr, ":")
 		if len(parts) != 2 {
@@ -127,6 +128,7 @@ func NewRedisExporter(redis RedisHost, namespace string) (*Exporter, error) {
 	e := Exporter{
 		redis:     redis,
 		namespace: namespace,
+        checkKeys: checkKeys,
 
 		duration: prometheus.NewGauge(prometheus.GaugeOpts{
 			Namespace: namespace,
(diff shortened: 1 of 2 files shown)


mfouilleul:
In my opinion we should think about adding a "more generic" way to add counters for these typical "get value" needs. The code use to create this typical metric _redis_queuelength_KEYNAME_ will be the same as for instance _redis_get_KEYNAME_. The only things to take care is the data type returned.

mfouilleul:
To complete with a example, I had a discussion with a dev who ask me if he can track a value (a incremental counter) through the redis_exporter (easy to graph), it is basically the same piece of code than for redis_queuelength_KEYNAME.

oliver006:
I like the idea of making this more generic.
Exporting set length could be another useful feature. Also, I wouldn't generate a metric per key but use labels, I think that's the more prometheus-esque way of doing things.

I'll be away from a computer the next few days but if you guys have idea what to do then go ahead, I'll catch up later.

jeremybz:
I would have liked to have metrics of the form redis_queuelength{ key="KEYNAME" }, but I wasn't clever enough to do it myself.  If you can point me in the right direction I'll be happy to change it.

oliver006:
Back online...

I think what's needed is a list of keys (comma separated? name: ?) which than can be used to run this list of commands:
- HLEN
- LLEN
- SCARD
- ZCARD
- PFCOUNT
- STRLEN
- more? 

and whatever succeeds is published as `redis_key_length{key=KEY}` (better name suggestions welcome). 
In addition, let's also provide a metric where we can export key values and use `GET` and, if conversion to float succeeds, publish it as `redis_key_value{key=KEY}` (using a separate config parameter).

It should be possible to specify which DB key names refer to, so something like `###:XXX` where `###` is the db number (defaulting to 0 if not present) and `XXX` is the key name.

@jeremybz - want to give it a try? Let me know. If not, no problem and I'll look into it later this week.
This is how you use labels when using labels: https://github.com/jeremybz/redis_exporter/blob/d38b9ff382dd819d0d5948466ef7b2ad5789c101/exporter/redis.go#L377

and here: https://github.com/jeremybz/redis_exporter/blob/d38b9ff382dd819d0d5948466ef7b2ad5789c101/exporter/redis.go#L102-L106
is how you register a metric with labels.

@mfouilleul - do you think this will address your use case ?

oliver006:
@jeremybz @mfouilleul see https://github.com/oliver006/redis_exporter/pull/24 - let me know what you think.

oliver006:
Closing this, functionality was added via #24

PR #24, opened 2016-09-17
Allow for exporting key values and sizes

oliver006:


mfouilleul:
Such a good news!

oliver006:
Alright then, let's do this.

Issue #26, opened 2016-09-23
keys-checks with ':'


mfouilleul:
Hello,

It seams that the key-check feature doesn't work with keys with ':' in the name, example "cntr:flood". 

Max.

oliver006:
Ah yes, it uses `:` to separate `db##:keyName`.

Here's an idea, the exporter will expect the key to urlencoded so in your case it would be `cntr%3aflood` and `:` will continue to be the separator.
Would that work?

oliver006:
See PR #27

PR #27, opened 2016-09-23
urlencode check keys

oliver006:


oliver006:
For #26

oliver006:
Made a little change. I kept the urlencoding but I changed the separator to `=` so it's easier to use `:` in a key name and it doesn't need to be urlencoded (but can be).

mfouilleul:
Good one!

PR #33, opened 2016-10-24
Improve metrics in line with best practices.

brian-brazil:
Counters should end with _total.
Units should be specified where known.
Group related metrics together.
Remove machine-level metrics, they should come from the node exporter.
Remove calculated rates, they should be calculated in Prometheus.
Convert units to seconds where possible.


oliver006:
Looks good, thanks for the PR!

PR #45, opened 2016-12-02
support REDIS_ADDR environment variable

rdtr:
Hi,

Thank you for the great exporter for Redis!

It it okay to change so that we can overwrite `redis.addr` (like `redis.password`) with an environment variable?

We already set redis address consumable as an environment variable from any docker containers with Kubernetes, so we want just pass it to this exporter too, instead of hard coding an address in a flag.

I added a very small func to deal with the default value. Let me know if it's a problem.

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,8 +13,8 @@ import (
 )
 
 var (
-	redisAddr     = flag.String("redis.addr", "redis://localhost:6379", "Address of one or more redis nodes, separated by separator")
-	redisPassword = flag.String("redis.password", os.Getenv("REDIS_PASSWORD"), "Password for one or more redis nodes, separated by separator")
+	redisAddr     = flag.String("redis.addr", getEnv("REDIS_ADDR", "redis://localhost:6379"), "Address of one or more redis nodes, separated by separator")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -89,3 +89,13 @@ func main() {
 	log.Printf("Connecting to redis hosts: %#v", addrs)
 	log.Fatal(http.ListenAndServe(*listenAddress, nil))
 }
+
+// getEnv gets an environment variable from a given key and if it doesn't exist,
+// returns defaultVal given.
+func getEnv(key string, defaultVal string) string {
+	envVal, ok := os.LookupEnv(key)
+	if !ok {
```
oliver006: --> 
```
if envVal, ok := os.LookupEnv(key); ok {
    return envVal
}
return defaultVal

oliver006:
I like that you put it in a helper function. Can you please compact the `if .. LookupEnv... ` a little as per my comment and then we can ship this.

rdtr:
Thank you for the reply!

I modified the code, updated README.md, and rebased my commit. Please check when you have chance.

oliver006:
Looks good, thanks for your PR!

Issue #39, opened 2016-11-15
Add redis_up gauge


filippog:
Hi!
thanks for redis_exporter!
In line with other exporters (e.g. memcached) it'd be nice to provide a gauge `redis_up` 0/1 whenever redis scraping fails or succeeds.

thanks!

oliver006:
There is `exporter_last_scrape_error`, see code here: https://github.com/oliver006/redis_exporter/blob/master/exporter/redis.go#L168

I think that provides what you're looking for?

filippog:
Sort of, it is the other way around wrt to what it means though. I think it'd be nice to have the `up` gauge for consistency with other exporters if anything, would also help counting how many redis instances can be scraped, alerting, etc.

oliver006:
Hmm, ok, I see. I'll see that I add an `up` gauge but in the meantime you can use the existing gauge to pretty much the same effect.

eagafonov:
Oops, I've implemented it accidentally... Wait for a MR soon

PR #46, opened 2016-12-17
Add redis_up gauge

eagafonov:
Closes https://github.com/oliver006/redis_exporter/issues/39

The implementation is straightforward: set redis_up to 0 before the connection is established (so any connection problems error will produce redis_up = 0) and set to 1 once any metric is fetched from redis server.

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -408,6 +408,8 @@ func (e *Exporter) scrape(scrapes chan<- scrapeResult) {
 		var c redis.Conn
 		var err error
 
+		scrapes <- scrapeResult{Name: "up", Addr: addr, Value: 0}
+
 		var options []redis.DialOption
 		if len(e.redis.Passwords) > idx && e.redis.Passwords[idx] != "" {
 			options = append(options, redis.DialPassword(e.redis.Passwords[idx]))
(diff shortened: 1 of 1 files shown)


exporter/redis.go
```diff
@@ -408,6 +408,8 @@ func (e *Exporter) scrape(scrapes chan<- scrapeResult) {
 		var c redis.Conn
 		var err error
 
+		scrapes <- scrapeResult{Name: "up", Addr: addr, Value: 0}
```
oliver006: Doesn't this mark the exporter as down until line 449 marks it as back up?

eagafonov: I suppose the exporter must not publish partial metrics i.e. if /metrics is called in the middle of scrape(). As far as I understand, metrics updated after scrape() exists so redis_up will be updated correctly.

Correct me if I'm wrong

oliver006: Nope, you're right, I'm wrong, this is good to go.

brian-brazil: You should be sending a given timeseries only once. Without testing it, redis_up will always be 0 and every time it's up Prometheus will also log a warning.

PR #48, opened 2016-12-21
Added CLUSTER INFO support when cluster enabled

vvanholl:


exporter/redis.go
```diff
@@ -436,16 +436,31 @@ func (e *Exporter) scrape(scrapes chan<- scrapeResult) {
...
 		log.Debugf("connected to: %s", addr)
 
+		is_cluster := false
+
 		info, err := redis.String(c.Do("INFO", "ALL"))
 		if err == nil {
 			err = e.extractInfoMetrics(info, addr, scrapes)
-		}
-		if err != nil {
+			if strings.Index(info, "cluster_enabled:1") != -1 {
```
oliver006: Why not move that `if` check to line 453 and only do the `CLUSTER INFO` if the first `extractInfoMetrics` in 443 succeeded? That way you also won't need the extra `is_cluster` variable.

exporter/redis.go
```diff
@@ -436,16 +436,31 @@ func (e *Exporter) scrape(scrapes chan<- scrapeResult) {
...
+		} else {
 			log.Printf("redis err: %s", err)
 			errorCount++
 			continue
 		}
 
+		if is_cluster {
+			info, err = redis.String(c.Do("CLUSTER", "INFO"))
+			if err == nil {
+				err = e.extractInfoMetrics(info, addr, scrapes)
```
oliver006: Any more metrics that need to be added to `metricMap` as a result of this?

oliver006:
Thanks for the PR!

exporter/redis.go
vvanholl: You are right. I am updating code for a new commit.

exporter/redis.go
vvanholl: metricMap role is to give more fancy names to metrics ?

oliver006: Yeah, see #33 which Brian submitted to improve metrics names in line with Prometheus best practices.

oliver006: I see you added two names to the map - let me know if this are all the names you want to add and I'll merge this.

vvanholl:
@oliver006 According to my opinion,  these are the 2 only metrics which should change name.

oliver006:
👍

PR #56, opened 2017-02-24
Use `redis_up` for templating

mrwacky42:
Believe it or not, I had a redis instance with no connected clients, so the templating failed.
`redis_up` is probably a more "robust" choice.


oliver006:
I totally believe it. Thanks for the PR!

PR #57, opened 2017-03-07
Add flag -redis.alias. Allow to give an alias to the redis node(s) address

StephaneBunel:
It could be more friendly (especially in graph) to display the alias than just the node address. Add a modified dashboard for Grafana.


StephaneBunel:
grafana prometheus redis host alias

StephaneBunel:
Useful when you have multiple Redis instance on the same server.

oliver006:
Thanks for the PR!

I'll be away from my laptop the next couple of days but will try to look at it as soon as I can.

oliver006:
👍

vvanholl:
@StephaneBunel : Good job !

Issue #32, opened 2016-10-24
An exporter should only monitor one host


brian-brazil:
I noticed that the redis exporter has the ability to scrape multiple redis servers, and adds a `addr` label that partially duplicates the `instance` label. Per https://prometheus.io/docs/instrumenting/writing_exporters/#deployment there should be a 1-to-1 mapping between exporters and application instances.

oliver006:
Some folks use the exporter to monitor e.g.  an external Azure Redis instance and I'd like to keep that as an option. 
I'm also not sure I understand what the actual problem is - the addional `addr` label? 

What is your suggestion to address this? 

Maybe drop/not add the `addr` label when no explicit redis instance address was provided?

brian-brazil:
> Some folks use the exporter to monitor e.g. an external Azure Redis instance and I'd like to keep that as an option.

What I'm suggesting doesn't preclude that.

> I'm also not sure I understand what the actual problem is - the addional addr label?

The `addr` label, and that you allow monitoring more than one redis from one exporter.

oliver006:
Ok, I see. Well, I like (and use) the ability to monitor more than one Redis instance and it doesn't really cause any harm or trouble so I'd prefer to keep this.

PR #60, opened 2017-03-23
Add Sentinel support

### LINKED ISSUE #59: Add sentinel support
Especially the master discovery part.

jmastr:
Sentinels are used for health checking, fallback strategy and master
IP address discovery. Latter is necessary in a Master/Slave/Sentinel
setup to retrieve the data of the correct Redis node (master) for the
exporter.

Two command line parameters are introduced: 'sentinel.addr' and
'sentinel.mastr'. Both are string slices and need to have the same
number of strings, since for master discovery we need the address of one
sentinel and the name of the master to discover.

If the parameters are provided and the master discovery was successful,
the IP address(es) of the discovered master(s) are appended to the
addresses given by 'redis.addr'.

If 'redis.addr' is not specified only the discovered addresses are taken
into account.

Closes #59

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,19 +6,24 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/oliver006/redis_exporter/exporter"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/garyburd/redigo/redis"
+	"github.com/FZambia/go-sentinel"
 )
 
 var (
-	redisAddr     = flag.String("redis.addr", getEnv("REDIS_ADDR", "redis://localhost:6379"), "Address of one or more redis nodes, separated by separator")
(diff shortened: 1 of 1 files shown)


oliver006:
Sorry I didn't have time to get back to you earlier.

I think I misunderstood your original issue about Sentinel support. I thought it was about adding additional metrics for redis sentinel, not about having sentinel service discovery in the exporter.

The redis instance discovery should be handled by Prometheus itself (via whatever SD mechanism you use right now, Consul, Kubernetes, etc). #32 has a bit more background and a link to here: https://prometheus.io/docs/instrumenting/writing_exporters/#deployment explaining the rational. 

Let me know if that makes sense.

jmastr:
I have three Redis setups in one Kubernetes cluster: one master, two slaves, and three sentinels each. If I understand the deployment strategy correctly, I need to put exporters to every master pod. The challenge here is, that I do not know the master before start, because Redis has its own master election mechanism. And even if I would know it before start, a pod can disappear and a new master can be elected.

oliver006:
I think the deployment strategy would be to have a redis_exporter with each redis pod.

jmastr:
I'll try that.

PR #73, opened 2017-05-21
Makefile

mjtrangoni:
Hi @oliver006,

I did a bunch of changes here, which are the things I enjoy using/seeing when building an exporter.
- Implement a **Makefile** using [promu](https://github.com/prometheus/promu) as build tool.
- Use more Prometheus Libraries.
- Update some **vendoring** adding **Glide** support to the Makefile. 
- Add a [megacheck](https://github.com/dominikh/go-tools/tree/master/cmd/megacheck) target and fix two issues detected by this tool.
- Fix `make test` checking if **redis-server** is running.
- Implement `docker` and `test-docker` targets.
- Use **govendor** instead of **glide**.
- Implement `promu crossbuild`.

What I would like to do in other PRs is,
- Deprecate **build.sh** if `promu` is working well with CircleCI.


oliver006:
Thanks for the PR! I'll try to find some time tomorrow, I had a quick look and got a couple of comments and questions.

.promu.yml
```diff
@@ -0,0 +1,17 @@
...
+    ldflags: |
+        -s
+        -X {{repoPath}}/vendor/github.com/prometheus/common/version.Version={{.Version}}
+        -X {{repoPath}}/vendor/github.com/prometheus/common/version.Revision={{.Revision}}
+        -X {{repoPath}}/vendor/github.com/prometheus/common/version.Branch={{.Branch}}
+        -X {{repoPath}}/vendor/github.com/prometheus/common/version.BuildUser={{user}}@{{host}}
+        -X {{repoPath}}/vendor/github.com/prometheus/common/version.BuildDate={{date "20060102-15:04:05"}}
+tarball:
+    files:
+        - LICENSE
```
oliver006: Just tried `promu tarball` and it complains about missing NOTICE file so I think you can remove these two.

.promu.yml
```diff
@@ -0,0 +1,17 @@
...
+        - NOTICE
```
oliver006: Can you also add the cross builds - i think that will come in handy with the automated circleCI builds:

```
crossbuild:
    platforms:
        - linux/amd64
        - linux/386
        - darwin/amd64
        - darwin/386
        - netbsd/amd64
        - netbsd/386
        - linux/arm
        - linux/arm64
```

oliver006:
I like where this is going, this is making the build process a lot better compared to what's there right now.
But it'd be good to also add/fix the automated build as part of this PR to use the make file and the new make system and not have two build processes in place.
I'm a little short with time but can help if needed, let me know.

mjtrangoni:
Ok, Thanks! I will be working on docker builds this week, and then we can drop `build.sh` too.

Makefile
```diff
@@ -0,0 +1,78 @@
+# MIT License
```
mjtrangoni: fixed

.promu.yml
```diff
@@ -0,0 +1,17 @@
+go:
+    cgo: true
+repository:
+    path: github.com/oliver006/redis_exporter
+build:
+    flags: -a -tags 'netgo static_build'
+    ldflags: |
```
mjtrangoni: fixed. There is a problem here, so I set CGO_ENABLE=0 before `promu`manually. See [#60](https://github.com/prometheus/promu/issues/60)

VERSION
```diff
@@ -0,0 +1 @@
+0.11
```
mjtrangoni: fixed

Makefile
```diff
@@ -0,0 +1,78 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+GO              ?= GO15VENDOREXPERIMENT=1 go
+GOPATH          := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
+PROMU           ?= $(GOPATH)/bin/promu
+MEGACHECK       ?= $(GOPATH)/bin/megacheck
```
mjtrangoni: fixed, I had to include $(MEGACHECK) next to megacheck target.

main.go
```diff
@@ -73,30 +59,26 @@ func main() {
 	}
 	prometheus.MustRegister(exp)
 
-	buildInfo := prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
mjtrangoni: This information is handled by promu, see the `.promu.yml`

.promu.yml
oliver006: 👍

.travis.yml
```diff
@@ -0,0 +1,10 @@
+sudo: false
```
oliver006: Why Travis? CircleCI is currently used for CI and I don't intend to switch to Travis (unless there are very good reasons).

Makefile
```diff
@@ -20,8 +25,12 @@ dependencies:
 	@glide --quiet update
 
 test:
-	@echo ">> running tests"
+	@echo ">> starting local redis-server on port $(REDIS_PORT)"
```
oliver006: I'd not make this part of the tests.
E.g. I run my local dev redis instance in a docker container and don't need that step. I think it's fair to assume that whoever runs the test has a local redis instance running.

.travis.yml
mjtrangoni: I don't want to switch to Travis. My idea is to use CircleCI with **docker** and TravisCI without docker, like prometheus does. Is it ok for you?

Makefile
mjtrangoni: Ok, good point! I will be checking if a redis-server is running, and if not, then I will be skipping `test` with a message. Is it ok?

oliver006: Sure, that'd be the most elegant way.

.travis.yml
oliver006: Interesting. But why use both? Currently, CircleCI is used for tests, docker build, Github release uploads, etc - what's the benefit of using both?

Makefile
mjtrangoni: Done.

.travis.yml
mjtrangoni: I will be implementing only CircleCI for the time, dropping this commit.

mjtrangoni:
I will be working on `promu crossbuild` tomorrow.

mjtrangoni:
Hi @oliver006, 
I am almost done here and need a new review. By the way, could you please check and fix the **deployment** part of `circle.yml`? It should be something similar to  [circle.yml](https://github.com/prometheus/node_exporter/blob/master/circle.yml), but I do not have the rights to test it.
Thanks!

oliver006:
Hey, thanks A LOT for your contributions, I'll try to get to them as soon as possible, probably later this week.

circle.yml
```diff
@@ -3,44 +3,46 @@ machine:
     - redis
     - docker
   post:
-    - /usr/bin/redis-server --port 6380:
+    - /usr/bin/redis-server --port 6379:
```
oliver006: I think this needs to be `6380`. It brings up a second redis instance to run multi-instance tests (one on port 6379 is brought up by circle - see line 3 of this file)

circle.yml
```diff
@@ -3,44 +3,46 @@ machine:
...
   post:
     - if [ -n "$COVERALLS_TOKEN" ]; then /home/ubuntu/.go_workspace/bin/goveralls -coverprofile=$COVERAGE_PROFILE -service=circle-ci -repotoken=$COVERALLS_TOKEN ;  fi
-
+    - cd $SRC_LOCATION && make
 
 deployment:
   publish:
     tag: /v.*/
     commands:
       - go get github.com/mitchellh/gox
```
oliver006: Not sure if gox and ghr are still needed now that we have `promu release` but that said - I'm not sure if `promu release` will work like this out of the box without some config wrangling. Let me look into that, possibly as a second PR.

main.go
```diff
@@ -2,58 +2,44 @@ package main
...
-		log.SetLevel(log.DebugLevel)
-		log.Debugln("Enabling debug output")
-	} else {
-		log.SetLevel(log.InfoLevel)
-	}
+	flag.Parse()
 
 	if *showVersion {
-		return
+		fmt.Fprintln(os.Stdout, version.Print("redis_exporter"))
```
oliver006: `fmt.Fprintln(os.Stdout` --> fmt.Println ?

test_image.sh
```diff
@@ -0,0 +1,39 @@
...
+    docker run -d -p "${port}":"${port}" --name redis_exporter "${docker_image}" > /dev/null
+}
+
+docker_stop() {
+    # Stop redis-server
+    docker rm --force $(docker ps -aqf "name=redis-server") $(docker ps -aqf "name=redis_exporter") > /dev/null
+}
+
+if [[ "$#" -ne 2 ]] ; then
+    echo "Usage: $0 quay.io/oliver006/redis_exporter:v0.12.0 9121" >&2
```
oliver006: quay? this should probably reference the docker hub address.

oliver006:
Gonna close this as it's been a good bit out of date now but I'll try to cherry pick commits and get them into master as I get around to it.

PR #75, opened 2017-06-09
[RDY] Support passing a file containing redis hosts

jburnham:
This allows one to manage a file separately from the service definition
and simply restart the service once the configured file is updated.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/csv"
 	"flag"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


oliver006:
Do you think it would make sense to add a feature to hook into the SIGHUP signal and allow to reload the file list?

main.go
```diff
@@ -54,14 +59,54 @@ func main() {
...
-	for len(passwords) < len(addrs) {
-		passwords = append(passwords, passwords[0])
+	if *redisFile != "" && *redisAddr != "" {
+		log.Fatal("Cannot specify both redis.addr and redis.file")
 	}
-	aliases := strings.Split(*redisAlias, *separator)
-	for len(aliases) < len(addrs) {
-		aliases = append(aliases, aliases[0])
+
+	if *redisFile != "" {
```
jburnham: Yea, I can do this.

jburnham:
@oliver006 SIGHUP support makes sense, but I believe is out of scope for the changes I wanted to make. I'm less comfortable with the changes I think would be necessary to support that logic and it could easily be implemented as a separate PR at a future date if you're ok with that. I have implemented the other suggestions that were made previously and this PR is ready for your review again. Thanks!

oliver006:
👍

robbiet480:
Can this be pushed to Docker Hub? Thanks!

oliver006:
Ah yes, totally forgot about that, done via https://github.com/oliver006/redis_exporter/releases/tag/v0.12

PR #83, opened 2017-07-03
--check-keys could use wildcards to search existing keys

albestia:
My first lines in golang... please be patient ;)

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -471,23 +471,39 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		if _, err := c.Do("SELECT", k.db); err != nil {
 			continue
 		}
-		if tempVal, err := c.Do("GET", k.key); err == nil && tempVal != nil {
-			if val, err := strconv.ParseFloat(fmt.Sprintf("%s", tempVal), 64); err == nil {
-				e.keyValues.WithLabelValues(addr, e.redis.Aliases[idx], "db"+k.db, k.key).Set(val)
+
+		obtainedKeys := []string{}
+		if strings.Contains(k.key, "*") {
+			log.Debugf("Provided check-keys contains '*' %s", k.key)
+			if tempVal, err := redis.Strings(c.Do("KEYS", k.key)); err == nil && tempVal != nil {
+				for _, tempKey := range tempVal {
+					log.Debugf("Append result: %s", tempKey)
+					obtainedKeys = append(obtainedKeys, tempKey)
+				}
 			}
+		} else {
+			obtainedKeys = append(obtainedKeys, k.key)
 		}
 
-		for _, op := range []string{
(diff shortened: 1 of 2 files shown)


exporter/redis.go
```diff
@@ -471,23 +471,39 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		if _, err := c.Do("SELECT", k.db); err != nil {
 			continue
 		}
-		if tempVal, err := c.Do("GET", k.key); err == nil && tempVal != nil {
-			if val, err := strconv.ParseFloat(fmt.Sprintf("%s", tempVal), 64); err == nil {
-				e.keyValues.WithLabelValues(addr, e.redis.Aliases[idx], "db"+k.db, k.key).Set(val)
+
+		obtainedKeys := []string{}
+		if strings.Contains(k.key, "*") {
```
oliver006: Interesting.

What do you think about doing the `KEYS` request every time, wildcard or not and if `k.key` matches a key in the DB then it'll pick it up anyway? 
Might simplify the code a little as the else block in line 485 isn't needed.

Thoughts?

exporter/redis.go
```diff
@@ -471,23 +471,39 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
...
-			"LLEN",
-			"SCARD",
-			"ZCARD",
-			"PFCOUNT",
-			"STRLEN",
-		} {
-			if tempVal, err := c.Do(op, k.key); err == nil && tempVal != nil {
-				e.keySizes.WithLabelValues(addr, e.redis.Aliases[idx], "db"+k.db, k.key).Set(float64(tempVal.(int64)))
-				break
+		for _, keyToSearch := range obtainedKeys {
```
oliver006: keyToSearch -> key

albestia:
Applied all your suggestions. I've tried to extend redis_test.go to do some tests on check-keys, but I'm too new to golang... I'll keep on trying and send you a new pull request if I'm able to do it.

oliver006:
Looks good!

If you want to give adding tests for wild-card matching a shot then you can have a look this test: 

https://github.com/albestia/redis_exporter/blob/1b7ee34a9b8431602c3d5cedcaff6f50ed0ed8c9/exporter/redis_test.go#L280

I'd add another test just like `TestKeyValuesAndSizes` and pass a wildcard string to `NewRedisExporter` and then the code down near 295-306 would need to get adjusted to look for a label for each of the keys matched.

Let me know if you want to give it a shot or I'll add it later.

albestia:
Did it my best, but unable to make it work. The worst thing is I'm unable to debug tests on my development environment. I'll read documentation en recreate my dev env to work with all needed tools.
Sorry.

oliver006:
No problem. I added a test for wildcard key matching, this should be good to go!

PR #87, opened 2017-07-24
Added support to retrieve latency stats from redis nodes

kpachhai:
There were lots of changes I needed to make than I originally anticipated.

1. Created two new metrics: latency_spike_timestamp(when the latency spike last occured) and latency_spike_millisecond(how much the latency spike was when it last occured)

2. Updated extractInfoMetrics method to handle this new latency stats

Let me know if it looks good.

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -279,37 +279,6 @@ func extractVal(s string) (val float64, err error) {
 	return
 }
 
-/* Helper function to extract results for latency stats as redis.Strings
-   complains if the output is a mix of byte array and something else
-   (in our case int64)
-*/
-func extractStrings(reply interface{}, err error) ([]string, error) {
-	if err != nil {
-		return nil, err
-	}
-	switch reply := reply.(type) {
-	case []interface{}:
-		result := make([]string, len(reply))
-		for i := range reply {
-			if reply[i] == nil {
-				continue
-			}
-			p, ok := reply[i].([]byte)
-			if !ok {
-				result[i] = fmt.Sprintf("%v", reply[i])
-			} else {
-				result[i] = string(p)
-			}
-		}
-		return result, nil
-	case nil:
-		return nil, redis.ErrNil
-	case redis.Error:
-		return nil, reply
-	}
-	return nil, fmt.Errorf("unexpected type for extractStrings, got type %T", reply)
-}
-
 /*
 	valid example: db0:keys=1,expires=0,avg_ttl=0
 */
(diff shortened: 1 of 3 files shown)


exporter/redis.go
```diff
@@ -136,6 +136,18 @@ func (e *Exporter) initGauges() {
 		Help:      "Avg TTL in seconds",
 	}, []string{"addr", "alias", "db"})
 
+	// Latency info
+	e.metrics["latency_spike_timestamp"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
oliver006: Can you change the metric name to `latency_spike_last` please?

exporter/redis.go
```diff
@@ -136,6 +136,18 @@ func (e *Exporter) initGauges() {
...
+		Namespace: e.namespace,
+		Name:      "latency_spike_timestamp",
+		Help:      "When the latency spike last occured",
+	}, []string{"addr", "alias", "event_name"})
+	e.metrics["latency_spike_millisecond"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
oliver006: latency_spike_millisecond -> latency_spike_milliseconds

exporter/redis.go
```diff
@@ -136,6 +136,18 @@ func (e *Exporter) initGauges() {
...
+		Namespace: e.namespace,
+		Name:      "latency_spike_millisecond",
+		Help:      "How much the latency spike was when it last occured",
```
oliver006: How about "Length of the last spike in milliseconds".

exporter/redis.go
```diff
@@ -467,6 +479,24 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		}
 	}
 
+	if reply, err := c.Do("LATENCY", "LATEST"); err != nil {
```
oliver006: I don't think the error is that useful. How about inverting the `if` statement and check for `err == nil` and then get rid of the `else` ?

exporter/redis.go
```diff
@@ -467,6 +479,24 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
...
+		log.Printf("redis err: %s", err)
+	} else {
+		var eventName string
+		var timestamp, millisecond, max int64
+		if tempVal, _ := reply.([]interface{}); len(tempVal) > 0 {
+			latencyResult := tempVal[0].([]interface{})
+			if _, err := redis.Scan(latencyResult, &eventName, &timestamp, &millisecond, &max); err != nil {
```
oliver006: Same thing here

exporter/redis.go
oliver006: (all the occurrences, map field, metrics name, etc)

exporter/redis.go
```diff
@@ -136,6 +136,18 @@ func (e *Exporter) initGauges() {
 		Help:      "Avg TTL in seconds",
 	}, []string{"addr", "alias", "db"})
 
+	// Latency info
+	e.metrics["latency_spike_timestamp"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: e.namespace,
+		Name:      "latency_spike_timestamp",
+		Help:      "When the latency spike last occured",
```
oliver006: What about "Unix timestamp of when the last latency spike occurred" ?

kpachhai:
New commit up. Fixed some naming issues that were addressed, removed some unneeded error checking and added tests in redis_test.go for newly added functionality.

kpachhai:
So, for some reason, it's failing the automated tests.
But, it worked fine when i tested on my own machine. What version of redis is the build using?

After looking into it, it's failing here:
--- FAIL: TestLatencySpike (0.01s)
	redis_test.go:62: couldn't setup redis, err: ERR Unsupported CONFIG parameter: LATENCY-MONITOR-THRESHOLD 
	redis_test.go:90: couldn't setup redis, err: ERR unknown command 'LATENCY' 
	redis_test.go:90: couldn't setup redis, err: ERR unknown command 'LATENCY' 

But, to me, this doesn't look right since i'm able to run "go test" just fine and i have no issues running locally.

oliver006:
Hmm, not sure what version of Redis is running on CircleCI, let me check.

oliver006:
Didn't have a chance to find out the Redis version on CircleCI but the error is
```
--- FAIL: TestLatencySpike (0.01s)
	redis_test.go:62: couldn't setup redis, err: ERR Unsupported CONFIG parameter: LATENCY-MONITOR-THRESHOLD 
	redis_test.go:90: couldn't setup redis, err: ERR unknown command 'LATENCY' 
	redis_test.go:90: couldn't setup redis, err: ERR unknown command 'LATENCY'
```

Maybe we should make the test optional and simply return if the `CONFIG LATENCY ...` COMMAND fails and at some point down the road we can see that the version of Redis in the test container will be updated.

kpachhai:
I have commented out the appropriate section until redis is updated to a more recent version in github container.

mjtrangoni:
Hi @kpachhai , 
I think you can try using a newest version in CircleCI. See, [here](https://discuss.circleci.com/t/how-to-run-redis-3-2/5815/5). My 2 cents.

kpachhai:
Nice. I updated CircleCI configuration and re-added the latency tests back. Worked like a charm. I think we're all good to go here.

PR #94, opened 2017-08-16
exclude a metric if it contains a dash

### LINKED ISSUE #93: invalid metric name "redis_cluster_stats_messages_auth-ack_received"
howdy, after a bad period on our redis cluster the exporter stopper working as we started seeing:

```
invalid metric name "redis_cluster_stats_messages_auth-ack_received"
```

There's also
```
redis_cluster_stats_messages_auth-req_sent
```

azr:
this fixed #93 for us :)

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -258,9 +258,6 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
 
 func includeMetric(s string) bool {
 
-	if strings.Contains(s, "-") {
-		return false
-	}
 	if strings.HasPrefix(s, "db") || strings.HasPrefix(s, "cmdstat_") || strings.HasPrefix(s, "cluster_") {
 		return true
 	}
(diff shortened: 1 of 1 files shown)


oliver006:
That would fix it 👍 

But what about instead changing the metric name and replace the `-` with `_` ?
In that way the metric is still exposed and available to users.

oliver006:
You could replace this line here:
https://github.com/azr/redis_exporter/blob/02859cc0fec36f8d529ccbcb9a16f8456f38d937/exporter/redis.go#L394
with something like 

`metricName := sanitizeMetricName(split[0])`

and 

```
func sanitizeMetricName(n string) {
    // replace "-" with "_" in n 
    // and maybe other characters?
    return n
}
```

azr:
Yup, looks better like this :)

azr:
But that metric is still there

oliver006:
Hmm, no idea why that is.

oliver006:
I just tried to reproduce that but don't see that error on Redis 3.2.8 - what redis version are you using and what's your cluster setup?

PR #103, opened 2017-09-20
Make exporter able to connect via bindings on cloud foundry

ihrigb:
Parameters via flags have priority, so the CF bindings are only
used if there are no addresses after evaluation of the flags and files
instead of localhost.


oliver006:
Thanks for the PR!

main.go
```diff
@@ -74,6 +75,10 @@ func main() {
 		addrs, passwords, aliases = loadRedisArgs(*redisAddr, *redisPassword, *redisAlias, *separator)
 	}
 
+	if (len(addrs) == 0 || addrsOnlyContainLocalhost(addrs)) && isOnCloudFoundry() {
```
oliver006: I think it would make more sense to check CF settings when e.g. there's a flag `use-cf` present. I'm not familiar with Cloud Foundry, what do you think?

main.go
```diff
@@ -103,6 +108,10 @@ func main() {
 						`))
 	})
 
+	if isOnCloudFoundry() {
```
oliver006: Why do this explicitly and not via the cmd line config?

main.go
```diff
@@ -175,3 +184,54 @@ func getEnv(key string, defaultVal string) string {
...
+func getCloudFoundryRedisBindings() ([]string, []string, []string) {
+	if !isOnCloudFoundry() {
+		return []string{}, []string{}, []string{}
+	}
+
+	var addrs []string
+	var passwords []string
+	var aliases []string
+
+	appEnv, _ := cfenv.Current()
```
oliver006: Does this need an `err` check? Would line 203 panic with nil reference if it fails?

main.go
```diff
@@ -175,3 +184,54 @@ func getEnv(key string, defaultVal string) string {
...
+			addrs = append(addrs, addr)
+			passwords = append(passwords, password)
+			aliases = append(aliases, alias)
+		}
+	}
+
+	return addrs, passwords, aliases
+}
+
+func containsTagCaseInsensitive(tags []string, tag string) bool {
```
oliver006: Is this used?

manifest.yml
```diff
@@ -0,0 +1,4 @@
+---
```
oliver006: What's this file for?

main.go
ihrigb: I need to doublecheck how to set this flag for a CF deployment. If I find out how to, I will make this dependent on the flag.

main.go
ihrigb: See above.

main.go
ihrigb: We already checked if we are on a CF environment, so err might not be interesting here. I will take another look to be sure.

main.go
ihrigb: Uhm no. Forgot to remove it.

manifest.yml
ihrigb: It is a deployment descriptor for CF. Take a look here for more information: https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html

oliver006: Ok, but why does it need to be included in the repo?

main.go
```diff
@@ -175,3 +181,34 @@ func getEnv(key string, defaultVal string) string {
...
 	return defaultVal
 }
+
+func getCloudFoundryRedisBindings() (addrs, passwords, aliases []string) {
+	if !cfenv.IsRunningOnCF() {
+		return
+	}
+
+	appEnv, err := cfenv.Current()
+	if err != nil {
```

main.go
```diff
@@ -26,6 +27,7 @@ var (
 	isDebug       = flag.Bool("debug", false, "Output verbose debug information")
 	logFormat     = flag.String("log-format", "txt", "Log format, valid options are txt and json")
 	showVersion   = flag.Bool("version", false, "Show version information and exit")
+	useCfBinding  = flag.Bool("use-cf-binding", false, "Use Cloud foundry service bindings")
```
oliver006: `use-cf-binding` --> `use-cf-bindings`
`Cloud foundry` -> `Cloud Foundry`

can you please also add an entry to the README for this option?

main.go
```diff
@@ -74,6 +76,10 @@ func main() {
 		addrs, passwords, aliases = loadRedisArgs(*redisAddr, *redisPassword, *redisAlias, *separator)
 	}
 
+	if *useCfBinding {
```
oliver006: nice, this is a lot better.

Can you make this part of the if block 67-77 and turn that into a switch statement?

manifest.yml
ihrigb: It defines basic necessities for the deployment, like the CF buildpack and the command (with flags) to be executed.
In the end it makes the deployment easier, as relevant stuff is predefined.

main.go
ihrigb: How would I put this into a switch statement?

main.go
ihrigb: Done

main.go
oliver006:
```
switch {
    case *redisFile != "": ....

    case *useCfBinding: ....

    default: ....


}
```

manifest.yml
oliver006: Ok. Can you move the file intro contrib and leave a note in the README?

main.go
oliver006: Thx. This is still missing the entry in the README though.

manifest.yml
ihrigb: Ok

main.go
ihrigb: Done

ihrigb:
Done.

oliver006:
Thanks!

PR #105, opened 2017-09-26
Add replication metric

siavashs:
This PR adds a `replication` metric, this is useful in cluster setups were you need to find the master node or some slave information.

Example:
```
redis_replication{master_host="10.91.64.152",master_link_status="up",master_port="6379",role="slave",slave_read_only="1"} 1
```

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -110,7 +110,7 @@ var (
 	}
 
 	instanceInfoFields = map[string]bool{"role": true, "redis_version": true, "redis_build_id": true, "redis_mode": true, "os": true}
-	slaveFields        = map[string]bool{"master_host": true, "master_port": true, "master_link_status": true, "slave_read_only": true}
+	slaveInfoFields    = map[string]bool{"master_host": true, "master_port": true, "slave_read_only": true}
 )
 
 func (e *Exporter) initGauges() {
@@ -121,11 +121,16 @@ func (e *Exporter) initGauges() {
 		Name:      "instance_info",
 		Help:      "Information about the Redis instance",
 	}, []string{"addr", "alias", "role", "redis_version", "redis_build_id", "redis_mode", "os"})
-	e.metrics["slave"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	e.metrics["slave_info"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: e.namespace,
-		Name:      "slave",
+		Name:      "slave_info",
 		Help:      "Information about the Redis slave",
-	}, []string{"addr", "alias", "master_host", "master_port", "master_link_status", "read_only"})
+	}, []string{"addr", "alias", "master_host", "master_port", "read_only"})
+	e.metrics["master_link_up"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: e.namespace,
+		Name:      "master_link_up",
+		Help:      "Master link status on Redis slave",
+	}, []string{"addr", "alias"})
 	e.metrics["db_keys"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: e.namespace,
 		Name:      "db_keys",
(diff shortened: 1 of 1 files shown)


oliver006:
Thanks for the PR!

Not sure if this works as intended, two things come to mind here at first glance:

1) this is missing (at least) `addr` in the labels so it's not possible to get the information for only one cluster only if the exporter is used to monitor multiple clusters.
2) Around line 444 you probably want to check if replication is on and not publish the metric otherwise.

Personally, I don't use replication with Redis so can't really judge if this info is useful (and if it's useful in this way, using labels).

siavashs:
Redis has replication enabled, so a standalone instance is a **master**:
```
# Replication
role:master
connected_slaves:0
...
```
The `role` value is a **string** so it should be a label.
My initial idea was to add `role` label to `instance_info` metric (which makes sense now, since replication is always enabled), so I think it's better to move it there and rename `replication` to `slave` and only enable it when `role == slave`.
Waiting for your comments.

oliver006:
Thanks for the explanation. I think that make sense to move the fields to the `instance_info` metrics  as they are always in the context of the currently scraped host.

siavashs:
I just made the discussed changes.

The metrics will look like this:
on **master**:
```
redis_instance_info{addr="redis://redis1:6379",alias="",os="Linux 4.4.0-31-generic x86_64",redis_build_id="a44a05d76f06a5d9",redis_mode="standalone",redis_version="2.8.4",role="slave"} 1
```
on **slave**:
```
redis_instance_info{addr="redis://redis2:6379",alias="",os="Linux 4.4.0-31-generic x86_64",redis_build_id="a44a05d76f06a5d9",redis_mode="standalone",redis_version="2.8.4",role="slave"} 1
redis_slave{addr="redis://redis2:6379",alias="",master_host="10.91.64.152",master_link_status="up",master_port="6379",read_only="1"} 1
```

shahsaifi:
Just checking, is this change available in master branch?

exporter/redis.go
```diff
@@ -119,7 +120,12 @@ func (e *Exporter) initGauges() {
 		Namespace: e.namespace,
 		Name:      "instance_info",
 		Help:      "Information about the Redis instance",
-	}, []string{"addr", "alias", "redis_version", "redis_build_id", "redis_mode", "os"})
+	}, []string{"addr", "alias", "role", "redis_version", "redis_build_id", "redis_mode", "os"})
+	e.metrics["slave"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: e.namespace,
+		Name:      "slave",
```
oliver006: Should this be `slave_info`? (Analogous to `instance_info` above?)

exporter/redis.go
```diff
@@ -119,7 +120,12 @@ func (e *Exporter) initGauges() {
...
+		Help:      "Information about the Redis slave",
+	}, []string{"addr", "alias", "master_host", "master_port", "master_link_status", "read_only"})
```
oliver006: I have a question about `master_link_status`: Would it make sense to break this out as a separate gauge so you can alert on when the link is down?

exporter/redis.go
```diff
@@ -325,6 +331,7 @@ func (e *Exporter) extractInfoMetrics(info, addr string, alias string, scrapes c
 	lines := strings.Split(info, "\r\n")
 
 	instanceInfo := map[string]string{}
+	slave := map[string]string{}
```
oliver006: Let's name this `slaveInfo`.

siavashs:
Sample output from latest changes:
```
# HELP redis_instance_info Information about the Redis instance
# TYPE redis_instance_info gauge
redis_instance_info{addr="redis://redis1:6379",alias="",os="Linux 4.4.0-31-generic x86_64",redis_build_id="a44a05d76f06a5d9",redis_mode="standalone",redis_version="2.8.4",role="slave"} 1

# HELP redis_slave_info Information about the Redis slave
# TYPE redis_slave_info gauge
redis_slave_info{addr="redis://redis1:6379",alias="",master_host="10.91.64.152",master_port="6379",read_only="1"} 1

# HELP redis_master_link_up Master link status on Redis slave
# TYPE redis_master_link_up gauge
redis_master_link_up{addr="redis://redis1:6379",alias=""} 1
```

exporter/redis.go
```diff
@@ -345,6 +357,22 @@ func (e *Exporter) extractInfoMetrics(info, addr string, alias string, scrapes c
...
 		}
 
+		if _, ok := slaveInfoFields[split[0]]; ok {
+			slaveInfo[split[0]] = split[1]
+			continue
+		}
+
+		if split[0] == "master_link_status" {
+			e.metricsMtx.RLock()
+			if split[1] == "up" {
```

PR #108, opened 2017-10-23
Changes for exporting redis only metrics if needed

nipuntalukdar:
Added support for exporting Redis only metrics if required. Default behavior is same as existing behavior, i.e., all Redis metrics including metrics for go runtime and go http handlers are exported. If the exporter is run with "-redis-only-metrics true", then only Redis specific metrices are exported.

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,23 +16,23 @@ import (
 )
 
 var (
-	redisAddr     = flag.String("redis.addr", getEnv("REDIS_ADDR", ""), "Address of one or more redis nodes, separated by separator")
-	redisFile     = flag.String("redis.file", getEnv("REDIS_FILE", ""), "Path to file containing one or more redis nodes, separated by newline. NOTE: mutually exclusive with redis.addr")
-	redisPassword = flag.String("redis.password", getEnv("REDIS_PASSWORD", ""), "Password for one or more redis nodes, separated by separator")
-	redisAlias    = flag.String("redis.alias", getEnv("REDIS_ALIAS", ""), "Redis instance alias for one or more redis nodes, separated by separator")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -28,6 +29,7 @@ var (
 	logFormat     = flag.String("log-format", "txt", "Log format, valid options are txt and json")
 	showVersion   = flag.Bool("version", false, "Show version information and exit")
 	useCfBindings = flag.Bool("use-cf-bindings", false, "Use Cloud Foundry service bindings")
+	redisOnly     = flag.Bool("redis-only-metrics", false, "Whether to export go runtime metrics also")
```
oliver006: Could you rename this to `redisMetricsOnly`?

main.go
```diff
@@ -86,16 +88,25 @@ func main() {
...
 
 	buildInfo := prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "redis_exporter_build_info",
 		Help: "redis exporter build_info",
 	}, []string{"version", "commit_sha", "build_date", "golang_version"})
-	prometheus.MustRegister(buildInfo)
 	buildInfo.WithLabelValues(VERSION, COMMIT_SHA1, BUILD_DATE, runtime.Version()).Set(1)
 
-	http.Handle(*metricPath, prometheus.Handler())
+	if !*redisOnly {
```
oliver006: Could you switch this and make it 
```
if *redisMetricsOnly {

} else {

}
```

nipuntalukdar:
Made changes as suggested in review comments.

oliver006:
Looks good, thank you!

PR #111, opened 2017-11-03
support redis-sentinel with k8s

cookingcodewithme:
Hello @oliver006,

I have updated redis_exporter for monitoring redis-sentinel on Kubernetes.

You can find changes below.
- add a new parameter as "-use-redis-sentinel" for retrieving redis master and slaves.
- add 2 kube-redis-exporter.yaml and  kube-redis-exporter-with-redis-sentinel.yaml 
- update the Dockerfile to use multi stage build.
- update README.md
- update build.sh to use docker tag instead of build image.

I haven't tested build process. But, I don't have any problems to use new Dockerfile and updated redis_exporter with K8S, Prometheus and Grafana.

Please feel free to review and let me know if you have any questions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.9 as builder
+FROM golang:1.8.3 as builder
 
 ARG VERSION=0.0.1
 WORKDIR /go/src/github.com/oliver006/redis_exporter
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,20 @@
-FROM gliderlabs/alpine
+FROM golang:1.9 as builder
+
+ARG VERSION=0.0.1
+WORKDIR /go/src/github.com/oliver006/redis_exporter
+
+COPY . .
+
+RUN go get -d -v .
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo  -ldflags "-X main.VERSION=${VERSION} -X main.BUILD_DATE=$(date +%F-%T) -X main.COMMIT_SHA1=`git rev-parse HEAD`" -o redis_exporter .
+
+
+FROM bash:latest
 MAINTAINER Oliver <oliver@21zoo.com>
 
-COPY dist/redis_exporter /bin/redis_exporter
+COPY --from=builder /go/src/github.com/oliver006/redis_exporter/redis_exporter /bin/redis_exporter
+COPY entrypoint.sh /entrypoint.sh
 
-EXPOSE     9121
-ENTRYPOINT [ "/bin/redis_exporter" ]
+RUN chmod +x /entrypoint.sh
+
+ENTRYPOINT [ "/entrypoint.sh" ]
--- a/build.sh
+++ b/build.sh
@@ -20,21 +20,16 @@ echo "Upload to Github"
 ghr -t $GITHUB_TOKEN -u $CIRCLE_PROJECT_USERNAME -r $CIRCLE_PROJECT_REPONAME --replace $CIRCLE_TAG dist/
 
 
+echo "Build Docker images"
+
 docker version
 
-gox --osarch="linux/386"   -ldflags "$GO_LDFLAGS" -output "dist/redis_exporter"
-
-echo "Build Docker images"
-docker build --rm=false -t "21zoo/redis_exporter:$CIRCLE_TAG" .
-docker build --rm=false -t "21zoo/redis_exporter:latest" .
+docker build --rm=false --build-arg VERSION=$CIRCLE_TAG -t "21zoo/redis_exporter:$CIRCLE_TAG" .
+docker tag "21zoo/redis_exporter:$CIRCLE_TAG" "oliver006/redis_exporter:$CIRCLE_TAG"
 
 docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
-docker push "21zoo/redis_exporter:latest"
-docker push "21zoo/redis_exporter:$CIRCLE_TAG"
 
-docker build --rm=false -t "oliver006/redis_exporter:$CIRCLE_TAG" .
-docker build --rm=false -t "oliver006/redis_exporter:latest" .
-docker push "oliver006/redis_exporter:latest"
+docker push "21zoo/redis_exporter:$CIRCLE_TAG"
 docker push "oliver006/redis_exporter:$CIRCLE_TAG"
 
 echo "Done"
--- a/contrib/kube-redis-exporter-with-redis-sentinel.yaml
+++ b/contrib/kube-redis-exporter-with-redis-sentinel.yaml
@@ -1,0 +1,118 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    name: sentinel
+    role: service
+  name: redis-sentinel
+spec:
+  ports:
+    - port: 26379
+      targetPort: 26379
+  selector:
+    redis-sentinel: "true"
+---
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  name: redis
+spec:
+  serviceName: redis-sentinel
+  replicas: 1
+  selector:
+    matchLabels:
+      name: sentinel
+  template:
+    metadata:
+      labels:
+        name: sentinel
+        redis-sentinel: "true"
+#      annotations:
+#        security.alpha.kubernetes.io/unsafe-sysctls: net.core.somaxconn=1024
+    spec:
+      containers:
+      - name: redis
+        image: cookingcodewithme/redis:latest
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 6379
+        resources:
+          limits:
+            cpu: "0.5"
+#        volumeMounts:
+#        - mountPath: /redis-master-data
+#          name: data
+      - name: sentinel
+        image: cookingcodewithme/redis:latest
+        imagePullPolicy: Always
+        env:
+          - name: SENTINEL
+            value: "true"
+        ports:
+          - containerPort: 26379
+#      volumes:
+#        - name: data
+#          emptyDir: {}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: default
+  name: redis-exporter
+  labels:
+    name: redis-exporter
+spec:
+  ports:
+  - name: redis-exporter
+    port: 9121
+  selector:
+    name: redis-exporter
+
+---
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  namespace: default
+  name: redis-exporter
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: redis-exporter
+    spec:
+      containers:
+        - name: redis-exporter
+          image: cookingcodewithme/redis_export:latest
+          imagePullPolicy: Always
+          resources:
+            requests:
+              cpu: "0.1"
+          ports:
+            - name: redis-exporter
+              containerPort: 9121
+          env:
+            - name: REDIS_ADDR
+              value: "redis-sentinel.default.svc.cluster.local:26379"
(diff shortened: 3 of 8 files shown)


oliver006:
Thanks for the PR - I'll try to get to reviewing it by this weekend!

oliver006:
In the meantime, maybe you could fix one issue, it CircleCI fails cause the code isn't properly formatted using `go fmt`.

Dockerfile
```diff
@@ -1,7 +1,20 @@
-FROM gliderlabs/alpine
+FROM golang:1.8.3 as builder
+
+ARG VERSION=0.0.1
+WORKDIR /go/src/github.com/oliver006/redis_exporter
+
+COPY . .
+
+RUN go get -d -v .
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo  -ldflags "-X main.VERSION=${VERSION} -X main.BUILD_DATE=$(date +%F-%T) -X main.COMMIT_SHA1=`git rev-parse HEAD`" -o redis_exporter .
```
oliver006: Question: What's the benefit of using the builder container as opposed to building the binary separately in the CircleCI runtime and the `COPY` it into the container? 

We now have the build config (LD_FLAGS, Golang version) in two places which I'd prefer to avoid.

Dockerfile
```diff
@@ -1,7 +1,20 @@
...
+
+
+FROM bash:latest
```
oliver006: Out of curiosity, why `bash` over `gliderlabs/alpine` ?

build.sh
```diff
@@ -20,21 +20,16 @@ echo "Upload to Github"
...
+echo "Build Docker images"
 
-gox --osarch="linux/386"   -ldflags "$GO_LDFLAGS" -output "dist/redis_exporter"
+docker version
 
-echo "Build Docker images"
-docker build --rm=false -t "21zoo/redis_exporter:$CIRCLE_TAG" .
-docker build --rm=false -t "21zoo/redis_exporter:latest" .
+docker build --rm=false --build-arg VERSION=$CIRCLE_TAG -t "21zoo/redis_exporter:$CIRCLE_TAG" .
+docker tag "21zoo/redis_exporter:$CIRCLE_TAG" "oliver006/redis_exporter:$CIRCLE_TAG"
```
oliver006: 👍 
That's a neat way to not build the container twice.

build.sh
```diff
@@ -20,21 +20,16 @@ echo "Upload to Github"
...
 
 docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
-docker push "21zoo/redis_exporter:latest"
```
oliver006: Question: will this still update the `latest` tag on docker hub to whatever was pushed last?

contrib/kube-redis-exporter-with-redis-sentinel.yaml
```diff
@@ -0,0 +1,118 @@
...
+    metadata:
+      labels:
+        name: sentinel
+        redis-sentinel: "true"
+#      annotations:
+#        security.alpha.kubernetes.io/unsafe-sysctls: net.core.somaxconn=1024
+    spec:
+      containers:
+      - name: redis
+        image: cookingcodewithme/redis:latest
```
oliver006: Why the `cookingcodewithme` image and not the official redis image?

contrib/kube-redis-exporter-with-redis-sentinel.yaml
```diff
@@ -0,0 +1,118 @@
...
+        ports:
+        - containerPort: 6379
+        resources:
+          limits:
+            cpu: "0.5"
+#        volumeMounts:
+#        - mountPath: /redis-master-data
+#          name: data
+      - name: sentinel
+        image: cookingcodewithme/redis:latest
```
oliver006: Ditto.

contrib/kube-redis-exporter-with-redis-sentinel.yaml
```diff
@@ -0,0 +1,118 @@
...
+#        security.alpha.kubernetes.io/unsafe-sysctls: net.core.somaxconn=1024
+    spec:
+      containers:
+      - name: redis
+        image: cookingcodewithme/redis:latest
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 6379
+        resources:
+          limits:
```
oliver006: Do you want to add a requests section for CPU as well? Maybe `0.1` or `0.05` ?

contrib/kube-redis-exporter-with-redis-sentinel.yaml
```diff
@@ -0,0 +1,118 @@
...
+      labels:
+        name: redis-exporter
+    spec:
+      containers:
+        - name: redis-exporter
+          image: cookingcodewithme/redis_export:latest
+          imagePullPolicy: Always
+          resources:
+            requests:
+              cpu: "0.1"
```
oliver006: This could probably be even less though tbh, I haven't looked at actual CPU requirements in a while.

contrib/kube-redis-exporter-with-redis-sentinel.yaml
```diff
@@ -0,0 +1,118 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: redis-exporter
+    spec:
+      containers:
+        - name: redis-exporter
+          image: cookingcodewithme/redis_export:latest
```
oliver006: cookingcodewithme --> official image?

main.go
```diff
@@ -222,3 +234,54 @@ func getCloudFoundryRedisBindings() (addrs, passwords, aliases []string) {
...
+		log.Warnln("Error while getting address of redis master", err)
+		return nil, nil, nil, err
+	} else {
+		addrs = append(addrs, master)
+
+		slaves, err := sntnl.Slaves()
+		if err != nil {
+			log.Warnln("Error while getting addresses of redis slaves", err)
+			return nil, nil, nil, err
+		} else {
```
oliver006: no need for the `else` if you return in the `if` block above.

main.go
```diff
@@ -222,3 +234,54 @@ func getCloudFoundryRedisBindings() (addrs, passwords, aliases []string) {
...
+				return nil, err
+			}
+			return c, nil
+		},
+	}
+
+	var addrs []string
+	master, err := sntnl.MasterAddr()
+	if err != nil {
+		log.Warnln("Error while getting address of redis master", err)
```
oliver006: Warnln --> Errorf/Errorln ?

main.go
```diff
@@ -222,3 +234,54 @@ func getCloudFoundryRedisBindings() (addrs, passwords, aliases []string) {
...
+		return nil, nil, nil, err
+	} else {
+		addrs = append(addrs, master)
```
oliver006: Might as well just declare addrs here with `master` as it's first element, you don't use the variable you declare in :260 outside this block.
```
  addrs := []string{ master }
```

main.go
```diff
@@ -222,3 +234,54 @@ func getCloudFoundryRedisBindings() (addrs, passwords, aliases []string) {
...
+
+		slaves, err := sntnl.Slaves()
+		if err != nil {
```
oliver006: This will make the exporter exit due to the `Fatalf` in line 90 - is that intended and desired? Is there a scenario where it makes to continue with just he master only if retrieving the slaves in line 268 failed?

main.go
```diff
@@ -222,3 +234,54 @@ func getCloudFoundryRedisBindings() (addrs, passwords, aliases []string) {
...
+			log.Warnln("Error while getting addresses of redis slaves", err)
+			return nil, nil, nil, err
+		} else {
+			for _, slave := range slaves {
+				if true == slave.Available() {
+					addrs = append(addrs, slave.Addr())
+					for len(passwords) < len(addrs) {
```
oliver006: This sets the password (and then in the next for loop) the aliases for all the discovered hosts to the same, right? 

Might move this down to right before line 284 to do it only once and also change it to `passwords = append(passwords, password)` (same for aliases) to make it more clear that they;re all set to the password/alias passed into this function.

Dockerfile
cookingcodewithme: Hi @oliver006,

Sorry for late response. 

- First of all, I had an issue with golang version on my local machine(golang 1.9.1). I had failed to compile redis_exporter on my laptop. But, multi-stage-build was a way to resolve that kinds of version or dependency issues on the machine.
- Anyone can build new update on local machine and test it easy.

Frankly, it doesn't matter whether or not using multi-stage-build. But, it is one of benefits of why we are using docker system.

FYI, you can find multi-stage-build at https://docs.docker.com/engine/userguide/eng-image/multistage-build/#use-multi-stage-builds.

Please let me know whether or not you want to keep the code.


Best regards,

Chang

contrib/kube-redis-exporter-with-redis-sentinel.yaml
cookingcodewithme: This is for testing of new docker image. I'll update the image path tonight.

oliver006: Ah ok. I'd prefer to wait till the redis image has the fix included, I'm a little hesitant to link to/endorse an image that's not the official one.

Dockerfile
oliver006: I see, thanks for explaining. Considering that the "real" builds are done in a fresh container in CircleCi I'd prefer to keep the Dockerfile the way it is right now and build on the CI host.

oliver006:
Actually, I only realized one thing just now, sorry: 

A big issue here is that the preferred way to set up monitoring and exporters in Prometheus is to have an exporter per instance of the monitored service. In this case, it would be an exporter sidecar per redis instance in your K8s cluster and then let Prometheus (via Kubernetes) handle the service discovery, not the exporter.

I think it'd still be very helpful to have example K8s config yml files to illustrate how that's done.
If you could make changes to your kubernetes yml files then they could very well serve as examples for future users that look how to set that up.
If you'd amend the kubernetes yml files to add a sidecar for the exporter  about here: 

https://github.com/oliver006/redis_exporter/pull/111/files#diff-8cdebcb89281ff21bffde987b52570dbR34

and annotate it the exporter container with `prometheus.io/scrape` then you're pretty much done.

What do you think? Does that make sense?

cookingcodewithme:
Hello @oliver006,

Thank you for informing the `prometheus.io/scrape`. I spent 3 hours to test the annotation and figured out something to discuss with you.

I think combining redis and redis_exporter is a good idea. If we use the way, the change for redis-sentinel is not required anymore. But, if you want to have a single redis_exporter without dependency with redis services. The change can be helpful and think about a standalone pod of redis_exporter.

The k8s yaml files were tested with prometheus-operator(https://github.com/coreos/prometheus-operator).
I realized that the job definition of "kubernetes-pods" is not defined on default setting of prometheus-operator. So, the annotation is not working with prometheus-operator.

I'll research more this week and trying to use the annotation(`prometheus.io/scrape`) with prometheus-operator.

Can you please advise me if I don't understand properly? 


Best regards,

Chang

oliver006:
Yeah, I think you understand correctly.

` But, if you want to have a single redis_exporter ...` - I think that's what we don't want to support, at least not build in with the sentinel service discovery in the redis_exporter.

Also think of the scenario when one of your redis sentinels goes down and is replaced. The redis_exporter will only do discovery once at start-up but then never again and if the redis instance is down and gets replaced by a new one it'll never find out about it. 
But if Prometheus queries Kubernetes for all redis_exporter instances (that are sidecars of a redis instance) then you won't have that problem.

cookingcodewithme:
Thank you for confirming my question.
I'll update the yaml file to put together redis and redis_exporter and test it.

oliver006:
Same as in #59 - make each redis instance have its own redis_exporter, that's the Prometheus way of handling this.

PR #112, opened 2017-11-10
Add slave offset metrics

S3RK:
Expose offset of connected slaves, useful to monitor replication lags

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -3,6 +3,7 @@ package exporter
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


paunin:
+

exporter/redis.go
```diff
@@ -334,6 +340,36 @@ func parseDBKeyspaceString(db string, stats string) (keysTotal float64, keysExpi
 	return
 }
 
+/*
+	slave0:ip=10.254.11.1,port=6379,state=online,offset=1751844676,lag=0
+	slave1:ip=10.254.11.2,port=6379,state=online,offset=1751844222,lag=0
+*/
+func parseConnectedSlaveString(slaveName string, slaveInfo string) (offset float64, ip string, state string, ok bool) {
+	ok = false
+	if matched, _ := regexp.MatchString(`^slave\d+`, slaveName); !matched {
```
oliver006: Use `strings.HasPrefix()` instead?

exporter/redis_test.go
```diff
@@ -552,7 +584,7 @@ func TestHTTPEndpoint(t *testing.T) {
 		`test_instance_info`,
 
 		// labels and label values
-		`addr="redis://localhost:6379"`,
+		`addr="redis://` + *redisAddr,
```

exporter/redis.go
S3RK: That was my first approach, but unfortunately, there are other parameters with this prefix.
For example
```
root@test-redis-node2-0:/data# redis-cli info | grep "^slave"
slave_repl_offset:79766288846
slave_priority:100
slave_read_only:1
```

oliver006: ahh, ok, thanks, that makes sense.

PR #116, opened 2017-11-22
Get metric for empty value of redis keySize

agileup:
current exporter do not gather metric from redis for empty key(s)
I want to monitor the size of some specific keys, even if the size is 0 such as empty list.

I'm not sure this changes are right way to go on, but please let me know if there is better way to solve my problem properly.


oliver006:
Hey there and thanks for the PR!

Here's a question: if a key wasn't found, should you explicitly set it's value to zero here: https://github.com/oliver006/redis_exporter/pull/116/files#diff-a75393c3715eaef61ada7312397db7f2R626 ?

Otherwise, if you for instance have a set with 5 elements and all 5 are removed between two scrape intervals then the stale metric value will stay around, no? 

What do you think?

oliver006:
Closing - please re-open if this is still being worked on.

Issue #138, opened 2018-03-05
[Feature request] Make redis_db_keys metric available by default with "0" value 


Blasterdick:
I have a kubernetes GKE v1.9.2 cluster with HPA scaling by the custom metrics via [Kubernetes Custom Metrics Adapter](https://github.com/DirectXMan12/k8s-prometheus-adapter). 

HPA with target Pod metric as `redis_db_keys` won't be initialized on creation if any pod would have no `redis_db_keys` metric attached. You'll get an error like: 

`failed to get pods metric value: unable to get metric redis_db_keys: unable to fetch metrics from API: the server could not find the metric redis_db_keys for pods`

So it would be awesome to have that metric be present by default with 0 value till it wouldn't be replaced by a real value. Hope, it can be done and won't break some internal logic.

brancz:
I think the crooks of the issue is here:

https://github.com/oliver006/redis_exporter/blob/fa5545286a5b4cc1849965e3193d4579aa0151f5/exporter/redis.go#L490-L492

Basically I think what's going wrong is that `ok` is false when the result is `0`, which is why the metric is not generated at all, but it should. Zeroing metrics is a best practice.

oliver006:
I think the reason `ok` is false in line 490 is not that the result is zero, the line with metrics is absent from the `CONFIG` call. Redis allows configuring the number of DBs available and doesn't show metrics for DBs that are empty.

`Zeroing metrics is a best practice.` - it depends, yes and no.
 
Not recording metrics for DBs that don't have keys and no metrics in Redis CONFIG seems OK to me.

But if this behavior is really desired then it can be done by first getting the max number of DBs (should be available via `CONFIG` ?), then collect metrics for each DB as provided by Redis and, last but not least, fill in all the blanks with zeros.
I currently don't have time to change that but happy to review and merge a PR.

brancz:
> Not recording metrics for DBs that don't have keys and no metrics in Redis CONFIG seems OK to me.

For all possible known time-series this is indeed best practice: https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics

I don't have time either, but I'm happy to advice someone who wants to take a stab at it.

oliver006:
Thanks to @s-saveljev this has been taken care of via #142

PR #142, opened 2018-03-14
set default db_keys & db_keys_expiring value to 0 for all available databases

s-saveljev:
For issue https://github.com/oliver006/redis_exporter/issues/138

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -370,14 +370,20 @@ func parseConnectedSlaveString(slaveName string, slaveInfo string) (offset float
 	return
 }
 
-func extractConfigMetrics(config []string, addr string, alias string, scrapes chan<- scrapeResult) error {
+func extractConfigMetrics(config []string, addr string, alias string, scrapes chan<- scrapeResult) (dbCount int, err error) {
 	if len(config)%2 != 0 {
-		return fmt.Errorf("invalid config: %#v", config)
+		return 0, fmt.Errorf("invalid config: %#v", config)
 	}
 
 	for pos := 0; pos < len(config)/2; pos++ {
 		strKey := config[pos*2]
 		strVal := config[pos*2+1]
+
+		if strKey == "databases" {
+			if dbCount, err = strconv.Atoi(strVal); err != nil {
+				return 0, fmt.Errorf("invalid config value for key databases: %#v", strVal)
+			}
+		}
 
 		// todo: we can add more configs to this map if there's interest
 		if !map[string]bool{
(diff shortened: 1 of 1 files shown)


oliver006:
That's awesome, thanks for the PR! I'll try to have this reviewed as soon as possible, definitely before the end of the week.

exporter/redis.go
```diff
@@ -116,6 +116,8 @@ var (
 
 	instanceInfoFields = map[string]bool{"role": true, "redis_version": true, "redis_build_id": true, "redis_mode": true, "os": true}
 	slaveInfoFields    = map[string]bool{"master_host": true, "master_port": true, "slave_read_only": true}
+
+	dataBasesCount = 0
```
oliver006: Not sure this will work if you keep per-redis-instance state in one global variable. If the exporter is used to scrape more than one instance then they'll get into each others way.

oliver006: It's a bit cumbersome but a way to solve this would be to have `extractConfigMetrics()` return `(int, error)` and you return the dbCount there so you can pass to `extractInfoMetrics()`

exporter/redis.go
```diff
@@ -524,6 +537,19 @@ func (e *Exporter) extractInfoMetrics(info, addr string, alias string, scrapes c
 		scrapes <- scrapeResult{Name: metricName, Addr: addr, Alias: alias, Value: val}
 	}
 
+	for dbIndex := 0; dbIndex < dataBasesCount; dbIndex++ {
+
+		dbName := "db" + strconv.Itoa(dbIndex)
+
+		_, exists := handledDBs[dbName]
```
oliver006: you can combine this with line 546

`if _, exists := handledDBs[dbName]; !exists {`

exporter/redis.go
s-saveljev: Thank you. Fixed it

exporter/redis.go
s-saveljev: combined

exporter/redis.go
```diff
@@ -370,15 +370,26 @@ func parseConnectedSlaveString(slaveName string, slaveInfo string) (offset float
 	return
 }
 
-func extractConfigMetrics(config []string, addr string, alias string, scrapes chan<- scrapeResult) error {
+func extractConfigMetrics(config []string, addr string, alias string, scrapes chan<- scrapeResult) (dataBasesCount int, err error) {
```
oliver006: dataBasesCount -> dbCount ?

exporter/redis.go
```diff
@@ -370,15 +370,26 @@ func parseConnectedSlaveString(slaveName string, slaveInfo string) (offset float
...
+		err = fmt.Errorf("invalid config: %#v", config)
+		return
 	}
 
 	for pos := 0; pos < len(config)/2; pos++ {
 		strKey := config[pos*2]
 		strVal := config[pos*2+1]
 
+		if strKey == "databases" {
+			var err error
```
oliver006: Last nitpick: you don't need this any longer as you now declare it up in line 373 and you can shorten the next few lines to
```
if dbCount, err = strconv.Atoi(strVal); err != nil {
  return 0, fmt.Errorf("invalid config value for key databases: %#v", strVal)
}
```

exporter/redis.go
```diff
@@ -370,15 +370,26 @@ func parseConnectedSlaveString(slaveName string, slaveInfo string) (offset float
 	return
 }
 
-func extractConfigMetrics(config []string, addr string, alias string, scrapes chan<- scrapeResult) error {
+func extractConfigMetrics(config []string, addr string, alias string, scrapes chan<- scrapeResult) (dataBasesCount int, err error) {
 	if len(config)%2 != 0 {
-		return fmt.Errorf("invalid config: %#v", config)
+		err = fmt.Errorf("invalid config: %#v", config)
+		return
```
oliver006: can you use `return 0, fmt.Errorf....` - makes it more explicit what's returned.

exporter/redis.go
s-saveljev: renamed

exporter/redis.go
s-saveljev: used your recommendation & shortened

oliver006:
Nice work, this looks good to go!

s-saveljev:
Good news! Thanks!

oliver006:
And released as https://github.com/oliver006/redis_exporter/releases/tag/v0.16.0

PR #148, opened 2018-03-22
Fix dbCount in cluster mode

oliver006:
What do you think @ccs018 ?


ccs018:
LGTM

ccs018:
Forgot to say:  Thanks for awesome turn around time!

oliver006:
Thanks for reviewing!

PR #149, opened 2018-03-23
Avoid using KEYS.

Tinche:
Hi,

here's a version using TYPE instead of KEYS.

We've lost the wildcard feature, but since it was neither documented (I think) nor feasible on large databases, c'est la vie. I changed some tests and added a test for lists.

I didn't know how to handle PFCOUNT, so I'll leave that up to you.

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -671,34 +671,38 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 			continue
 		}
 
-		obtainedKeys := []string{}
-		if tempVal, err := redis.Strings(doRedisCmd(c, "KEYS", k.key)); err == nil && tempVal != nil {
-			for _, tempKey := range tempVal {
-				log.Debugf("Append result: %s", tempKey)
-				obtainedKeys = append(obtainedKeys, tempKey)
-			}
-		}
+		if keyType, err := redis.String(c.Do("TYPE", k.key)); err == nil && keyType != "" {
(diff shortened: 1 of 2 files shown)


oliver006:
Thank you for the PR - I'll try to have a look over the weekend.

exporter/redis.go
```diff
@@ -671,34 +671,38 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
...
-				"STRLEN",
-			} {
-				if tempVal, err := doRedisCmd(c, op, key); err == nil && tempVal != nil {
+			keyLabel := k.key
+
+			switch keyType {
+			case "string":
+				if tempVal, err := c.Do("GET", k.key); err == nil && tempVal != nil {
+					if val, err := strconv.ParseFloat(fmt.Sprintf("%s", tempVal), 64); err == nil {
+						e.keyValues.WithLabelValues(addr, e.redis.Aliases[idx], dbLabel, keyLabel).Set(val)
```

oliver006:
As for the `wildcard feature` - not sure if we can lose that, it was added explicitly in https://github.com/oliver006/redis_exporter/pull/83 so I'd assume that's used and we can't really cut that (and yeah, that should be documented).
How about running the supplied keys through `KEYS` only when a key with `*` was provided and we need to expand the list of keys?

Tinche:
Sounds reasonable, I'll try getting around to incorporating your feedback soon!

oliver006:
Closing, this has been addressed via #172

PR #151, opened 2018-03-27
Adding support for Openshift

ivanovaleksandar:
The idea is to add quick way of deploying the exporter on Openshift. 

Any suggestions on what to edit or changed are welcomed.


oliver006:
This is great, thank you!
Could you move the openshift template to the `contrib/` folder and rename it to `openshift-template.yaml`? In that way we can keep all these files in the same folder.

ivanovaleksandar:
Done! :)

oliver006:
Thanks!

PR #158, opened 2018-05-09
Improve Dockerfile

murata100:


oliver006:
Thanks for the PR!

I like the idea of using a build image but have a look here: 
https://github.com/oliver006/redis_exporter/blob/master/build.sh#L5 
the go build command currently gets a bunch of parameters - can you make sure they're used in the Dockerfile build as well?

oliver006:
Closing..

PR #159, opened 2018-05-09
Add `maxclients` to config metrics

fajpunk:
I would like to be able to monitor connected clients vs. the configured `maxclients` setting.


oliver006:
Nice, thank you!

PR #165, opened 2018-06-06
Optional Redis Lua script for metric collection

wojas:
Allow an optional Redis Lua script to collect extra metrics.

This can be enabled with a new `-script` flag. An example script is provided in contrib.

I needed a way to export metrics that are collected within a Redis hash and `-check-keys` does not support this. Adding support for [Redis Lua scripts](https://redis.io/commands/eval) allows a user to collect whatever metrics they want from Redis without writing a custom exporter.

The script metrics end up in a new `redis_script_value` vector.

### DIFF
--- a/exporter/redis_test.go
+++ b/exporter/redis_test.go
@@ -503,6 +503,36 @@ func TestKeyValuesAndSizesWildcard(t *testing.T) {
 	}
 }
 
+func TestScript(t *testing.T) {
+
+	e, _ := NewRedisExporter(defaultRedisHost, "test", "")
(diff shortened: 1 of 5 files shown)


oliver006:
❤️ 

Thanks for the PR! I'll try to get this reviewed later today or at least in the next couple of days.

exporter/redis_test.go
```diff
@@ -503,6 +503,36 @@ func TestKeyValuesAndSizesWildcard(t *testing.T) {
 	}
 }
 
+func TestScript(t *testing.T) {
```

wojas:
Thanks for the fast merge and release!

PR #166, opened 2018-06-12
add slow logs count

EriconYu:
add slow logs count .


oliver006:
Thanks for the PR!

exporter/redis.go
```diff
@@ -182,6 +182,11 @@ func (e *Exporter) initGauges() {
 		Name:      "command_call_duration_seconds_sum",
 		Help:      "Total amount of time in seconds spent per command",
 	}, []string{"addr", "alias", "cmd"})
+	e.metrics["slowlog_len"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: e.namespace,
+		Name:      "slowlog_len_sum",
```
oliver006: `slowlog_len_sum` --> `slowlog_length`

exporter/redis.go
```diff
@@ -182,6 +182,11 @@ func (e *Exporter) initGauges() {
...
+		Help:      "Total slowlog",
+	}, []string{"addr", "alias", "cmd"})
```
oliver006: The `cmd` label isn't needed. See also below.

exporter/redis.go
```diff
@@ -732,6 +737,12 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		}
 	}
 
+	if reply, err := c.Do("SLOWLOG", "LEN"); err == nil {
+		e.metricsMtx.RLock()
+		e.metrics["slowlog_len"].WithLabelValues(addr, e.redis.Aliases[idx], "slowlog len").Set(float64(reply.(int64)))
```
oliver006: You can remove `"slowlog len"` - the cmd label doesn't really add anything useful to the metric.

oliver006: Also, instead of `float64(reply.(int64)` can you please use [ParseFloat](https://golang.org/pkg/strconv/#ParseFloat) ? There's a couple of examples in the code already where that's being used.

exporter/redis_test.go
```diff
@@ -326,6 +326,7 @@ func TestExporterMetrics(t *testing.T) {
 		"loading_dump_file", // testing renames
 		"config_maxmemory",  // testing config extraction
 		"config_maxclients", // testing config extraction
+		"slowlog_len",
```

exporter/redis.go
EriconYu: just put it for ... funny😊

exporter/redis.go
EriconYu: and slowlog_count?Emmm.....I don't know which is the best....

exporter/redis.go
EriconYu: EngHeng~That's a good suggestoin~

EriconYu: But the value whick the cmd 'slow len' returned is interface{},it is int64 actually.So i think that float64(reply.(int64) maybe is a better way.

exporter/redis.go
oliver006: :-)

exporter/redis.go
oliver006: Let's go with `slowlog_length` for now.

exporter/redis.go
oliver006: Ah yes, good catch, didn't see it's of type interface{} that can be cast to int64.

exporter/redis.go
EriconYu: OKay~

EriconYu: All fixed~ Thx your help~

PR #168, opened 2018-06-20
Add start_time_seconds

xginn8:
start_time is a more idiomatic way to store the process uptime,
and is what is provided by the Prometheus library itself.

Since this is a breaking change, I am happy to add this as a second
metric if necessary

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -54,7 +54,8 @@ type scrapeResult struct {
 var (
 	metricMap = map[string]string{
 		// # Server
-		"process_id": "process_id",
+		"uptime_in_seconds": "uptime_in_seconds",
+		"process_id":        "process_id",
 
 		// # Clients
 		"connected_clients":          "connected_clients",
@@ -475,7 +476,6 @@ func (e *Exporter) extractInfoMetrics(info, addr string, alias string, scrapes c
 				e.metrics["start_time_seconds"].WithLabelValues(addr, alias).Set(float64(time.Now().Unix()) - uptime)
 				e.metricsMtx.RUnlock()
 			}
-			continue
 		}
 
 		if slaveOffset, slaveIp, slaveState, ok := parseConnectedSlaveString(fieldKey, fieldValue); ok {
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -54,8 +54,7 @@ type scrapeResult struct {
 var (
 	metricMap = map[string]string{
 		// # Server
-		"uptime_in_seconds": "uptime_in_seconds",
-		"process_id":        "process_id",
+		"process_id": "process_id",
 
 		// # Clients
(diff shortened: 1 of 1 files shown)


ccs018:
Since this would be a breaking change either adding this as a second metric or upreving the major version of the release would be needed.

To me, "start time" implies a specific timestamp whereas "uptime" implies a duration.  The "uptime" terminology is rather consistent with other unix/linux references.  Since the Redis metric "uptime_in_seconds" is describing a duration and there is also an "uptime_in_days" metric and renaming a metric would confuse users, my opinion is to reject this PR.

xginn8:
It's worth noting that the Prometheus best practices for instrumentation explicitly recommend exporting a timestamp over the time since an event: https://prometheus.io/docs/practices/instrumentation/#timestamps-not-time-since

ccs018:
That's fine, but simply renaming the Redis metric "uptime_in_seconds" to "start_time_seconds" does NOT align with the Prometheus best practice.  Th metric is still reporting a duration, not a timestamp.  If you want a metric reporting the time at which Redis started, then you either need to 1) submit a PR to the Redis repo to add such a metric to Redis itself or 2) have redis_exporter add a metric which it derives from the "uptime_in_seconds" by getting the current time and subtracting "uptime_in_seconds".

ccs018:
Got it - I skimmed the PR too quickly before.

I think it would be better if the PR was purely additive - that is, add your new start_time_seconds, but leave uptime_in_seconds as is.  Taking something away that people may be using is generally not a good idea.

exporter/redis.go
```diff
@@ -133,6 +133,11 @@ func (e *Exporter) initGauges() {
 		Name:      "slave_info",
 		Help:      "Information about the Redis slave",
 	}, []string{"addr", "alias", "master_host", "master_port", "read_only"})
+	e.metrics["start_time_seconds"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: e.namespace,
+		Name:      "start_time_seconds",
+		Help:      "Start time of the Redis instance since unix epoch in seconds.",
```
ccs018: "since" implies a duration.  Suggest changing this to: "Start time of the Redis instance in unix epoch seconds."

xginn8: the epoch is always just seconds since 0, or 00:00:00 UTC, Thursday, 1 January 1970, see https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go#L82

oliver006:
@xginn8 - thanks for the PR! I'll try to catch up with the conversation and review it by the end of the week.
@ccs018 - thanks for jumping in and providing early feedback on the PR!

oliver006:
Thanks for making this an additive change @xginn8 . I agree, we shouldn't be breaking existing functionality.

This looks good now!

Issue #167, opened 2018-06-19
No metrics


prdev1:
Dears,
I've deployed redis exporter in cloud foundry with `cf push <app_name> -f <manifest_path>`.
I guess I need to configure the list of redis hosts somewhere in the exporter app. May I know where should I configure this ? With just cf push, redis exporter doesn't scrape any metrics form the Redis VMs. 

Regards

oliver006:
I'm not really familiar with how service discovery in CloudFoundry works but maybe @ihrigb knows more, he added Cloud Foundry support.

ihrigb:
@prdev1
I am currently without proper internet. I will write a proper answer tomorrow.

ihrigb:
Alright, service discovery in CF happens here: https://github.com/oliver006/redis_exporter/blob/master/exporter/discovery.go#L81
We get all bound services with tag "redis" and then read connection information from the "credentials" json object.

So you do not need to define any list of redis hosts, but just bind your redis service instances to the exporter application.

prdev1:
@ihrigb 
I created a redis service & bound it to the redis exporter application. And then the application crashes ! 


form the logs : 

```
   2018-06-20T14:02:27.16+0200 [APP/PROC/WEB/0] ERR time="2018-06-20T12:02:27Z" level=info msg="Redis Metrics Exporter <<< filled in by build >>>    build date: <<< filled in by build >>>    sha1: <<< filled in by build >>>    Go: go1.8.7"
   2018-06-20T14:02:27.16+0200 [APP/PROC/WEB/0] ERR panic: interface conversion: interface {} is nil, not string
   2018-06-20T14:02:27.16+0200 [APP/PROC/WEB/0] ERR goroutine 1 [running]:
   2018-06-20T14:02:27.16+0200 [APP/PROC/WEB/0] ERR github.com/oliver006/redis_exporter/exporter.GetCloudFoundryRedisBindings(0x563f00000004, 0x41, 0xc42004feb0, 0x4, 0x4, 0xc420136638, 0xc42004fdc0, 0x0, 0x47084fd830fe5e3b)
   2018-06-20T14:02:27.16+0200 [APP/PROC/WEB/0] ERR 	/tmp/gobuildpack.gopath661670186/.go/src/github.com/oliver006/redis_exporter/exporter/discovery.go:89 +0x89e
   2018-06-20T14:02:27.16+0200 [APP/PROC/WEB/0] ERR main.main()
[...]

PR #169, opened 2018-06-20
Add alternative credentials claim for host.

ihrigb:
Adresses #167


oliver006:
Thanks for the PR! I'll try to catch up with the issue in #167 and the code change here by the end of week and review it.

oliver006:
Looks good!

PR #172, opened 2018-06-26
Avoid using KEYS, continued

fajpunk:
Continues the work that @Tinche did in #149.

* Add an arg, `check-single-keys` to get size and value with `TYPE`
* Modify the behavior of `check-keys`
  * Get keys from globs with `SCAN` rather than `KEYS`
  * If a key doesn't have any glob characters in it, just add it to the
    list of things to check with `TYPE`

I added an argument rather than decide `SCAN` vs `TYPE` based on the
presence of glob characters because, technically, redis key names can
contain glob characters, according to https://redis.io/topics/data-types-intro:

    Redis keys are binary safe, this means that you can use any binary
    sequence as a key, from a string like "foo" to the content of a JPEG
    file. The empty string is also a valid key.

The only way to be able to check for key names with glob characters in
them is by being able to explicitly specify keys that you don't want to
glob for.

Some questions:

1. Using `SCAN` rather than `KEYS` is probably what we want if people want to use globs, but is it possible that for some databases, the implementation here will perform worse than `KEYS` (even though it will be less likely to lock up the database)?  Are you OK with this?

1. I wasn't sure what you meant by the comment to make a `RedisExporter` struct if an argument gets added to `NewRedisExporter`.  I added an argument, I'm glad to do whatever is meant by that comment, I just need a little more clarification :)

1. I sort of took an alternate strategy with the tests, setting up specific keys for specific tests, rather than adding to the `setupDBKeys` function.  Let me know if you'd rather I add stuff to `setupDBKeys`.

This ended up being a lot more changes than I thought it would, I'm glad to incorporate any feedback, and my feelings won't be hurt at all if this is rejected outright :)


exporter/redis.go
```diff
@@ -189,9 +196,39 @@ func (e *Exporter) initGauges() {
...
+			return keys, err
+		}
+
+		keys = append(keys, dbKeyPair{db, key})
+	}
+	return keys, err
+}
+
 // NewRedisExporter returns a new exporter of Redis metrics.
 // note to self: next time we add an argument, instead add a RedisExporter struct
```
fajpunk: Not sure exactly what you mean by this, glad to do whatever with some more direction.

exporter/redis.go
```diff
@@ -599,6 +629,98 @@ func doRedisCmd(c redis.Conn, cmd string, args ...interface{}) (reply interface{
...
+		return info, err
+	}
+
+	info.keyType = keyType
+
+	switch keyType {
+	case "none":
+		return info, errNotFound
+	case "string":
+		if size, err := redis.Int64(c.Do("PFCOUNT", key)); err == nil {
```
fajpunk: I figured there's no need to retrieve the key and check the first few chars before calling `PFCOUNT`; just try it and use it if it doesn't fail.

exporter/redis.go
```diff
@@ -189,9 +196,39 @@ func (e *Exporter) initGauges() {
...
+		return keys, err
+	}
+	for _, k := range strings.Split(keysArgString, ",") {
+		db := "0"
+		key := ""
+		frags := strings.Split(k, "=")
+		switch len(frags) {
+		case 1:
+			db = "0"
+			key, err = url.QueryUnescape(strings.TrimSpace(frags[0]))
```
fajpunk: Out of curiosity, why are these being `QueryUnescape`d?

oliver006:
Thanks for the PR!
I'll try to find some time this week to review it and also get back to you regarding your questions.

exporter/redis.go
```diff
@@ -599,6 +629,98 @@ func doRedisCmd(c redis.Conn, cmd string, args ...interface{}) (reply interface{
...
+func scanForKeys(c redis.Conn, pattern string) ([]string, error) {
+	iter := 0
+	keys := []string{}
+	for {
+		arr, err := redis.Values(c.Do("SCAN", iter, "MATCH", pattern))
+		if err != nil {
+			return keys, fmt.Errorf("error retrieving '%s' keys", pattern)
+		}
+
+		iter, _ = redis.Int(arr[0], nil)
```
oliver006:
```		
		k, _ := redis.Strings(arr[1], nil)
		keys = append(keys, k...)

		if iter, _ = redis.Int(arr[0], nil); iter == 0 {
		    break
		}
...
```

oliver006: And should you check first that `len(arr) == 2` ?

exporter/redis.go
```diff
@@ -599,6 +629,98 @@ func doRedisCmd(c redis.Conn, cmd string, args ...interface{}) (reply interface{
...
+func getKeysFromPatterns(c redis.Conn, keys []dbKeyPair) (expandedKeys []dbKeyPair, err error) {
+	expandedKeys = []dbKeyPair{}
+	for _, k := range keys {
+		if regexp.MustCompile(`[\?\*\[\]\^]+`).MatchString(k.key) {
+			_, err := c.Do("SELECT", k.db)
+			if err != nil {
+				return expandedKeys, err
+			}
+			keyNames, err := scanForKeys(c, k.key)
+			if err != nil {
```
oliver006: why abort here and not check the remaining values in keys?

exporter/redis.go
```diff
@@ -684,42 +806,39 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		}
 	}
 
+	log.Debugf("e.singleKeys: %#v", e.singleKeys)
+	allKeys := append([]dbKeyPair{}, e.singleKeys...)
+
 	log.Debugf("e.keys: %#v", e.keys)
-	for _, k := range e.keys {
+	scannedKeys, err := getKeysFromPatterns(c, e.keys)
+	if err != nil {
```
oliver006: if an error happens here, is there a point to continue with `scannedKeys` ?

exporter/redis.go
fajpunk: Good point; changed.

exporter/redis.go
fajpunk: Nope; changed.

fajpunk:
Ok, I addressed that round of feedback.  Sorry it took me so long to get to it!

main.go
```diff
@@ -19,7 +19,8 @@ var (
 	redisPassword    = flag.String("redis.password", getEnv("REDIS_PASSWORD", ""), "Password for one or more redis nodes, separated by separator")
 	redisAlias       = flag.String("redis.alias", getEnv("REDIS_ALIAS", ""), "Redis instance alias for one or more redis nodes, separated by separator")
 	namespace        = flag.String("namespace", "redis", "Namespace for metrics")
-	checkKeys        = flag.String("check-keys", "", "Comma separated list of keys to export value and length/size")
+	checkKeys        = flag.String("check-keys", "", "Comma separated list of key-patterns to export value and length/size, searched for with SCAN")
+	checkSingleKeys  = flag.String("check-single-keys", "", "Comma separated list of single keys to export value and length/size")
```
oliver006: Can you please add the new flag to the README?

fajpunk: Oops, sure thing.  Updated.

oliver006: Thx!

README.md
```diff
@@ -89,7 +89,8 @@ Name               | Description
 -------------------|------------
 debug              | Verbose debug output
 log-format         | Log format, valid options are `txt` (default) and `json`.
-check-keys         | Comma separated list of keys to export value and length/size, eg: `db3=user_count` will export key `user_count` from db `3`. db defaults to `0` if omitted.
+check-keys         | Comma separated list of key patterns to export value and length/size, eg: `db3=user_count` will export key `user_count` from db `3`. db defaults to `0` if omitted. The key patterns specified with this flag will be found using [SCAN](https://redis.io/commands/scan).  Use this option if you need glob pattern matching; `check-keys` is faster for non-pattern keys.
```
oliver006: Last part of this line, instead of `check-keys is faster for non-pattern keys.` - shouldn't it be `check-single-keys is faster for non-pattern keys.` ?

fajpunk: Sorry, I just saw this.  You're correct; fixed.

PR #175, opened 2018-07-02
add empty help text to be compatible with prometheus >= 2.3.1

carlo-blohm:


oliver006:
Thanks for the PR!
I'll try to have a look at it by the end of the week the latest and will get back to you.

exporter/redis.go
```diff
@@ -787,6 +787,7 @@ func (e *Exporter) setMetrics(scrapes <-chan scrapeResult) {
 			e.metrics[name] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 				Namespace: e.namespace,
 				Name:      name,
+				Help:      "empty", // needs to be set for prometheus >= 2.3.1
```
oliver006: Maybe make this `name + " metric"` so they're not all the same?

carlo-blohm:
For now there is no real value to the HELP text, but I made it uniqe.

PR #179, opened 2018-07-27
Pull request for issue "All configs with env variables"

dkanchev:
Added getEnvBool helper function to parse env variables of type boolean. Now all config parameters can be set as env variables. Things to consider:

1. The new helper function uses strconv because strings are converted to boolean via ParseBool. This accepts variations of true and false as well as 0 and 1. If there is a better way to achieve this I am all ears :)

2. Some of the names of env variables are too generic in my opinion and could cause issues on systems where other programs have set env variables with the same names. For example "DEBUG" and "VERSION" are two such env variables. Still, I decided to keep the format which has already been used and just capitalized the names the same way REDIS_FILE and REDIS_PASSWORD have been changed. We might want to change those env variable names.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"runtime"
+        "strconv"
 
 	"github.com/oliver006/redis_exporter/exporter"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


dkanchev:
Thanks to @jivkoa for pointing out that the helper function can be rewritten in this way:

```go
func getEnvBool(key string) (envValBool bool) {
	if envVal, ok := os.LookupEnv(key); ok {
		envValBool, _ = strconv.ParseBool(envVal)
	}
	return
}
```

oliver006:
Thanks for the PR!
I'll try to get it reviewed over the weekend but in the meantime, can you run `go fmt` on the `main.go` file so the tests pass? Thx!

main.go
```diff
@@ -18,18 +19,18 @@ var (
...
+	checkSingleKeys  = flag.String("check-single-keys", getEnv("CHECK_SINGLE_KEYS", ""), "Comma separated list of single keys to export value and length/size")
+	scriptPath       = flag.String("script", getEnv("SCRIPT", ""), "Path to Lua Redis script for collecting extra metrics")
+	separator        = flag.String("separator", getEnv("SEPARATOR", ","), "separator used to split redis.addr, redis.password and redis.alias into several elements.")
+	listenAddress    = flag.String("web.listen-address", getEnv("WEB_LISTEN_ADDRESS", ":9121"), "Address to listen on for web interface and telemetry.")
+	metricPath       = flag.String("web.telemetry-path", getEnv("WEB_TELEMETRY_PATH", "/metrics"), "Path under which to expose metrics.")
+	isDebug          = flag.Bool("debug", getEnvBool("DEBUG"), "Output verbose debug information")
+	logFormat        = flag.String("log-format", getEnv("LOG_FORMAT", "txt"), "Log format, valid options are txt and json")
+	showVersion      = flag.Bool("version", getEnvBool("VERSION"), "Show version information and exit")
+	useCfBindings    = flag.Bool("use-cf-bindings", getEnvBool("USE-CF-BINDINGS"), "Use Cloud Foundry service bindings")
```
oliver006: Let's change this to `REDIS_EXPORTER_USE_CF_BINDINGS`.

mjtrangoni:
Hi @dkanchev @oliver006,

You could maybe switch to [kingpin](https://github.com/alecthomas/kingpin), as this is all provided out-of-the-box. See [here](https://github.com/alecthomas/kingpin/blob/1d8c6233d4eb2cd727aec70b32949eeb814a12d7/args_test.go#L80).

BTW, the prometheus project self use kingpin instead of golang`s flag.

oliver006:
kingpin looks pretty cool, thanks for pointing that out!

That said, I have no preference for one flag package over another as long as the change is backward incompatible, e.g. the flags or (existing) env variables change.

main.go
```diff
@@ -18,18 +19,18 @@ var (
...
+	checkSingleKeys  = flag.String("check-single-keys", getEnv("REDIS_EXPORTER_CHECK_SINGLE_KEYS", ""), "Comma separated list of single keys to export value and length/size")
+	scriptPath       = flag.String("script", getEnv("REDIS_EXPORTER_SCRIPT", ""), "Path to Lua Redis script for collecting extra metrics")
+	separator        = flag.String("separator", getEnv("REDIS_EXPORTER_SEPARATOR", ","), "separator used to split redis.addr, redis.password and redis.alias into several elements.")
+	listenAddress    = flag.String("web.listen-address", getEnv("REDIS_EXPORTER_WEB_LISTEN_ADDRESS", ":9121"), "Address to listen on for web interface and telemetry.")
+	metricPath       = flag.String("web.telemetry-path", getEnv("REDIS_EXPORTER_WEB_TELEMETRY_PATH", "/metrics"), "Path under which to expose metrics.")
+	isDebug          = flag.Bool("debug", getEnvBool("REDIS_EXPORTER_DEBUG"), "Output verbose debug information")
+	logFormat        = flag.String("log-format", getEnv("REDIS_EXPORTER_LOG_FORMAT", "txt"), "Log format, valid options are txt and json")
+	showVersion      = flag.Bool("version", getEnvBool("REDIS_EXPORTER_VERSION"), "Show version information and exit")
```
oliver006: I don't think this one makes sense, can you remove it please?

dkanchev: Done :)

dkanchev:
@oliver006 this has been changed :)

PR #195, opened 2018-10-25
Always return a redis_key_size metric

victortrac:
We use this redis_size_size to do some autoscaling of pods, but the HPA complains when the metric is missing and doesn't know what to do when the queue depth is drained too quickly (it can't scale down properly). So this change is intended to always return a 0 value when keys are all zero-sized.

Additionally, I ran `dep ensure -update`.


victortrac:
We are running redis_exporter with the `--check-keys *` flag, since we only care about the aggregate sum of key lengths instead of a particular key (we also don't know the names of keys ahead of time). 

But yes, you are right - if we listed each key using `check-single-keys`, then it seems like this PR wouldn't be needed.  

Do you have any other suggestions?

oliver006:
One thing I could think of as a workaround is this: if the problem is that the timeseries you use for scaling sometimes just disappears, then maybe try to prevent this from happening via the query you use. You could try using "max(timeseries, 0)` or synthetically generating an "always zero" timeseries you can use for that. But hard to say without knowing more about how the autoscaler works.

victortrac:
Using `max(timeseries, 0)` was the first idea, but this doesn't seem possible with our metrics system at the moment (stackdriver monitoring). What's the issue that you have with emitting a value of 0 from redis-server with db and key == nil?

oliver006:
`...doesn't seem possible with our metrics system at...` - you ingest the Prometheus metrics into Stackdriver? Is that why it's not possible? Did you try recording a new timeseries and use that in Stackdriver?  (I'm still not sure I full understand what your stack/chain of systems looks like from redis_exporter onwards)

`...the issue that you have with...` - it seems like a hack to work around a shortcoming in a different system and I'd prefer to find a better way to address the issue than introducing a timeseries that doesn't exist.

victortrac:
> Did you try recording a new timeseries and use that in Stackdriver?

Well, we're using redis_exporter to get a redis queue depth into stackdriver. It doesn't seem ideal to have _something else_ send the 0 metric on this timeseries when the redis_exporter can do it with slight modifications. 

I understand your hesitation, but IMO there are valid use cases when a prometheus exporter should emit a 0-valued scalar without having to use something like `max(timeseries, 0)` at the db (for instance, it denotes that the exporter and redis are actually running).  `0` for a "nil" db and key seems the cleanest to me, but if you don't agree, I'll close this PR and maintain a fork for our unique use-case.

oliver006:
You're right, there are use cases when Prometheus should emit a 0-valued scalar to express the absence of something but I'm still not convinced that making up the `nil` db & key are the right way to go about this.

> have _something else_ send the 0 metric

Sorry if I was unclear. Prometheus has a feature called ["recording rules"](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/
) - it allows you to generate a new time series inside Prometheus that is the result of a query.
What I meant to suggest was that you could use that to record a new series e.g. called `victor_key_size` and have that be the sum of all the `redis_key_size` metrics and floored at zero.
Now you can export that metric to Stackdriver and use it for autoscaling. 

If this isn't working for you then using your own fork is probably way to go.

victortrac:
We don't actually run prometheus. Stackdriver monitoring supports a scraper that understands emitted prometheus-style metrics but the underlying metrics service is stackdriver, not prometheus. 

In any case, I'll maintain the fork for our own needs.  

Thanks.

PR #205, opened 2018-12-09
Fix linter issues

mjtrangoni:
Hi @oliver006,

I fixed here some trivial linter issues. PTAL if I should be changing or dropping something.

### DIFF
--- a/exporter/discovery.go
+++ b/exporter/discovery.go
@@ -45,8 +45,7 @@ func LoadRedisFile(fileName string) ([]string, []string, []string, error) {
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	err = file.Close()
-	if err != nil {
+	if err = file.Close(); err != nil {
 		return nil, nil, nil, err
 	}
 	// For each line, test if it contains an optional password and alias and provide them,
--- a/exporter/discovery.go
+++ b/exporter/discovery.go
@@ -2,8 +2,8 @@ package exporter
 
 import (
 	"encoding/csv"
-	"filepath"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/cloudfoundry-community/go-cfenv"
--- a/exporter/discovery.go
+++ b/exporter/discovery.go
@@ -2,6 +2,7 @@ package exporter
 
 import (
 	"encoding/csv"
+	"filepath"
 	"os"
 	"strings"
 
@@ -34,7 +35,7 @@ func LoadRedisFile(fileName string) ([]string, []string, []string, error) {
 	var addrs []string
 	var passwords []string
 	var aliases []string
-	file, err := os.Open(fileName)
+	file, err := os.Open(filepath.Clean(fileName))
 	if err != nil {
 		return nil, nil, nil, err
 	}
@@ -44,7 +45,10 @@ func LoadRedisFile(fileName string) ([]string, []string, []string, error) {
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	file.Close()
+	err = file.Close()
+	if err != nil {
+		return nil, nil, nil, err
+	}
 	// For each line, test if it contains an optional password and alias and provide them,
 	// else give them empty strings
 	for _, record := range records {
--- a/exporter/discovery.go
+++ b/exporter/discovery.go
@@ -9,7 +9,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// loadRedisArgs loads the configuration for which redis hosts to monitor from either
+// LoadRedisArgs loads the configuration for which redis hosts to monitor from either
 // the environment or as passed from program arguments. Returns the list of host addrs,
 // passwords, and their aliases.
 func LoadRedisArgs(addr, password, alias, separator string) ([]string, []string, []string) {
@@ -28,7 +28,7 @@ func LoadRedisArgs(addr, password, alias, separator string) ([]string, []string,
 	return addrs, passwords, aliases
 }
 
-// loadRedisFile opens the specified file and loads the configuration for which redis
+// LoadRedisFile opens the specified file and loads the configuration for which redis
 // hosts to monitor. Returns the list of hosts addrs, passwords, and their aliases.
 func LoadRedisFile(fileName string) ([]string, []string, []string, error) {
 	var addrs []string
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -521,12 +521,12 @@ func (e *Exporter) extractInfoMetrics(info, addr string, alias string, scrapes c
 			}
 		}
 
-		if slaveOffset, slaveIp, slaveState, ok := parseConnectedSlaveString(fieldKey, fieldValue); ok {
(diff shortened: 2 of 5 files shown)


exporter/redis.go
```diff
@@ -659,6 +659,8 @@ var errNotFound = errors.New("key not found")
 // getKeyInfo takes a key and returns the type, and the size or length of the value stored at that key.
 func getKeyInfo(c redis.Conn, key string) (info keyInfo, err error) {
 
+	var size int64
```
oliver006: What's the reason for pulling this out?

exporter/redis.go
```diff
@@ -711,27 +713,34 @@ func scanForKeys(c redis.Conn, pattern string) ([]string, error) {
...
+			log.Debugln(err)
+		}
 	}
 
 	return keys, nil
 }
 
 // getKeysFromPatterns does a SCAN for a key if the key contains pattern characters
 func getKeysFromPatterns(c redis.Conn, keys []dbKeyPair) (expandedKeys []dbKeyPair, err error) {
+	var keyNames []string
```
oliver006: Same, why move this up?

exporter/redis.go
```diff
@@ -747,6 +756,7 @@ func getKeysFromPatterns(c redis.Conn, keys []dbKeyPair) (expandedKeys []dbKeyPa
 }
 
 func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx int) error {
+	var config []string
```
oliver006: Same, why move this up?

exporter/redis.go
```diff
@@ -799,7 +808,8 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 	isClusterEnabled := strings.Contains(infoAll, "cluster_enabled:1")
 
 	if isClusterEnabled {
-		if clusterInfo, err := redis.String(doRedisCmd(c, "CLUSTER", "INFO")); err == nil {
+		var clusterInfo string
```
oliver006: Same, why pull this out?

main.go
```diff
@@ -113,8 +116,14 @@ func main() {
 
 	if *redisMetricsOnly {
 		registry := prometheus.NewRegistry()
-		registry.Register(exp)
-		registry.Register(buildInfo)
+		err = registry.Register(exp)
+		if err != nil {
```
oliver006: --> `if err := registry.Register(exp); err != nil { `

main.go
```diff
@@ -113,8 +116,14 @@ func main() {
...
+			log.Error(err)
+		}
+		err = registry.Register(buildInfo)
+		if err != nil {
```
oliver006: same

main.go
```diff
@@ -98,7 +100,8 @@ func main() {
 	}
 
 	if *scriptPath != "" {
-		script, err := ioutil.ReadFile(*scriptPath)
+		var script []byte
```
oliver006: why move this out?

exporter/discovery.go
```diff
@@ -44,7 +45,10 @@ func LoadRedisFile(fileName string) ([]string, []string, []string, error) {
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	file.Close()
+	err = file.Close()
+	if err != nil {
```
oliver006: `if err := file.Close(); err != nil {`

exporter/redis.go
mjtrangoni: This is reported by `vetshadow`, a check for shadowed variables. The problem is, as the commit message said, the `err` variable shadowing. 

As I said, if you do not agree, I can drop it.

exporter/redis.go
mjtrangoni: same here

exporter/redis.go
mjtrangoni: same here

exporter/redis.go
mjtrangoni: same here, vetshadow

main.go
mjtrangoni: same here, vetshadow

exporter/redis.go
```diff
@@ -711,27 +713,34 @@ func scanForKeys(c redis.Conn, pattern string) ([]string, error) {
 			return keys, fmt.Errorf("invalid response from SCAN for pattern: %s", pattern)
 		}
 
-		k, _ := redis.Strings(arr[1], nil)
+		k, err := redis.Strings(arr[1], nil)
+		if err != nil {
```
oliver006: I don't think the is is necessary but if you really want to check for and  error then you should take appropriate action if there is one, e.g not append k to keys if err != nil.

exporter/redis.go
```diff
@@ -711,27 +713,34 @@ func scanForKeys(c redis.Conn, pattern string) ([]string, error) {
...
+			log.Debugln(err)
+		}
 		keys = append(keys, k...)
 
-		if iter, _ = redis.Int(arr[0], nil); iter == 0 {
+		if iter, err = redis.Int(arr[0], nil); iter == 0 {
```
oliver006: I think this change is not need nd you should remove 725-727\. you can ignore err here, iter will be zero even if an error is returned and there's no value in logging err, we can simply break.

exporter/redis.go
oliver006: I'd remove the named return variables, make 736 an instantiation (`:=`) and then the shadowing issue should be resolved.

exporter/redis.go
oliver006: I'd prefer to not move config up but keep it in the if as is right now, the shadowing isn't a problem here.

exporter/redis.go
oliver006: same here

main.go
oliver006: let's keep this is as is, the shadowing is not going to cause issues.

main.go
```diff
@@ -113,8 +116,12 @@ func main() {
 
 	if *redisMetricsOnly {
 		registry := prometheus.NewRegistry()
-		registry.Register(exp)
-		registry.Register(buildInfo)
+		if err = registry.Register(exp); err != nil {
+			log.Error(err)
```
oliver006: Let's just  `MustRegister` - if an error occurs (which it shouldn't) then continuing doesn't make much sense anyway but we should fail.

main.go
```diff
@@ -113,8 +116,12 @@ func main() {
...
+		}
+		if err = registry.Register(buildInfo); err != nil {
```
oliver006: same

main.go
```diff
@@ -133,6 +140,9 @@ func main() {
 </body>
 </html>
 						`))
+		if err != nil {
```
oliver006: this doesn't really add anything useful, I think it's ok to ignore err here (and remove it from line 127).

mjtrangoni:
@oliver006 PTAL. I dropped the vet shadow commit.

oliver006:
@mjtrangoni - thanks for making the changes. There are still a few review comments that were not addressed, esp in main.go and in exporter/redis.go around lines 717 and 722
Once those addressed we should be good to 🚢 this.

oliver006:
Closing - please re-open if this is still being worked on.

PR #188, opened 2018-09-18
add all metrics from 'redis info' persistence section

zakharovvi:
Hello @oliver006 !

Thank you for this project. We've been using redis_exporter for over a year and it helps a lot.

I want to use these metrics in alerts rules:
* rdb_last_save_time
* rdb_last_bgsave_status
* aof_last_bgrewrite_status
* aof_last_write_status

Also I added other missed "persistence" metrics for consistency:
* rdb_bgsave_in_progress
* rdb_last_cow_size
* aof_last_cow_size

PR #180, opened 2018-07-30
Add total_system_memory

EriconYu:
Add total_system_memory

PR #210, opened 2018-12-15
bring back some accidentally dropped commits

oliver006:
@zakharovvi - no idea what happened to that commit, maybe a bad rebase on my end? 

Either way, this PR should bring back the metrics you added in #188 and also clean up the names of three of the `aof_` metrics by adding the proper suffix.


Supermathie:
I notice that there's some other commits (all of them? didn't really check) from the v0.21 line missing from master:

https://github.com/oliver006/redis_exporter/compare/v0.21.2?expand=1

oliver006:
Just checked, I see the commits for #180 but there are definitely more commits missing, let me bring them back as well.

PR #217, opened 2019-01-05
Add Alpine Docker image

mdawar:
I added a Docker build stage for an Alpine based image to give more choices to users like me who would like to use a shell inside the container for variable interpolation or any other purposes, I hope you don't mind this change, we can build the image as usual and choose the targets, for example:

A normal build will use the last build stage which is based on the scratch image:

    $ docker build -t oliver006/redis_exporter .

Or we can specify the target using `--target scratch`.

To build the Alpine based image:

    $ docker build --target alpine -t oliver006/redis_exporter:alpine .

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -12,18 +12,31 @@ echo "docker login done"
 export BUILD_ARGS="--rm=false --build-arg TAG=${CIRCLE_TAG} --build-arg SHA1=${CIRCLE_SHA1} --build-arg DATE=$(date +%F-%T)"
 echo  "BUILD_ARGS: $BUILD_ARGS"
 
-docker build -t "21zoo/redis_exporter:$CIRCLE_TAG" $BUILD_ARGS .
-docker push     "21zoo/redis_exporter:$CIRCLE_TAG"
+# Scratch image
+docker build --target scratch \
+             -t "21zoo/redis_exporter:$CIRCLE_TAG" \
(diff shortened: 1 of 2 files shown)


oliver006:
Thanks for the PR!
I think this make sense, I'll get to reviewing your changes in the next couple of days.

build.sh
```diff
@@ -12,18 +12,31 @@ echo "docker login done"
...
-docker push     "21zoo/redis_exporter:latest"
-
-docker build -t "oliver006/redis_exporter:$CIRCLE_TAG" $BUILD_ARGS  .
-docker push     "oliver006/redis_exporter:$CIRCLE_TAG"
-
-docker build -t "oliver006/redis_exporter:latest" $BUILD_ARGS  .
-docker push     "oliver006/redis_exporter:latest"
-
+# Scratch image
+docker build --target scratch \
```
oliver006: `--target scratch` - this is neat, didn't know you can do that.

mdawar: Yes we can name build stages and reference them this way, also we can use `--from` to copy files from external images and not only build stages. Thank you for approving the request, BTW very nice work.

oliver006: Interesting, thanks for explaining. I clearly did not keep up with what you can do with multistage builds...

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v0.26.0

PR #220, opened 2019-01-08
PR for Issues#219

redzippo1:
1 And attribute slave_port to metric "redis_connected_slave_offset".
2 Fix metric "redis_instance_info" by split function extractInfoMetrics to extractInfoMetrics and extractClusterInfoMetrics.


redzippo1:
update test func TestParseConnectedSlaveString

redzippo1:
You are absolute right.
I 'll try it in a better way.

redzippo1:
I have updated "redis.go" for implement in a better way.

oliver006:
Sorry, I only got back to this now, this week has been super busy.

To be honest, I don't think I quite follow what you did in the last commit with `handlerFunc()` and I think it's a bit hard to understand. Your changes up until the last commit seemed ok, with my only suggestion to refactor that repeated code into one function. Maybe you can go back and see if you can continue from commit https://github.com/oliver006/redis_exporter/pull/220/commits/dee20b984ab7df6b3cffbbd8287a4184a7b47afd and take it from there? What do you think?

redzippo1:
Thranks for reply.
Let me introduce the reason way I did it in that way.

1 ClusterInfo and redisInfo seems different in some detail, the old function "extractInfoMetrics" use "padDBKeyCounts bool" parameter to discriminate them is not the best way. And I think as least we have to split it in two function(because of the differents become more and more).
2 In the commit [dee20b9](https://github.com/oliver006/redis_exporter/commit/dee20b984ab7df6b3cffbbd8287a4184a7b47afd), the way to handle clusterInfo and redisInfo use a big function called "extractInfoMetrics", maybe it can be more beautiful.
3 I think “closure” is one of features of Golang(compare to JAVA, whitch is the major language I use), and what I did is implement it in a "closure" way.

If you think what I say is not right, I 'd like to do it in your way.
Thank you!

oliver006:
Ah, I see what you mean, thanks for explaining. I agree, `extractInfoMetrics()` has gotten pretty long but I find the closure approach hard to read and the code isn't currently using any so I'd prefer to keep it that way.

How about this: you start with https://github.com/oliver006/redis_exporter/commit/dee20b984ab7df6b3cffbbd8287a4184a7b47afd and take the code from line to 595 to 620 here https://github.com/oliver006/redis_exporter/blob/dee20b984ab7df6b3cffbbd8287a4184a7b47afd/exporter/redis.go#L595-L620 and move it into a function called `registerMetric()` or so and use it in line 595 and also line 670 and you should be done?

redzippo1:
Thanks for reply.
I will do it in that way.
: )

oliver006:
Thanks!

redzippo1:
I have update code in the way we discussed.

Issue #221, opened 2019-01-08
Wrong key name logged if key not found


vmoebius:
If redis key names provided by `--check-single-keys` doesn't exist at time of scraping then a debug message:

    "Key key not found not found when trying to get type and size."

is logged, i.e. the actual error string is interpolated instead of the name of the missing key.

oliver006:
Closed by #222 - thanks @vmoebius

PR #222, opened 2019-01-08
fixed issue#221: apply correct key name while formatting debug log

vmoebius:
My suggestion to fix issue #221.

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -859,7 +859,7 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		if err != nil {
 			switch err {
 			case errNotFound:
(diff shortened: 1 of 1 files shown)


exporter/redis.go
```diff
@@ -859,7 +859,7 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		if err != nil {
 			switch err {
 			case errNotFound:
-				log.Debugf("Key %v not found when trying to get type and size.", err)
+				log.Debugf("Key '%v' not found when trying to get type and size.", k.key)
```
oliver006: I think you can make this `%s` as k.key is a string.

vmoebius: You are right.

oliver006:
Thx!

PR #231, opened 2019-01-23
Enable support for having the redis password in a file

javsalgar:
Signed-off-by: Javier J. Salmeron Garcia <jsalmeron@bitnami.com>

In some contexts, providing the password using REDIS_PASSWORD is not recommended, as this information could be potentially extracted by privileged containers that could execute commands like `docker inspect`. The same could be said to anyone that could access the kubelet host where the container is running. Because of this, some deployments require the ability to pass passwords as files. This PR enables this feature.

I set that REDIS_PASSWORD and REDIS_PASSWORD_FILE (the new variable that points to a file containing the password) to be mutually exclusive, so only one can be declared at the same time. This change does not break backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,22 +112,23 @@ cf push -f contrib/manifest.yml
 
 ### Flags
 
-Name               | Description
--------------------|------------
-debug              | Verbose debug output
-log-format         | Log format, valid options are `txt` (default) and `json`.
-check-keys         | Comma separated list of key patterns to export value and length/size, eg: `db3=user_count` will export key `user_count` from db `3`. db defaults to `0` if omitted. The key patterns specified with this flag will be found using [SCAN](https://redis.io/commands/scan).  Use this option if you need glob pattern matching; `check-single-keys` is faster for non-pattern keys.
-check-single-keys  | Comma separated list of keys to export value and length/size, eg: `db3=user_count` will export key `user_count` from db `3`. db defaults to `0` if omitted.  The keys specified with this flag will be looked up directly without any glob pattern matching.  Use this option if you don't need glob pattern matching;  it is faster than `check-keys`.
-script             | Path to Redis Lua script for gathering extra metrics.
-redis.addr         | Address of one or more redis nodes, comma separated, defaults to `redis://localhost:6379`.
-redis.password     | Password to use when authenticating to Redis
-redis.alias        | Alias for redis node addr, comma separated.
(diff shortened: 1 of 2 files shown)


javsalgar:
Note that I'm not very experienced with Go, so any suggestion would be very appreciated :)

oliver006:
Thanks for the PR and welcome to writing Go! I'll review this as soon as possible.

Question though: if someone gains access to a Kubernetes host or is able to run docker inspect - wouldn't they also be able to access a running container (e.g. via exec and run `sh` or `bash`) and inspect the file system of a redis_exporter container and be able to extract the contents of the password file?

javsalgar:
Potentially, this would be possible as well but, from the feedback I got, the attacker would normally do a docker inspect of all of the running containers and extract the env vars. Using this method forces you to know which container you want to attack, enter it, and find the file that contains the credentials. So, this method is not a perfect solution but a way to make things a bit more complicated (security through a bit more obscurity). This method is what other containers like Postgres, and for some companies, having this feature is essential for adopting a container.

javsalgar:
Hi, 

Did you have time to take a look? Is there anything I can help with?  we have a PR in the official Redis Helm chart that depends on this feature

oliver006:
Sorry, my bad, I'll try to get to this before end of day.

javsalgar:
Thank you so much! :)

javsalgar:
README updated

README.md
```diff
@@ -112,22 +112,23 @@ cf push -f contrib/manifest.yml
...
+check-keys          | Comma separated list of key patterns to export value and length/size, eg: `db3=user_count` will export key `user_count` from db `3`. db defaults to `0` if omitted. The key patterns specified with this flag will be found using [SCAN](https://redis.io/commands/scan).  Use this option if you need glob pattern matching; `check-single-keys` is faster for non-pattern keys.
+check-single-keys   | Comma separated list of keys to export value and length/size, eg: `db3=user_count` will export key `user_count` from db `3`. db defaults to `0` if omitted.  The keys specified with this flag will be looked up directly without any glob pattern matching.  Use this option if you don't need glob pattern matching;  it is faster than `check-keys`.
+script              | Path to Redis Lua script for gathering extra metrics.
+redis.addr          | Address of one or more redis nodes, comma separated, defaults to `redis://localhost:6379`.
+redis.password      | Password to use when authenticating to Redis
+redis.password-file | Path to a file containing the password to use when authenticating to Redis (note: this is mutually exclusive with `redis.password`)
```
oliver006: 👍

oliver006:
Releases as https://github.com/oliver006/redis_exporter/releases/tag/v0.28.0

PR #232, opened 2019-01-29
Run as non-root user for secure environments

dotLou:
I'm running this in a kubernetes environment where there's a pod security policy disallowing the use of non-root images.

The industry recommendation is typically to run as a numbered user to avoid conflicts with actual host users.

This PR sets the user and group to `59000` to be able to run in those conditions.


oliver006:
You're welcome. I'll cut a release as soon as I have #231 reviewed and merged, should be soon.

oliver006:
Releases as https://github.com/oliver006/redis_exporter/releases/tag/v0.28.0

PR #234, opened 2019-02-17
Tile38 support

1995parham:
[Tile38](https://github.com/tidwall/tile38) is a geospatial database that is using redis communication protocol and structure. I add a flag to support metric collection from tile38 databases. Tile38 offer a command named `Server` that is not available in `redis_exporter` and I add it with this PR.

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -896,10 +896,12 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 
 	e.extractInfoMetrics(infoAll, addr, e.redis.Aliases[idx], scrapes, dbCount)
 
+	// SERVER command only works on tile38 database. check the following link to
+	// findout more: https://tile38.com/
 	if serverInfo, err := redis.Strings(doRedisCmd(c, "SERVER")); err == nil {
 		e.extractTile38Metrics(serverInfo, addr, e.redis.Aliases[idx], scrapes)
 	} else {
-		log.Errorf("Redis SERVER err: %s", err)
+		log.Debugf("Tile38 SERVER err: %s", err)
 	}
 
 	if reply, err := doRedisCmd(c, "LATENCY", "LATEST"); err == nil {
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -277,7 +277,7 @@ func parseKeyArg(keysArgString string) (keys []dbKeyPair, err error) {
 
 // NewRedisExporter returns a new exporter of Redis metrics.
 // note to self: next time we add an argument, instead add a RedisExporter struct
-func NewRedisExporter(host RedisHost, namespace, checkSingleKeys, checkKeys string, isTile38 bool) (*Exporter, error) {
+func NewRedisExporter(host RedisHost, namespace, checkSingleKeys, checkKeys string) (*Exporter, error) {
 
 	e := Exporter{
 		redis:     host,
@@ -314,8 +314,6 @@ func NewRedisExporter(host RedisHost, namespace, checkSingleKeys, checkKeys stri
 		}),
 	}
 
-	e.isTile38 = isTile38
-
 	var err error
 
 	if e.keys, err = parseKeyArg(checkKeys); err != nil {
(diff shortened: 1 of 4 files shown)


oliver006:
Thanks for the PR!
I'll try to have a look in the next couple of days and will get back to you.

1995parham:
@oliver006 Thanks for your time

exporter/redis.go
```diff
@@ -139,6 +141,26 @@ var (
 		// # Cluster
 		"cluster_stats_messages_sent":     "cluster_messages_sent_total",
 		"cluster_stats_messages_received": "cluster_messages_received_total",
+
+		// # Tile38
+		// based on https://tile38.com/commands/server/
+		"aof_size":        "aof_size_bytes",
+		"avg_item_size":   "avg_item_size_bytes",
+		"cpus":            "cpus_total",
+		"heap_released":   "heap_released_byte",
```
oliver006: `heap_released_bytes` ?

exporter/redis.go
```diff
@@ -139,6 +141,26 @@ var (
...
+		"heap_size":       "heap_size_bytes",
+		"http_transport":  "http_transport",
```
oliver006: Not familiar with Tile38 - what does this metric store/expose?

1995parham: You can see [here](https://tile38.com/commands/server/) to find out more about tile38 metrics.

1995parham:
I agree, so will remove the tile38 flag then always execute server command and parse it whether a given connection is til38 or not.
Do you think it will be good to add `tile38` prefix for its metrics?

exporter/redis.go
oliver006: :bump:

exporter/redis.go
oliver006: I don't see `http_transport` on that page - what is that one actually exporting?

exporter/redis.go
```diff
@@ -859,6 +896,12 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 
 	e.extractInfoMetrics(infoAll, addr, e.redis.Aliases[idx], scrapes, dbCount)
 
+	if serverInfo, err := redis.Strings(doRedisCmd(c, "SERVER")); err == nil {
+		e.extractTile38Metrics(serverInfo, addr, e.redis.Aliases[idx], scrapes)
+	} else {
+		log.Errorf("Redis SERVER err: %s", err)
```
oliver006: Maybe leave a comment before line 899 that this will succeed for Tile38 servers only and then make this a `Debugf()`, otherwise it'll spam the logs for regular redis instances.

1995parham: @oliver006 Really thanks for your time.
`http_transport` exposes a boolean that indicates HTTP transport is using or not.
About debugging it is my mistake and I will switch it to debug level log.
And what do you think about tests? Are you agree with adding tile38 into circle-ci and tests?

exporter/redis.go
```diff
@@ -896,10 +896,12 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 
 	e.extractInfoMetrics(infoAll, addr, e.redis.Aliases[idx], scrapes, dbCount)
 
+	// SERVER command only works on tile38 database. check the following link to
```

exporter/redis.go
```diff
@@ -859,6 +896,14 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
...
+	// findout more: https://tile38.com/
```
oliver006: findout --> "find out"

exporter/redis.go
oliver006: Ah ok, thanks for explaining. Your changes look good but yeah, adding a test for the newly introduced parsing function is probably a good idea.

exporter/redis.go
1995parham: I fixed it 😄

exporter/redis.go
1995parham: Ok, I will add tests for tile38.

PR #235, opened 2019-02-19
Adding replication lag metric

samber:
Adding a new metric for measuring replication lag.

```
# HELP redis_connected_slave_lag_seconds Lag of connected slave
# TYPE redis_connected_slave_lag_seconds gauge
redis_connected_slave_lag_seconds{addr="**********",alias="",slave_ip="********",slave_port="6379",slave_state="online"} 1
```

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -444,8 +444,8 @@ func parseConnectedSlaveString(slaveName string, slaveInfo string) (offset float
 
 	lag, err = strconv.ParseFloat(connectedSlaveInfo["lag"], 64)
 	if err != nil {
-		log.Debugf("Can not parse connected slave lag, got: %s", connectedSlaveInfo["lag"])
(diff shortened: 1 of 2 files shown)


oliver006:
Change looks good but quick question, is lag always expected to be there, even for older versions of redis? Or do you need to make this change in a backwards compatible way where you domn't fail (aka return) in line 448 if the lag info is not present?

samber:
You're right @oliver006 !! 🤭

It's been added to redis in 3.0, in 2013.

I found the commit => https://github.com/antirez/redis/commit/37c29e037b08159e901227f0184973442bb86c2d

I think 2.8 is still used a lot. I'll make a new commit ;)

samber:
@oliver006 done

oliver006:
Nice work, thanks for the PR!

PR #236, opened 2019-02-21
Use the source repo references in the redis_exporter documentation

tariq1890:


oliver006:
Good catch!

But the second link is missing `/master/` - it should be `https://raw.githubusercontent.com/oliver006/redis_exporter/master/contrib/openshift-template.yaml`

oliver006:
Looks good, thanks for the PR!

PR #237, opened 2019-03-01
Streams metrics support.

kak-tus:
There are new key type in Redis 5 - Streams.
So, we get ability to monitor them.


oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v0.30.0

PR #239, opened 2019-03-08
log the error from scanForKeys

rikonor:
I'm not sure if you were not logging the error intentionally, but it'd be nice if it could be logged.

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -818,7 +818,7 @@ func scanForKeys(c redis.Conn, pattern string) ([]string, error) {
 	for {
 		arr, err := redis.Values(c.Do("SCAN", iter, "MATCH", pattern))
 		if err != nil {
-			return keys, fmt.Errorf("error retrieving '%s' keys", pattern)
+			return keys, fmt.Errorf("error retrieving '%s' keys err: %s", pattern, err)
 		}
 		if len(arr) != 2 {
 			return keys, fmt.Errorf("invalid response from SCAN for pattern: %s", pattern)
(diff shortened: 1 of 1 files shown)


oliver006:
Makes sense, thanks for the PR!

exporter/redis.go
```diff
@@ -844,7 +844,7 @@ func getKeysFromPatterns(c redis.Conn, keys []dbKeyPair) (expandedKeys []dbKeyPa
 			}
 			keyNames, err := scanForKeys(c, k.key)
 			if err != nil {
-				log.Errorf("error with SCAN for pattern: %#v", k.key)
+				log.Errorf("error with SCAN for pattern: %#v: %s", k.key, err)
```
oliver006: Can you please change this to `"error with SCAN for pattern: %#v err: %s"`

rikonor:
I just realized the same thing happens here: https://github.com/oliver006/redis_exporter/pull/239/files#diff-a75393c3715eaef61ada7312397db7f2R819

Are you OK with me bubbling that error up too?

oliver006:
Yeah, do that while you're at it.

oliver006:
And heads up, I just merged #238 so you need to rebase as i think I changed related code.

rikonor:
OK, rebased, reformatted the log and added the other one as well.

rikonor:
Thanks for the quick merge. We ended up finding out that we had a KEYS call which was killing Redis. Switched that out and all's well now. 😄

oliver006:
Oh interesting. Yeah, KEYS calls returning large sets of keys can be dangerous. Glad to hear you figured it out.

PR #240, opened 2019-03-19
add last_slow_execution_duration_seconds

sKabYY:
add last_slow_execution_duration_seconds

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -247,6 +247,11 @@ func (e *Exporter) initGauges() {
 		Name:      "slowlog_last_id",
 		Help:      "Last id of slowlog",
 	}, []string{"addr", "alias"})
+	e.metrics["last_slow_execution_duration_seconds"] = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: e.namespace,
+		Name:      "last_slow_execution_duration_seconds",
+		Help:      "The amount of time needed for last slow execution, in seconds",
+	}, []string{"addr", "alias"})
 }
 
 // splitKeyArgs splits a command-line supplied argument into a slice of dbKeyPairs.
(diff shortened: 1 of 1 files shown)


exporter/redis.go
```diff
@@ -1011,15 +1016,18 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 
 	if values, err := redis.Values(c.Do("SLOWLOG", "GET", "1")); err == nil {
 		var slowlogLastId int64 = 0
+		var lastSlowExecutionDurationSeconds float64 = 0
 
 		if len(values) > 0 {
 			if values, err = redis.Values(values[0], err); err == nil && len(values) > 0 {
 				slowlogLastId = values[0].(int64)
+				lastSlowExecutionDurationSeconds = float64(values[2].(int64)) / 1e6
```
oliver006: Question: line 1022 only checks for len(values) > 0 so this doesn't guarantee that we actually have three elements (you're accessing `[2]`) - has this value always been present or was it added recently and you need to make this change backward compatible?

sKabYY: According to the official document of Redis, this value is available since 2.2.12. The `len(values)` in line 1022 is not less than 4 if the outer `len(values) > 0` (line 1021).

I will add the `>= 2` guard and make a new pull request.

exporter/redis.go
```diff
@@ -1021,7 +1021,9 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
 		if len(values) > 0 {
 			if values, err = redis.Values(values[0], err); err == nil && len(values) > 0 {
 				slowlogLastId = values[0].(int64)
-				lastSlowExecutionDurationSeconds = float64(values[2].(int64)) / 1e6
+				if (len(values) >= 2) {
```
oliver006: This would need to be `> 2` if you're accessing the third (`[2]`) element.

exporter/redis.go
oliver006: 👍 
make it `> 2` as you want at least three elements. And then this should be ready for merge.

exporter/redis.go
sKabYY: Fixed!

oliver006:
Nice!

PR #243, opened 2019-03-21
Split key-value line on first colon

### LINKED ISSUE #242: IPv6 address in redis output breaks parsing
When there's an IPv6 address in the output of Redis, the line is skipped, resulting in missing metrics.

For example:
```
slave0:ip=2a00:1450:400e:808::200e,port=6379,state=online,offset=9694740951,lag=1
```
This line will never be handled.

Line of code causing this:
https://github.com/oliver006/redis_exporter/blob/e3f43cfe305f8d20d931de6eb84e32c664b7edd1/exporter/redis.go#L627-L630

jobec:
Fixes #242

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -624,11 +624,7 @@ func (e *Exporter) extractInfoMetrics(info, addr string, alias string, scrapes c
 			continue
 		}
 
-		split := strings.Split(line, ":")
-		if len(split) != 2 {
-			continue
-		}
-
+		split := strings.SplitN(line, ":", 2)
 		fieldKey := split[0]
 		fieldValue := split[1]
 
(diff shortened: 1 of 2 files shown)


exporter/redis.go
```diff
@@ -624,11 +624,7 @@ func (e *Exporter) extractInfoMetrics(info, addr string, alias string, scrapes c
 			continue
 		}
 
-		split := strings.Split(line, ":")
-		if len(split) != 2 {
```
oliver006: I'd keep the `len == 2` check. What if there's no `:` in the string?

oliver006:
Thanks for the PR!

exporter/redis.go
jobec: Isn't that the check before it?

https://github.com/oliver006/redis_exporter/blob/e3f43cfe305f8d20d931de6eb84e32c664b7edd1/exporter/redis.go#L623-L625

oliver006: You're very right, didn't see that in the diff. Good to go!

jobec:
A new release would be great ;-)

oliver006:
Coming right up!

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v0.31.0

jobec:
Could it be you forgot the binaries?

PR #249, opened 2019-04-08
Update vendoring

SuperQ:
* Update to Go modules.
* Update vendoring to current releases.


oliver006:
Thanks for the PR, I got a couple of questions.

First, I'm not that familiar with go modules but from whatever I've read so far i thought the idea is to not use the vendor directory any longer but more or less "pin" everything via the go.* file(s) and download the dependencies at build time (not unlike things used to work with "go get" about 3-4 years ago) - is that not the case here? I see you're using go modules but there's still the vendor directory with all the dependencies.

Also, it looks like a lot of stuff that `dep` would trim was added as part of the vendor update (tests, test fixtures, etc) - is that intentional?

And one of the tests failed, not sure if that's a flaky test or a result of the vendor update:
` redis_test.go:1101: didn't find test_exporter_scrapes_total `

SuperQ:
You can still use the vendor directory with go modules, we still do this in the main Prometheus projects. We use `go build -mod=vendor` as part of our Makefiles to force the use of the vendor directory.

As for the files in the vendor directory, `go mod vendor` does similar trimming. It only brings in what's necessary afaik.

SuperQ:
If you want to drop the vendoring, that's also fine. You'll want to add `go mod download` to any build pipelines.

SuperQ:
Looking at the code, I think the test is flakey due to the exporter `Collect()` function calling `e.scrape()` in a goroutine. This is causing a race as no scrapes have been completed.

oliver006:
> You can still use the vendor directory with go modules, we still do this in the main Prometheus projects. We use `go build -mod=vendor` as part of our Makefiles to force the use of the vendor directory.

Interesting, so you really just use go modules instead of `dep` to manage the vendor directory.
`add go mod download to any build pipelines.` - eh, let's keep the vendor directory for now, that's fine, no need to download all the dependencies for every build.

As for the test, this one I've never seen fail before so I'm not sure it's related to the goroutine you mentioned but let me have a look.

oliver006:
Something must have changed with the implementation of the prometheus golang library, if you change https://github.com/oliver006/redis_exporter/blob/e0906c8759e9a6a8cacd281ea198fb8fe2018ea1/exporter/redis_test.go#L1071 to `prometheus.Counter` then it'll pick up the values just fine.

SuperQ:
Ahh, good catch, yes, I was testing things yesterday and had a suspicion that it was related to the prometheus client library update.

SuperQ:
Ahh, coveralls math. :frowning_face:

oliver006:
lol

oliver006:
This should be good to go - will merge and cut a new release.

SuperQ:
Thanks, this should bring some reasonable performance improvements due to the Prometheus library upgrade.

PR #250, opened 2019-04-08
Improve build script

SuperQ:
* Use single gox command to parallelize build.
* Automatically package any archs gox spits out.
* Add sha256sums for packages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ coverage.out
 dist/
 pkg/
 src/
+.build/
 .DS_Store
 .idea
 *.rdb
--- a/build-github-binaries.sh
+++ b/build-github-binaries.sh
@@ -1,4 +1,11 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+
+set -u -e -o pipefail
+
+if [[ -z "${DRONE_TAG}" ]] ; then
+  echo 'ERROR: Missing DRONE_TAG env'
+  exit 1
+fi
 
 echo "Building binaries for Github"
 echo ""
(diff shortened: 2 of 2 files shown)


oliver006:
Nice, this is a lot better than what we got right now.

oliver006:
https://cloud.drone.io/oliver006/redis_exporter/198/1/8

```

GO_LDFLAGS:  -s -w -extldflags "-static" -X main.VERSION=v0.33.0 -X  main.COMMIT_SHA1=09b985767518c57735b7f11b66c15808c4561db4 -X  main.BUILD_DATE=2019-04-09-17:37:28
--
```

:-/

Do the `go build` commands in the build script need the  `-mod=vendor` flag as well?

oliver006:
I guess I could add `-mod=vendor` to the GO_LDFLAGS variable?

SuperQ:
Hrm, I don't know how `gox` passes in those flags. You might need to add `GO111MODULE: on` in the env section of the drone config.

oliver006:
Good suggestion, let me try that.

SuperQ:
It may also need `-mod=vendor` on the GO_LDFLAGS in order to enforce the use of the vendor dir.

oliver006:
Setting the env var seems to have fixed the build, now on to the next issue: 

`rm: cannot remove 'dist/redis_exporter-v0.33.0.windows-386.zip': No such file or directory`

I guess we got to ignore errors here cause for the first zip there's nothing to delete.

SuperQ:
Ahh, yea, `[[ -f $f ]] && rm $f`. Forgot that rm fails on removal of nothing.

PR #253, opened 2019-04-23
For redis 2.4 the INFO command should be passed without arguments.

i02sopop:
When running the redis exporter with redis 2.4 there's an error running the INFO command, since it doesn't accept the ALL parameter. The proposed fix is to run INFO without arguments if INFO ALL fails.


oliver006:

Issue #251, opened 2019-04-11
Compose/ibmcloud  redis exporter with ssl `rediss://` does not work 


wennnds:
when setting up exporter for ibmcloud, compose redis doesn't produce any metrics from prometheus

wennnds:
does the part to add SSL cert as part of argument exist? @oliver006

oliver006:
Sorry, not familiar with how IBM Cloud works or sets up Prometheus and./or exporters.

wennnds:
are there plans on SSL cert as an arg to connect to rediss database?

oliver006:
No plans right now, sorry.

desourav:
@oliver006 Here are the relevant logs ::
````
time="2019-04-30T16:20:54Z" level=debug msg="aborting for addr: rediss://xxxx:15242 - redis err: dial rediss: unknown network rediss"
time="2019-04-30T16:21:09Z" level=debug msg="Trying DialURL(): rediss://xxxx:15242"
time="2019-04-30T16:21:09Z" level=debug msg="DialURL() failed, err: x509: certificate signed by unknown authority"
````
We want to ignore the SSL verification. Is there a flag we can pass to do the same?

oliver006:
There is no flag available right now but definitely something that can be added in the future. Is the cert self-signed?

desourav:
@oliver006 yes its a self signed cert.

oliver006:
Ok.
Implementing a flag to disable cert validation shouldn't be that hard, I think it might just be a matter of setting a flag when calling `redis.DialURL()`.
If I have a chance to add it in the future then I'll let you now but in the meantime PRs are welcome.

oliver006:
Version 1.0.0 now has a flag to skip TLS verification, closing this.

aa7955:
The flag to skip TLS certs is not what is needed since we're making `rediss` connections. In fact, an environment variable that allows you to add the self-signed certificate is what's needed. Skipping TLS verification will not work. Maybe I will submit a PR for this one.

oliver006:
Did you look at the `--tls-client-key-file` and `--tls-client-cert-file` params and tried using them to make the exporter use your cert?

aa7955:
Yeah, but there needs to be a CA cert flag because the self-signed certificate doesn't require a key, just a CA cert.

tangerstein:
[...]

Issue #173, opened 2018-06-27
Put the release file in a folder inside the .tar.gz


grantstephens:
So this might be a weird one, but essentially all the other exporters that I've come across and use put the binary inside another folder in the release .tar.gz- this might be annoying or whatever, but at least it was consistent, except for this exporter.
Would it be possible to adjust the release process to make this possible?

oliver006:
Hey there,

my instinct is to not change it at this point because probably people rely on it to not be in a folder and their puppet, ansible, whatever scripts would break if out of a sudden future releases are in a folder.

Is this currently causing serious issues for you or is it more of a convenience thing?

grantstephens:
Thanks for the reply and sorry for my delayed reply.
So we currently have a generic exporter deploy script which covers all the exporters we have encountered so far (around 8 off the top of my head) where the binary is another folder deep. 
So I guess it is a convenience thing, but it is also a conformity thing. 
Don't know if it is worth leaving the issue open to see if anybody else comes across it and can give a thumbs up if they agree?

dkanchev:
@oliver006 @grantstephens Can't we use a symlink and this way work around this problem?

zoonage:
+1 for this, I would like to add it to https://github.com/lest/prometheus-rpm without having to add customisations to the build process for this exporter.

grantstephens:
@dkanchev it looks like that would be possible. Something like `ln -s ../redis_exporter redis_exporter-vxxxx.linux-arch/redis_exporter` should do it.

oliver006:
Heads up, PR #256 will fix this and put the binary in a folder. I'm expecting to merge it some time next week.

oliver006:
As of v1.0.0 the binary is now in a folder named e.g. `redis_exporter-v1.0.0.linux-amd64`.

Issue #146, opened 2018-03-21
Allow command renaming


dgrothaus-ku:
It would be nice if the exporter supported renamed dangerous commands.

Some cloud Redis installations (AWS) or password protected instances use renamed CONFIG, FLUSHDB and FLUSHALL commands.

A flag to provide the name of renamed commands would make otherwise unavailable config values available.

oliver006:
Agreed, that'd be useful to have (and this has come up before as a request).
I personally don't have a need for renaming commands but would be happy to review and merge a PR.

oliver006:
You can now rename the `CONFIG` command using the `--config-command` flag, closing this.

Issue #228, opened 2019-01-18
Bring back total_system_memory?


juris:
I understand that you've dropped `total_system_memory` because [node_exporter has the same functionality](https://github.com/oliver006/redis_exporter/pull/203).

However, not every Redis instance comes with node_exporter. In my case it is AWS ElastiCache Redis. Yes, I know I can use CloudWatch exporter, but we have multiple ElastiCache Redises running and CloudWatch metrics cost money :(

Would you consider bringing `total_system_memory` metric back?

oliver006:
I think we should bring back `total_system_memory` (I guess as `total_system_memory_bytes`). It was removed here https://github.com/oliver006/redis_exporter/pull/203/files#diff-a75393c3715eaef61ada7312397db7f2L78 and in retrospect I think I would have preferred to keep it because there are more users like you that use e.g. AWS ElastiCache Redis.
Open a PR if you like or I might be able to get to it in the next couple of days.

oliver006:
Closed by #229

SuperQ:
I'm extremely annoyed by this. I understand the desire to have this metric, but however inconvenient it is, this violates best practices for Prometheus exporters.

At a minimum, this should be behind a feature flag and disabled by default.

SuperQ:
Might I also suggest using the cloudwatch_exporter, the [costs are not that large](https://github.com/prometheus/cloudwatch_exporter#cost).

oliver006:
Thanks for your feedback, @SuperQ. I understand that exporting `total_system_memory` is not in line with the best practices.
But the metric is readily available from Redis and not everyone can or wants to run another exporter just for that one metric. It's not just about $$$ cost but also cognitive and maintenance overhead and for me that outweighs the concerns regarding best practices.

SuperQ:
The problem is, where do you end that convenience? Should the redis exporter also include system CPU information?

For the majority of users that are already running the node_exporter, this exporter now exposes duplicate information.

This is why the Prometheus project is designed the way it is. Separate concerns for separate systems.

Please put this metric behind a default-disabled feature flag so that the majority of users don't get duplicate data.

oliver006:
As I said, I *understand* why you're raising this issue, I just don't think it's that big a problem.

Let me think about the feature flag suggestion.

SuperQ:
I will see about adding the feature flag.

oliver006:
[...]

PR #256, opened 2019-05-01
redis_exporter v1.0.0

### LINKED ISSUE #126: Parallelize requests
It looks like the exporter queries each redis node sequentially, would it be possible to parallelize these requests?

### LINKED ISSUE #178: Export counters as counters
I work on the [Stackdriver integration with Prometheus](https://github.com/Stackdriver/stackdriver-prometheus) and a user filed [an issue](https://github.com/Stackdriver/stackdriver-prometheus/issues/14#issue-343468080) because this redis exporter exports counters as gauges, and Stackdriver doesn't allow rate computations over gauge. I know this isn't a major problem for Prometheus (it can mostly cause glitches on process restart), but would you consider fixing the [setMetrics method](https://github.com/oliver006/redis_exporter/blob/master/exporter/redis.go#L782) to set the right metric type? Any workaround on our side will be hard to maintain. Thanks!

### LINKED ISSUE #204: promtool check metrics issues
Hi @oliver006 ,

Checking the redis_exporter `v0.23.0 output with promtool, I have the following result,

```
# curl -s http://fqdn:9121/metrics | promtool check metrics
redis_commands_duration_seconds_total non-counter metrics should not have "_total" suffix
redis_commands_processed_total non-counter metrics should not have "_total" suffix
redis_commands_total non-counter metrics should not have "_total" suffix
redis_connections_received_total non-counter metrics should not have "_total" suffix
redis_evicted_keys_total non-counter metrics should not have "_total" suffix
redis_expired_keys_total non-counter metrics should not have "_total" suffix
redis_keyspace_hits_total non-counter metrics should not have "_total" suffix
redis_keyspace_misses_total non-counter metrics should not have "_total" suffix
redis_net_input_bytes_total non-counter metrics should not have "_total" suffix
redis_net_output_bytes_total non-counter metrics should not have "_total" suffix
redis_rejected_connections_total non-counter metrics should not have "_total" suffix
```

If you agree, I could send a PR for fixing this, and also adding this check to CircleCI.

oliver006:
### Why
This project was started over four years ago and a lot has changed since. This is an effort to get the redis_exporter to more closely adhere to the [Best Practices for Writing Exporters](https://prometheus.io/docs/instrumenting/writing_exporters/) and clean up the code.

The biggest change is that the exporter no longer directly supports configuring (and scraping) multiple redis instances but pushes that responsibility to the Prometheus server. If this is how you currently use the exporter, then have a look [at this section](https://github.com/oliver006/redis_exporter/blob/oh_1.0_here_we_come/README.md#prometheus-configuration-to-scrape-several-hosts) how to configure the Prometheus server to scrape multiple instances.

However, if you already use the exporter as intended by the Prometheus project and scrape only one instance (and e.g. use the `--redis.addr` and `--redis.password` cmd line flags) then little to nothing will change for you. 

### Breaking changes in this PR:
- The exporter will no longer directly do any Service Discovery and/or scrape multiple redis instances at once.\
All this functionality should happen in the Prometheus server.<br> You can still scrape multiple redis instances with one exporter, [see the README for an example ](https://github.com/oliver006/redis_exporter/blob/oh_1.0_here_we_come/README.md#prometheus-configuration-to-scrape-several-hosts) Prometheus server configuration.
  - this means all the `addr` and `alias` labels are gone, use Prometheus server label rewriting for this if need/want them back.
  - introduced `/scrape` endpoint to allow scraping a particular instance
    - [x] fixes #126 Parallelize requests
    - [see the README for an example ](https://github.com/oliver006/redis_exporter/blob/oh_1.0_here_we_come/README.md#prometheus-configuration-to-scrape-several-hosts)
- The exporter now uses ConstMetrics to correctly export counters and gauges. as per the best practices
  - fixes #178 Export counters as counters
  - fixes #204 fix promtool linting errors
- `total_system_memory_bytes` will not be included by default any longer, use a command line flag to include it anyway.
- renamed `latency_spike_milliseconds` -> latency_spike_seconds
- renamed all the cpu metrics to end with `_seconds`
- `last_scrape_duration` and `scrape_error` metrics are now per-instance (if you scrape multiple instances) and `scrape_error` includes the error as a label value
- Tile38 metrics are only scraped if `--is-tile38` command line flag is set
- [x] #173 Put the release file in a folder inside the .tar.gz

<br>

### New features:
- [x] related to #251 - added a flag to allow skipping cert validation
- [x] #146 allow overriding CONFIG command name

<br>

### Todo
- [x] remove 1.0.0-rc drone step
- [x] fix up link to README in this PR to link to master for future reference


oliver006:
man-crossing-street

oliver006:
This is available on docker as `v1.0.0-rc.1`

`docker run -d --name redis_exporter -p 9121:9121 oliver006/redis_exporter:v1.0.0-rc.1`

oliver006:
@SuperQ - this PR addresses the issue you raised in #228 - do you mind having a look, at the feature in particular and maybe the PR in general? Could always some more feedback here, esp if the general direction makes sense.

@brian-brazil - this PR addresses #31 that you raised a million years ago. Mind having a look if the general idea of the changes is somewhat sane?  (cc @jkohen re: issue #178 )

README.md
```diff
@@ -19,23 +19,83 @@ Supports Redis 2.x, 3.x, 4.x, and 5.x
...
+```yaml
 
-...
+scrape_configs:
+  ## config for the multiple redis targets that the exporter will scrape
+  - job_name: 'redis_exporter_targets'
+    static_configs:
+      - targets:
+        - redis://first-redis-host:6379
+        - redis://h:password@second-redis-host:6379
```
brian-brazil: This is not a safe way to pass a password.

exporter/redis.go
```diff
@@ -108,21 +102,12 @@ var (
...
-		"rejected_connections":       "rejected_connections_total",
-		"expired_keys":               "expired_keys_total",
-		"evicted_keys":               "evicted_keys_total",
-		"keyspace_hits":              "keyspace_hits_total",
-		"keyspace_misses":            "keyspace_misses_total",
-		"pubsub_channels":            "pubsub_channels",
-		"pubsub_patterns":            "pubsub_patterns",
-		"latest_fork_usec":           "latest_fork_usec",
+		"instantaneous_ops_per_sec": "instantaneous_ops_per_sec",
+		"instantaneous_input_kbps":  "instantaneous_input_kbps",
```
brian-brazil: bytes/s is preferred

exporter/redis.go
```diff
@@ -108,21 +102,12 @@ var (
...
+		"instantaneous_output_kbps": "instantaneous_output_kbps",
+		"pubsub_channels":           "pubsub_channels",
+		"pubsub_patterns":           "pubsub_patterns",
+		"latest_fork_usec":          "latest_fork_usec",
```
brian-brazil: seconds

exporter/redis.go
```diff
@@ -162,96 +141,62 @@ var (
...
+		"rejected_connections":   "rejected_connections_total",
+		"total_net_input_bytes":  "net_input_bytes_total",
+		"total_net_output_bytes": "net_output_bytes_total",
+
+		"expired_keys":    "expired_keys_total",
+		"evicted_keys":    "evicted_keys_total",
+		"keyspace_hits":   "keyspace_hits_total",
+		"keyspace_misses": "keyspace_misses_total",
+
+		"used_cpu_sys":           "used_cpu_sys_seconds_total",
```
brian-brazil: Used is implied here

exporter/redis.go
```diff
@@ -282,104 +227,146 @@ func parseKeyArg(keysArgString string) (keys []dbKeyPair, err error) {
...
-			Namespace: namespace,
-			Name:      "exporter_last_scrape_error",
-			Help:      "The last scrape error status.",
-		}),
+
+		targetScrapeDuration: prometheus.NewSummary(
+			prometheus.SummaryOpts{
+				Namespace: opts.Namespace,
+				Name:      "target_scrape_collection_duration_seconds",
+				Help:      "Duration of collections by the SNMP exporter",
```
brian-brazil: SNMP?

exporter/redis.go
```diff
@@ -282,104 +227,146 @@ func parseKeyArg(keysArgString string) (keys []dbKeyPair, err error) {
...
 
-	e.initGauges()
+	if opts.IncludeMetricTotalSysMemory {
+		metricMapGauges["total_system_memory"] = "total_system_memory_bytes"
+	}
+
+	e.metricDescriptions = map[string]*prometheus.Desc{}
+	e.metricDescriptions["up"] = newMetricDescr(opts.Namespace, "up", "Information about the Redis instance", nil)
+	e.metricDescriptions["instance_info"] = newMetricDescr(opts.Namespace, "instance_info", "Information about the Redis instance", []string{"role", "redis_version", "redis_build_id", "redis_mode", "os"})
+	e.metricDescriptions["last_scrape_duration"] = newMetricDescr(opts.Namespace, "exporter_last_scrape_duration_seconds", "The last scrape duration", nil)
```
brian-brazil: You'd generally provide information about this scrape.

exporter/redis.go
```diff
@@ -502,13 +489,26 @@ func extractConfigMetrics(config []string, addr string, alias string, scrapes ch
...
 		if val, err := strconv.ParseFloat(strVal, 64); err == nil {
-			scrapes <- scrapeResult{Name: fmt.Sprintf("config_%s", config[pos*2]), Addr: addr, Alias: alias, Value: val}
+			e.registerGaugeValue(ch, fmt.Sprintf("config_%s", config[pos*2]), val)
 		}
 	}
 	return
 }
 
-func (e *Exporter) extractTile38Metrics(info []string, addr string, alias string, scrapes chan<- scrapeResult) error {
+func (e *Exporter) registerGaugeValue(ch chan<- prometheus.Metric, metric string, val float64, labels ...string) {
```
brian-brazil: The name of this function doesn't make it obvious that this is actually a ConstMetric.

exporter/redis.go
```diff
@@ -552,64 +550,52 @@ func (e *Exporter) handleMetricsCommandStats(addr string, alias string, fieldKey
...
-			e.metrics["connected_slave_lag_seconds"].WithLabelValues(
-				addr,
-				alias,
-				slaveIp,
-				slavePort,
-				slaveState,
-			).Set(lag)
+	if slaveOffset, slaveIP, slavePort, slaveState, slaveLag, ok := parseConnectedSlaveString(fieldKey, fieldValue); ok {
+		e.registerGaugeValue(ch,
+			"connected_slave_offset",
```
brian-brazil: What's the unit for this?

exporter/redis.go
```diff
@@ -936,152 +1017,35 @@ func (e *Exporter) scrapeRedisHost(scrapes chan<- scrapeResult, addr string, idx
...
 			latencyResult := tempVal[0].([]interface{})
-			if _, err := redis.Scan(latencyResult, &eventName, &spikeLast, &milliseconds, &max); err == nil {
-				e.metricsMtx.RLock()
-				e.metrics["latency_spike_last"].WithLabelValues(addr, e.redis.Aliases[idx], eventName).Set(float64(spikeLast))
-				e.metrics["latency_spike_milliseconds"].WithLabelValues(addr, e.redis.Aliases[idx], eventName).Set(float64(milliseconds))
-				e.metricsMtx.RUnlock()
+			var spikeLast, spikeDuration, max int64
+			if _, err := redis.Scan(latencyResult, &eventName, &spikeLast, &spikeDuration, &max); err == nil {
+				spikeDuration = spikeDuration / 1e6
+				e.registerGaugeValue(ch, "latency_spike_last", float64(spikeLast), eventName)
```
brian-brazil: timestamp_seconds usually

oliver006:
Thanks @brian-brazil - that's god feedback.

exporter/redis.go
oliver006: I think bytes. Rename it to `connected_slave_offset_bytes`?

exporter/redis.go
oliver006: 👍 - renamed

exporter/redis.go
oliver006: Changed.

exporter/redis.go
oliver006: Ok, moved that up to the exporter level and made it a summary.

exporter/redis.go
oliver006: copy-pasta, fixed

exporter/redis.go
oliver006: Removed `used_`

exporter/redis.go
oliver006: Removed them, they can be calculated from other metrics.

main.go
```diff
@@ -38,23 +37,19 @@ func getEnvBool(key string) (envValBool bool) {
...
+		listenAddress            = flag.String("web.listen-address", getEnv("REDIS_EXPORTER_WEB_LISTEN_ADDRESS", ":9121"), "Address to listen on for web interface and telemetry.")
+		metricPath               = flag.String("web.telemetry-path", getEnv("REDIS_EXPORTER_WEB_TELEMETRY_PATH", "/metrics"), "Path under which to expose metrics.")
+		logFormat                = flag.String("log-format", getEnv("REDIS_EXPORTER_LOG_FORMAT", "txt"), "Log format, valid options are txt and json")
+		configCommand            = flag.String("config-command", getEnv("REDIS_EXPORTER_CONFIG_COMMAND", "CONFIG"), "What to use for the CONFIG command")
+		isDebug                  = flag.Bool("debug", getEnvBool("REDIS_EXPORTER_DEBUG"), "Output verbose debug information")
+		showVersion              = flag.Bool("version", false, "Show version information and exit")
+		redisMetricsOnly         = flag.Bool("redis-only-metrics", getEnvBool("REDIS_EXPORTER_REDIS_ONLY_METRICS"), "Whether to also export go runtime metrics")
+		inclMetricTotalSysMemory = flag.Bool("incl-metric-total-system-memory", getEnvBool("REDIS_EXPORTER_INCL_METRIC_TOTAL_SYS_MEM"), "Whether to include the redis_total_system_memory_bytes metric")
```
SuperQ: How about a simple `include-system-metrics` flag.

oliver006: Yeah, I like it, let's do that.

exporter/redis.go
SuperQ: Yes, that sounds good.

oliver006: 👍

README.md
SuperQ: Agreed.

I would suggest that for `/scrape` targets that `redis://` is implied. It doesn't make sense to remotely scrape a unix socket. Only allowing `host:port`, would be a good thing to filter for in the input.

oliver006: I don't think we can simply drop the scheme from the target as some instances are using SSL and the `rediss://` scheme is used for that.

But I agree, this is not the best way to pass a password and I'm open for suggestions on how to enable users to scrape password-protected instances.

brian-brazil: What we do elsewhere is have the password come from a config file.

oliver006: > What we do elsewhere is have the password come from a config file.

That's doable what what would that look like then someone wants to scrape several password protected instances use the `/scrape` endpoint?

brian-brazil: We have different modules in the snmp/blackbox exporter for that, which is basically another url parameter.

oliver006: I'm not sure I understand what you mean.

brian-brazil: https://github.com/prometheus/snmp_exporter/blob/master/generator/FORMAT.md

oliver006: Hmm, I see. So it looks like you use the module name to indicate to the exporter which auth profile to use?
I guess I could do something similar. Have a password file deployed with the exporter where each password is associated with an ID and then the prometheus server can send the password_id as a parameter.

SuperQ: I would suggest having one password from the normal command line/env methods to start with and iterate on additional passwords. This way we can get this merged and have a good start on better functionality.

oliver006: Sounds good, let's start with that so this PR can move forward.

main.go
```diff
@@ -145,9 +115,10 @@ func main() {
 		http.Handle(*metricPath, promhttp.Handler())
 	}
 
+	http.HandleFunc("/scrape", exp.ScrapeHandler)
```
zoonage: Isn't `/metrics` the standard endpoint? Also I think most exporters provide this path as an argument

oliver006: Good question! Yes, `/metrics` normally is the standard endpoint and that's what's still used for scraping the exporter metrics. It's configurable via `--web.telemetry-path` and hooked up in main.go line 115.
`/scrape` is what's used for scraping multiple redis instances,  [see the README here](https://github.com/oliver006/redis_exporter/blob/oh_1.0_here_we_come/README.md#prometheus-configuration-to-scrape-several-hosts).

oliver006:
Let's do this.

oliver006:
Github has issues with notifications right now, the exporter binaries and docker image for 1.0.0 will be available as soon as these have been resolved.

roman-vynar:
> The exporter will no longer directly do any Service Discovery and/or scrape multiple redis instances at once.

That's very sad. We are running multiple redis containers on a single instance which means we now need to run multiple redis_exporter processes :(

We use Prometheus and we do Service Discovery but we don't want to run multiple redis_exporters :(

SuperQ:
You don't need multiple exporters. The new version can still scrape multiple redis instances. It is configured more like the blackbox exporter.

oliver006:
@roman-vynar - [see here](https://github.com/oliver006/redis_exporter#prometheus-configuration-to-scrape-several-hosts) for more information how to configure Prometheus to scrape multiple hosts.

amangupta0297:
Hi,

Can anyone help me with redis exporter metrics
redis-exporter version: 1.10.0
exporter running at : 9121
grafana dashboard: 763

prometheus config:
  - job_name: 'redis_exporter'
    static_configs:
    - targets: ['DNS:9121']

target is healthy under prometheus, but still data is not coming over grafana.

When i see, the metrics

curl http://DNS:9121/metrics | grep redis


redis_exporter_build_info{build_date="2018-09-20-18:15:12",commit_sha="8bb0b841e9a70b0348f69483e58fea01d521c47a",golang_version="go1.10.4",version="v0.21.2"} 1
redis_exporter_last_scrape_duration_seconds 0.000292674
redis_exporter_last_scrape_error 1
redis_exporter_scrapes_total 125
redis_up{addr="redis://localhost:6379",alias=""} 0


issue: only very few metrics are coming, because of which data is not coming over grafana.

metrics  required:
1. redis_uptime_in_seconds
2. redis_connected_clients
3. redis_memory_used_bytes/ redis_memory_max_bytes

none of the data is coming, can anyone help. What am i missing

Thanks in advance

PR #263, opened 2019-05-27
Fix include-system-metrics env var name on Readme

tinproject:
Fix include-system-metrics env var name on Readme.


oliver006:
Thanks for the PR!

PR #264, opened 2019-05-30
Support TLS client authentication

joesis:
It adds two features to support scraping `rediss` instances in our setup.

1. Don't remove password when the scheme is `rediss`.
1. Add options to support client certificate.

I'm not sure if it's any useful to other users but if the basic ideas are accepted, I can change tests and README accordingly.


exporter.go
```diff
@@ -179,8 +180,10 @@ func (e *Exporter) ScrapeHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// get rid of username/password info in "target" so users don't send them via http
-	u.User = nil
+	if u.Scheme != "rediss" {
+		// get rid of username/password info in "target" so users don't send them via http
```
joesis: It's unclear to me why `http` is mentioned here. Shouldn't it be RESP(REdis Serialization Protocol)?

oliver006: Ah yes, so this is the core of the problem. 
The Recis URI is transmitted from the Prometheus server to the exporter as part of the http req when scraping the redis instance.
If you need to configure the exporter to use a password then use the command line flag (and yes, right now this will only allow to configure one password across all the instances you intend to scrape. if you need more than use more redis_exporters).

exporter.go
```diff
@@ -967,14 +970,15 @@ func getKeysFromPatterns(c redis.Conn, keys []dbKeyPair) (expandedKeys []dbKeyPa
 	return expandedKeys, err
 }
 
-func (e *Exporter) connectToRedis(skipTLSVerification bool) (redis.Conn, error) {
+func (e *Exporter) connectToRedis(skipTLSVerification bool, tlsClientCertificates []tls.Certificate) (redis.Conn, error) {
```
oliver006: I was actually thinking, maybe we can just get rid of the two parameters to e.connectToRedis() and directly use `e.options.SkipTLSVerification` and `e.options.ClientCertificates` in the function.

exporter.go
oliver006: See here for discussion in the original PR: https://github.com/oliver006/redis_exporter/pull/256#discussion_r283007985

joesis: Oh I know what you mean, it's to remind users to not leak Redis password when scraping. Frankly I doubt in what degree it can prevent users from doing that. Careless people would configure URL with password regardless, those who do care can do things like putting exporter in controlled environment (we run it on the same machine of the Prometheus server), or adding a reverse proxy in front of exporter to terminate TLS.

I know our setup may not be a typical scenario, so I'm good to remove this change, but I feel the ability to batch scraping password protected instances can be useful to some others, so maybe we can see if there's a better idea.

exporter.go
joesis: Good idea. Will do.

exporter.go
oliver006: Right, you got it.
And yes, I there are ways to send an extra piece of config over from the prometheus server to the exporter using the `params` config of the Prometheus server but that's not implemented (yet).

exporter.go
oliver006: Looks good now, thanks!

oliver006:
@joesis - are you still working on this PR? The cert support looks good and, once the README is done, I'd be happy to merge this.

exporter.go
joesis: @oliver006 do you think this is okay? As you mentioned, even though Redis is encrypted, the password can be leaked if the user doesn't have security in mind. IMO a caution in README is sufficient as security is the users' responsibility.

oliver006: I'd remove this. Let's not let people pass a password over the wire. But let's keep your cert changes, they're good.

joesis:
Okay changed README @oliver006.

oliver006:
The last build failed with a rather odd error and it seems like there's stuff missing from your drone file (the redis 2.8 test instance is not up).
Can you make sure your branch is rebased off of origin/master and then kick the drone build to run again? 
Aside from that, I think your change looks pretty good!

joesis:
Hmm I could have sworn that I saw CI passed but that can't be true... Rebased master and all green now!

README.md
```diff
@@ -128,6 +128,8 @@ redis-only-metrics     | REDIS_EXPORTER_REDIS_ONLY_METRICS    | Whether to also
 include-system-metrics | REDIS_EXPORTER_INCL_SYSTEM_METRICS   | Whether to include system metrics like `total_system_memory_bytes`, defaults to false.
 is-tile38              | REDIS_EXPORTER_IS_TILE38             | Whether to scrape Tile38 specific metrics, defaults to false.
 skip-tls-verification  | REDIS_EXPORTER_SKIP_TLS_VERIFICATION | Whether to to skip TLS verification
+tls-client-key-file    | REDIS_EXPORTER_TLS_CLIENT_KEY_FILE   | Full path to the client key file if the server requires TLS client authentication
```
oliver006: Question: your docs say `full path to the client key file`. Does that mean the expected parameter is a path (and the redis lib figures out what the file is) or is it the full name of the client key file?

oliver006: fyi, the text here is somewhat inconsistent with the `--help` output which says `Optional client certificate file...`

oliver006:
Looks good, just one last question regarding the README.

README.md
joesis: It's the path including the file name which is passed to `tls.LoadX509KeyPair`. I should make the doc consistent with the flag description. Which is less ambiguous to you?

oliver006: Yeah, in that case maybe dbe a bit more explicit that it's a file name that's expected: 
e.g. `Name of the client key file (including full path) if the server requires TLS client authentication`

joesis: Updated both the README and flag descriptions!

oliver006: Woooooooot !

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.0.3 - thanks for the PR!

PR #265, opened 2019-05-30
Make grafana dashboard import ask for datasource

kierenevans:
This was lost in the 1.0.0 version when compared to `contrib/grafana_prometheus_redis_dashboard_exporter_version_0.3x.json`.

Thanks for an awesome exporter + dashboard :)


oliver006:
Ah nice, good catch, thanks for the PR!

PR #276, opened 2019-06-18
Update README.md

yes5144:
I try to use like this , finally it works

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,12 @@ scrape_configs:
         target_label: instance
       - target_label: __address__
         replacement: <<REDIS-EXPORTER-HOSTNAME>>:9121
+  ## it's also necessary
+  ## config for scraping the exporter itself
+  - job_name: 'redis_exporter'
+    static_configs:
+      - targets:
+        - <<REDIS-EXPORTER-HOSTNAME>>:9121
 ```
 
 The `targets-redis-instances.json` should look something like this:
@@ -99,7 +105,7 @@ The `targets-redis-instances.json` should look something like this:
 ```json
 [
   {
-    "targets": [ "redis-host-01:6379", "redis-host-02:6379"],
+    "targets": [ "redis://redis-host-ip-01:6379", "redis://redis-host-ip-02:6379"],
     "labels": { }
   }
 ]


README.md
```diff
@@ -92,14 +92,20 @@ scrape_configs:
         target_label: instance
       - target_label: __address__
         replacement: <<REDIS-EXPORTER-HOSTNAME>>:9121
+  ## it's also necessary
```
oliver006: I'd reword this, it's not necessary (for redis instance scraping to work) but probably a very good idea to scrape the exporter itself so how about `It's also highly recommended to scrape the exporter itself` ?

README.md
```diff
@@ -92,14 +92,20 @@ scrape_configs:
...
+  ## config for scraping the exporter itself
+  - job_name: 'redis_exporter'
```
oliver006: This needs another name, we already have `redis_exporter` in line 83.
How about changing line 83 to `redis_instances`?

README.md
```diff
@@ -92,14 +92,20 @@ scrape_configs:
...
+        - <<REDIS-EXPORTER-HOSTNAME>>:9121
 ```
 
 The `targets-redis-instances.json` should look something like this:
 
 ```json
 [
   {
-    "targets": [ "redis-host-01:6379", "redis-host-02:6379"],
+    "targets": [ "redis://redis-host-ip-01:6379", "redis://redis-host-ip-02:6379"],
```
oliver006: Two things?
- first, what was the actual fix? using the `redis://` prefix? It works without that being present so not sure what's going on here
- you changed the name to use the `-ip` suffix but it doesn't need to be the IP, it can be the DNS hostname so I'd prefer to leave it like that to not make people think the actual IP is required.

yes5144: Thank you for your time and patience. 
After resolving the domain name, I try it again.
```
#### these are three examples work well.	
	"targets": [ "redis://127.0.0.1:22222", "redis://127.0.0.1:55555"],
    "targets": [ "redis-dns-name.com:22222", "redis-dns-name.com:55555"],
    "targets": [ "redis://redis-dns-name.com:22222", "redis://redis-dns-name.com:55555"],

#### but the follow example, It can't work
    "targets": [ "127.0.0.1:22222", "127.0.0.1:55555"],

#### error like this	
[root@TEST02 redis]# curl  http://10.1.15.217:9121/scrape?target="127.0.0.1:22222"
Invalid 'target' parameter, parse err: parse 127.0.0.1:22222: first path segment in URL cannot contain colon
```
Thank you very much

README.md
yes5144: It's OK,  thank you

README.md
yes5144: yes, you are right, it's my  careless

README.md
oliver006: I think your curl is wrong because of the quotes, try `curl  "http://10.1.15.217:9121/scrape?target=127.0.0.1:22222"`

PR #293, opened 2019-08-03
Added client list details to use with PromQL

justinasfour04:
I added a function to add CLIENT LIST stats that are group by client name in order to get a count of each client connected instead of just the total.

Please feel free to improve this push request and also please give me your opinion on how it can be improved in order to provide more labels to make this more flexible

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ web.telemetry-path     | REDIS_EXPORTER_WEB_TELEMETRY_PATH    | Path under which
 redis-only-metrics     | REDIS_EXPORTER_REDIS_ONLY_METRICS    | Whether to also export go runtime metrics, defaults to false.
 include-system-metrics | REDIS_EXPORTER_INCL_SYSTEM_METRICS   | Whether to include system metrics like `total_system_memory_bytes`, defaults to false.
 is-tile38              | REDIS_EXPORTER_IS_TILE38             | Whether to scrape Tile38 specific metrics, defaults to false.
-export-client-list     | REDIS_EXPORTER_EXPORT_CLIENT_LIST    | Whether to scrape Client List specific metrics
+export-client-list     | REDIS_EXPORTER_EXPORT_CLIENT_LIST    | Whether to scrape Client List specific metrics, defaults to false.
 skip-tls-verification  | REDIS_EXPORTER_SKIP_TLS_VERIFICATION | Whether to to skip TLS verification
 tls-client-key-file    | REDIS_EXPORTER_TLS_CLIENT_KEY_FILE   | Name of the client key file (including full path) if the server requires TLS client authentication
 tls-client-cert-file   | REDIS_EXPORTER_TLS_CLIENT_CERT_FILE  | Name the client cert file (including full path) if the server requires TLS client authentication
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ web.telemetry-path     | REDIS_EXPORTER_WEB_TELEMETRY_PATH    | Path under which
 redis-only-metrics     | REDIS_EXPORTER_REDIS_ONLY_METRICS    | Whether to also export go runtime metrics, defaults to false.
 include-system-metrics | REDIS_EXPORTER_INCL_SYSTEM_METRICS   | Whether to include system metrics like `total_system_memory_bytes`, defaults to false.
 is-tile38              | REDIS_EXPORTER_IS_TILE38             | Whether to scrape Tile38 specific metrics, defaults to false.
+export-client-list     | REDIS_EXPORTER_EXPORT_CLIENT_LIST    | Whether to scrape Client List specific metrics
 skip-tls-verification  | REDIS_EXPORTER_SKIP_TLS_VERIFICATION | Whether to to skip TLS verification
 tls-client-key-file    | REDIS_EXPORTER_TLS_CLIENT_KEY_FILE   | Name of the client key file (including full path) if the server requires TLS client authentication
 tls-client-cert-file   | REDIS_EXPORTER_TLS_CLIENT_CERT_FILE  | Name the client cert file (including full path) if the server requires TLS client authentication
--- a/exporter.go
+++ b/exporter.go
@@ -306,7 +306,7 @@ func NewRedisExporter(redisURI string, opts ExporterOptions) (*Exporter, error) 
 		"slowlog_length":                       {txt: `Total slowlog`},
 		"start_time_seconds":                   {txt: "Start time of the Redis instance since unix epoch in seconds."},
 		"up":                                   {txt: "Information about the Redis instance"},
-		"connected_clients_details":            {txt: "Details about connected clients", lbls: []string{"host", "port", "name", "age", "idle", "flags", "qbuf", "qbuffree", "cmd"}},
+		"connected_clients_details":            {txt: "Details about connected clients", lbls: []string{"host", "port", "name", "fd", "age", "idle", "flags", "db", "sub", "psub", "multi", "qbuf", "qbuffree", "obl", "oll", "omem", "events", "cmd"}},
 	} {
 		e.metricDescriptions[k] = newMetricDescr(opts.Namespace, k, desc.txt, desc.lbls)
 	}
@@ -392,7 +392,7 @@ func extractVal(s string) (val float64, err error) {
 	id=11 addr=127.0.0.1:63508 fd=8 name= age=6321 idle=6320 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=setex
 	id=14 addr=127.0.0.1:64958 fd=9 name= age=5 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=26 qbuf-free=32742 obl=0 oll=0 omem=0 events=r cmd=client
 */
-func parseClientListString(clientInfo string) (id float64, host string, port string, fd string, name string, age string, idle string, flags string, db string, sub string, psub string, multi string, qbuf string, qbuffree string, obl string, omem string, events string, cmd string, ok bool) {
+func parseClientListString(clientInfo string) (id float64, host string, port string, fd string, name string, age string, idle string, flags string, db string, sub string, psub string, multi string, qbuf string, qbuffree string, obl string, oll string, omem string, events string, cmd string, ok bool) {
 	var err error
 	ok = false
 	if matched, _ := regexp.MatchString(`^id=\d+ addr=\d+`, clientInfo); !matched {
@@ -430,6 +430,7 @@ func parseClientListString(clientInfo string) (id float64, host string, port str
 	qbuf = connectedClient["qbuf"]
 	qbuffree = connectedClient["qbuffree"]
 	obl = connectedClient["obl"]
+	oll = connectedClient["oll"]
 	omem = connectedClient["omem"]
 	events = connectedClient["events"]
 	cmd = connectedClient["cmd"]
@@ -875,10 +876,10 @@ func (e *Exporter) extractTile38Metrics(ch chan<- prometheus.Metric, c redis.Con
 func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
 	if reply, err := redis.String(doRedisCmd(c, "CLIENT", "LIST")); err == nil {
 		clients := strings.Split(reply, "\n")
-		
+
 		for i := 0; i < len(clients)-1; i++ {
-			if id, host, port, fd, name, age, idle, flags, db, sub, psub, multi, qbuf, qbuffree, obl, omem, events, cmd, ok := parseClientListString(clients[i]); ok {
-				e.registerConstMetricGauge(ch, "connected_clients_details", id, host, port, fd, name, age, idle, flags, db, sub, psub, multi, qbuf, qbuffree, obl, omem, events, cmd)
+			if id, host, port, fd, name, age, idle, flags, db, sub, psub, multi, qbuf, qbuffree, obl, oll, omem, events, cmd, ok := parseClientListString(clients[i]); ok {
+				e.registerConstMetricGauge(ch, "connected_clients_details", id, host, port, fd, name, age, idle, flags, db, sub, psub, multi, qbuf, qbuffree, obl, oll, omem, events, cmd)
 			}
 		}
 	}
--- a/exporter.go
+++ b/exporter.go
@@ -59,6 +59,7 @@ type ExporterOptions struct {
 	InclSystemMetrics   bool
 	SkipTLSVerification bool
 	IsTile38            bool
+	IsClientList		bool
 	ConnectionTimeouts  time.Duration
 }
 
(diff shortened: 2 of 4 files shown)


oliver006:
Thanks for the PR! I was away from my computer the last two days but will try to review as soon as possible.

main.go
```diff
@@ -53,6 +53,7 @@ func main() {
 		tlsClientCertFile   = flag.String("tls-client-cert-file", getEnv("REDIS_EXPORTER_TLS_CLIENT_CERT_FILE", ""), "Name of the client certificate file (including full path) if the server requires TLS client authentication")
 		isDebug             = flag.Bool("debug", getEnvBool("REDIS_EXPORTER_DEBUG"), "Output verbose debug information")
 		isTile38            = flag.Bool("is-tile38", getEnvBool("REDIS_EXPORTER_IS_TILE38"), "Whether to scrape Tile38 specific metrics")
+		isClientList        = flag.Bool("is-client-list", getEnvBool("REDIS_EXPORTER_IS_CLIENT_LIST"), "Whether to scrape Client List specific metrics")
```
oliver006: Let's call this (and all the related variables) `export-client-list`

exporter.go
```diff
@@ -819,6 +821,23 @@ func (e *Exporter) extractTile38Metrics(ch chan<- prometheus.Metric, c redis.Con
...
 }
 
+func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
+	if reply, err := redis.String(doRedisCmd(c, "CLIENT", "LIST")); err == nil {
+		clients := strings.Split(reply, "\n")
+		
+		var clientsMap = map[string]int{}
+		for i := 0; i < len(clients) - 1; i++ {
+			clientInfo := strings.Split(clients[i], " ")
+			name := strings.Split(clientInfo[3], "=")[1]
```
oliver006: After calling Split()`, you should check that the resulting array has enough elements before accessing e.g. `[1]`, see e.g. https://github.com/oliver006/redis_exporter/blob/9aef08c64d755d66f92dbc894cbdd3be5a28bd01/exporter.go#L379-L380 for a code example.

justinasfour04:
@oliver006 If I call it connected_clients then the other connected_clients that comes from doing INFO in redis needs to be removed

justinasfour04:
How would I build multiple metrics for the same metric name. The process tends to complain that I can't do that

justinasfour04:
Also why are tests failing?

oliver006:
Ah my bad, I didn't remember that `connected_clients` is already in use.
Maybe go with `connected_clients_details` for the new metrics?

oliver006:
Not sure about the tests, something is going terribly wrong and way too much debug output is produced.
You can run the tests locally but need to bring up a redis instance. You can use the docker-compose in the contrib directory to bring up whatever you need .

justinasfour04:
> Ah my bad, I didn't remember that `connected_clients` is already in use.
> Maybe go with `connected_clients_details` for the new metrics?

justinasfour04:
@oliver006 How would you leverage all the labels and build metrics because right now the gauge metric is for instances of name

justinasfour04:
@oliver006 None of the tests show FAIL why would your tests fail

oliver006:
`checking gofmt ` - this failed, you should go `go fmt` on your code to get the formatting right.

exporter.go
```diff
@@ -305,6 +306,7 @@ func NewRedisExporter(redisURI string, opts ExporterOptions) (*Exporter, error)
 		"slowlog_length":                       {txt: `Total slowlog`},
 		"start_time_seconds":                   {txt: "Start time of the Redis instance since unix epoch in seconds."},
 		"up":                                   {txt: "Information about the Redis instance"},
+		"connected_clients_details":            {txt: "Total number of connected clients", lbls: []string{"host", "port", "name", "age", "idle", "flags", "qbuf", "qbuffree", "cmd" }},
```
oliver006: metrics descr is wrong, still says `Total number of connected clients`, try something like `Details about connected clients`

exporter.go
```diff
@@ -385,6 +387,41 @@ func extractVal(s string) (val float64, err error) {
 	return
 }
 
+/*
+	Valid Examples
```
oliver006: 👍

exporter.go
```diff
@@ -385,6 +387,41 @@ func extractVal(s string) (val float64, err error) {
...
+	for _, kvPart := range strings.Split(clientInfo, " ") {
+		vPart := strings.Split(kvPart, "=")
+		if len(vPart) != 2 {
+			log.Debugf("Invalid format for client list string, got: %s", kvPart)
+			return
+		}
+		connectedClient[vPart[0]] = vPart[1]
+	}
+
+	hostPortString := strings.Split(connectedClient["addr"], ":")
```
oliver006: This needs a `len(hostPortString) == 2` check

exporter.go
```diff
@@ -819,6 +856,23 @@ func (e *Exporter) extractTile38Metrics(ch chan<- prometheus.Metric, c redis.Con
 	}
 }
 
+func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
+	if reply, err := redis.String(doRedisCmd(c, "CLIENT", "LIST")); err == nil {
+		clients := strings.Split(reply, "\n")
+		
+		var clientsMap = map[string]int{}
```
oliver006: Why still aggregate in the exporter now that you have all the pieces of information about a client? 

You should submit the metric with all the labels so the loop in 870-872 is no longer needed and you can call `registerConstMetricGauge()` right away in 866.

exporter_test.go
```diff
@@ -238,6 +238,36 @@ func TestTile38(t *testing.T) {
...
+
+		chM := make(chan prometheus.Metric)
+		go func() {
+			e.Collect(chM)
+			close(chM)
+		}()
+
+		found := false
+		for m := range chM {
+			if strings.Contains(m.Desc().String(), "connected_clients") {
```
oliver006: You need to check for `connected_clients_details`

exporter_test.go
```diff
@@ -238,6 +238,36 @@ func TestTile38(t *testing.T) {
 	}
 }
 
+func TestExportClientList(t *testing.T) {
+	if os.Getenv("TEST_EXPORT_CLIENT_LIST") == "" {
```
oliver006: Remove this, we always want to run the test.

exporter_test.go
```diff
@@ -238,6 +238,36 @@ func TestTile38(t *testing.T) {
...
+		t.Skipf("TEST_EXPORT_CLIENT_LIST not set - skipping")
+	}
+
+	for _, isExportClientList := range []bool{true, false} {
+		e, _ := NewRedisExporter(os.Getenv("TEST_EXPORT_CLIENT_LIST"), ExporterOptions{Namespace: "test", ExportClientList: isExportClientList})
```
oliver006: Reuse the exporter that's available via `TEST_REDIS_URI` instead of introducing a new env variable.

main.go
```diff
@@ -53,6 +53,7 @@ func main() {
 		tlsClientCertFile   = flag.String("tls-client-cert-file", getEnv("REDIS_EXPORTER_TLS_CLIENT_CERT_FILE", ""), "Name of the client certificate file (including full path) if the server requires TLS client authentication")
 		isDebug             = flag.Bool("debug", getEnvBool("REDIS_EXPORTER_DEBUG"), "Output verbose debug information")
 		isTile38            = flag.Bool("is-tile38", getEnvBool("REDIS_EXPORTER_IS_TILE38"), "Whether to scrape Tile38 specific metrics")
+		exportClientList    = flag.Bool("export-client-list", getEnvBool("REDIS_EXPORTER_EXPORT_CLIENT_LIST"), "Whether to scrape Client List specific metrics")
```
oliver006: Let's add this to the README as well.

exporter_test.go
```diff
@@ -238,6 +238,36 @@ func TestTile38(t *testing.T) {
...
+		found := false
+		for m := range chM {
+			if strings.Contains(m.Desc().String(), "connected_clients") {
+				found = true
+				break
+			}
+		}
+
+		if isExportClientList && !found {
+			t.Errorf("connected_clients was *not* found in isExportClientList metrics but expected")
```
oliver006: same, use `connected_clients_details`

exporter_test.go
```diff
@@ -238,6 +238,36 @@ func TestTile38(t *testing.T) {
...
+		} else if !isExportClientList && found {
+			t.Errorf("connected_clients was *found* in isExportClientList metrics but *not* expected")
```
oliver006: same, `use connected_clients_details`

exporter.go
justinasfour04: @oliver006 but I need a count of how many clients share names and such?
How would you do this?

exporter.go
justinasfour04: I don't really want details as much as I want to export number clients sharing labels

exporter.go
oliver006: Once you have all the client details exported you can then get the aggregate in the Prometheus server by using PromQL and do something like `sum(connected_clients_details) by (name)`

exporter.go
oliver006: Sure but other people might find the details useful and/or want to aggregate by host instead of name so this would be useful information.

exporter.go
justinasfour04: Also what metric, I still don't see it

oliver006: The metric would be `connected_clients_details` and the labels would be as defined by you in line 309 of this file.

oliver006: It would look something like

`connected_clients_details{host="10.x.x.x" port="4321" flags="aC"} 1`.

justinasfour04: this function: registerConstMetricGauge takes a float64 value so I still don't know what I'm passing to the function I just don't see it

justinasfour04: I see what you meant and I think I got it

justinasfour04:
@oliver006 1 last thing, we just need to get the test written

justinasfour04:
@oliver006 ready for a final review 👍

exporter.go
oliver006: Yes, this looks a lot better! I'll give this another round of review later today.

exporter.go
```diff
@@ -305,6 +306,7 @@ func NewRedisExporter(redisURI string, opts ExporterOptions) (*Exporter, error)
 		"slowlog_length":                       {txt: `Total slowlog`},
 		"start_time_seconds":                   {txt: "Start time of the Redis instance since unix epoch in seconds."},
 		"up":                                   {txt: "Information about the Redis instance"},
+		"connected_clients_details":            {txt: "Details about connected clients", lbls: []string{"host", "port", "name", "fd", "age", "idle", "flags", "db", "sub", "psub", "multi", "qbuf", "qbuffree", "obl", "oll", "omem", "events", "cmd"}},
```
oliver006: Let's reduce the number of labels a bit, not all of them are super useful and having that many (esp fsasdt changing ones like age, idle, etc) will blow up the namespace a lot.

Let's go with `"host", "port", "name", "flags", "db"` for now.

exporter.go
```diff
@@ -385,6 +387,58 @@ func extractVal(s string) (val float64, err error) {
 	return
 }
 
+/*
+	Valid Examples
+	id=11 addr=127.0.0.1:63508 fd=8 name= age=6321 idle=6320 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=setex
+	id=14 addr=127.0.0.1:64958 fd=9 name= age=5 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=26 qbuf-free=32742 obl=0 oll=0 omem=0 events=r cmd=client
+*/
+func parseClientListString(clientInfo string) (id float64, host string, port string, fd string, name string, age string, idle string, flags string, db string, sub string, psub string, multi string, qbuf string, qbuffree string, obl string, oll string, omem string, events string, cmd string, ok bool) {
```
oliver006: This can now return a lot less results (or simple the map plus the bool for `ok`)

exporter.go
```diff
@@ -819,6 +873,18 @@ func (e *Exporter) extractTile38Metrics(ch chan<- prometheus.Metric, c redis.Con
 	}
 }
 
+func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
+	if reply, err := redis.String(doRedisCmd(c, "CLIENT", "LIST")); err == nil {
+		clients := strings.Split(reply, "\n")
+
+		for i := 0; i < len(clients)-1; i++ {
```
oliver006: Go has a nice way to iterate over slices:
```
for _, c := range clients {
    if ..., ok := parseClientListString(c); ok {
    }
}
```

exporter.go
justinasfour04: I like it

justinasfour04: age is probably good and idle for showing old clients and clients that are doing nothing

exporter.go
```diff
@@ -385,6 +387,48 @@ func extractVal(s string) (val float64, err error) {
...
+	for _, kvPart := range strings.Split(clientInfo, " ") {
+		vPart := strings.Split(kvPart, "=")
+		if len(vPart) != 2 {
+			log.Debugf("Invalid format for client list string, got: %s", kvPart)
+			return
+		}
+		connectedClient[vPart[0]] = vPart[1]
+	}
+
+	if id, err = strconv.ParseFloat(connectedClient["id"], 64); err != nil {
```
oliver006: `id` is no longer used, you can remove this

exporter.go
```diff
@@ -819,6 +863,18 @@ func (e *Exporter) extractTile38Metrics(ch chan<- prometheus.Metric, c redis.Con
 	}
 }
 
+func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
+	if reply, err := redis.String(doRedisCmd(c, "CLIENT", "LIST")); err == nil {
+		clients := strings.Split(reply, "\n")
+
+		for _, c := range clients {
+			if id, host, port, name, age, idle, flags, db, cmd, ok := parseClientListString(c); ok {
+				e.registerConstMetricGauge(ch, "connected_clients_details", id, host, port, name, age, idle, flags, db, cmd)
```
oliver006: Don't set the value of the gauge to `id`, set it to `1.0` (and you can either discard `id` or return it as a string from `parseClientListString() and use it (but if you use it you need to add it as a label in 309) - no strong preference).

In that way you can use `sum()` in PromQL to figure out how many clients per IP, name, etc you have.

README.md
```diff
@@ -127,6 +127,7 @@ web.telemetry-path     | REDIS_EXPORTER_WEB_TELEMETRY_PATH    | Path under which
 redis-only-metrics     | REDIS_EXPORTER_REDIS_ONLY_METRICS    | Whether to also export go runtime metrics, defaults to false.
 include-system-metrics | REDIS_EXPORTER_INCL_SYSTEM_METRICS   | Whether to include system metrics like `total_system_memory_bytes`, defaults to false.
 is-tile38              | REDIS_EXPORTER_IS_TILE38             | Whether to scrape Tile38 specific metrics, defaults to false.
+export-client-list     | REDIS_EXPORTER_EXPORT_CLIENT_LIST    | Whether to scrape Client List specific metrics
```
oliver006: can you please add `, defaults to false.`

exporter.go
justinasfour04: id is used
check extractConnectedClients function

exporter.go
justinasfour04: @oliver006 id is used

justinasfour04: OHHHHHHH why didn't you say something earlier 🤔

oliver006: sooooooooooooooooooorry !

justinasfour04: @oliver006 That's what i was struggling with the whole time

oliver006: :-) nice work getting this to work!

oliver006:
Nice work, thanks for the PR!

PR #308, opened 2019-09-03
Automatically Multi-target to ARM

HurricanKai:
Simply uses Github Actions to multi-target to arm/arm64/amd64 it's a bit of a mess, but that could be fixed by merging the Commits together

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -27,3 +27,4 @@ jobs:
         username: ${{ secrets.DOCKER_USERNAME }}
         password: ${{ secrets.DOCKER_PASSWORD }}
         extra_build_params: --build-arg GOARM=${{ matrix.goarch }}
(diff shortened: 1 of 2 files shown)


oliver006:
I don't think I understand what this PR is about, can you explain please?

oliver006:
Builds are currently handled by drone (for docker builds) and the `build-github-binaries.sh` script (for Github binaries) and I have no plans to move to Github Actions.
If there's an interest/need for arm docker containers than it should be done within that context.

HurricanKai:
@oliver006 I don't know how Drone works, but really all that would be need is to set `GOARCH`, and run 3 times, for arm, arm64, and amd64. Would be greatly appreaciated, as this is the only piece of the redis Helm Chart that is missing for simple installation everywhere.

oliver006:
See #312

PR #312, opened 2019-09-08
ARM docker builds

oliver006:
Trying to address the issue raised in #312

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -84,7 +84,7 @@ steps:
   - name: release-docker-image-scratch
     image: plugins/docker
     settings:
-      tags: "latest,${DRONE_TAG}"
+      tags: "latest,${DRONE_TAG},latest-amd64,${DRONE_TAG}-amd64"
       dockerfile: ./Dockerfile
       repo: oliver006/redis_exporter
       target: scratch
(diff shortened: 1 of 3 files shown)


oliver006:
@HurricanKai - this PR will add docker images `oliver006/redis_exporter-linux-arm64` and `oliver006/redis_exporter-linux-arm`, do the Helm charts also require the `oliver006/redis_exporter-linux-amd64` image or is that redundant?

HurricanKai:
@oliver006 It just references a Version tag, see https://github.com/helm/charts/tree/master/stable/redis-ha#configuration 
I'll create a pull request over there to inform about the `-amd64/arm/arm64` addition.

Once buildx is out of experimental, this could also be merged back together, and just be one image for all platforms, see https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images
Right now this is Ideal, Thank you :smile:

oliver006:
`It just references a Version tag` - but it's publishing to two new repos, not using new tags (the way e.g. node-exporter does it if I understand their setup right), is that ok? Or should it be using tags instead of repos? 

` I'll create a pull request over there to inform about the -amd64/arm/arm64 addition.` - as per my update, there currently is no `-amd64` as that would be the same as the regular release, do you absolutely need a `-amd64` release or is the one without suffix sufficient?

oliver006:
My apologies for the delay here. I updated the PR to publish the ARM docker images to the same repo but tag them with `-arm[64` and `-amd64` - is that more in line with what you mean?

HurricanKai:
Yes, looks good!

oliver006:
Cool, let's see if this works.

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.2.1

And on Docker Hub: https://hub.docker.com/r/oliver006/redis_exporter/tags

oliver006:
Though there might be something wrong, I see the arm images tagged as amd64:

image

Let me investigate what's going on.

oliver006:
This looks better now:

image

I don't have access to a ARM system, if you can give these images a spin and let me know if they work for you that'd be much appreciated.

PR #313, opened 2019-09-19
Make latency spike duration metrics accurately reflect data

kevinfulkerson:
In 0.x versions, the duration of the most recent latency spike was expressed using the `latency_spike_milliseconds` metric. This metric name actually made a lot of sense when compared with the [Redis docs for these values](https://redis.io/topics/latency-monitor#latency-latest). In 1.x versions, this metric seems to have been replaced with a `latency_spike_timestamp_seconds` metric along with a transformation that is applied to the value to try to match the new unit.

The name of the new metric suggests to me that it might have been intended to be a replacement for `latency_spike_last` or something, but that it replaced the wrong metric. At any rate, this change adds the old metric (`latency_spike_milliseconds`) back in, and corrects the transformation for the existing metric to not truncate and to express the value in the advertised unit.

Would like to get your feedback on this, so let me know what you think. I'm down to make more changes if you have any further suggestions. Thanks for your work!

### DIFF
--- a/exporter_test.go
+++ b/exporter_test.go
@@ -187,7 +187,7 @@ func TestLatencySpike(t *testing.T) {
 			// Because we're dealing with latency, there might be a slight delay
 			// even after sleeping for a specific amount of time so checking
 			// to see if we're between +-5 of our expected value
-			if val > float64(TimeToSleep)-5 && val < float64(TimeToSleep)+5 {
+			if val > float64(TimeToSleep)+5 || val < float64(TimeToSleep)-5 {
 				t.Errorf("values not matching, %f != %f", float64(TimeToSleep), val)
 			}
 		}
--- a/exporter_test.go
+++ b/exporter_test.go
@@ -177,15 +177,17 @@ func TestLatencySpike(t *testing.T) {
 	}()
 
 	for m := range chM {
-		if strings.Contains(m.Desc().String(), "latency_spike_milliseconds") {
+		if strings.Contains(m.Desc().String(), "latency_spike_duration_seconds") {
 			got := &dto.Metric{}
 			m.Write(got)
 
-			val := got.GetGauge().GetValue()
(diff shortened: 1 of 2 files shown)


oliver006:
Thanks for the PR, that's definitely a bug.

Two things:  First,  while you're at it, can you please change the metric name, i think `latency_spike_duration_seconds` is more appropriate.
And secondly,  `_seconds` is recommended as per Prometheus best practices for the metric unit suffix so let's use that and not bring back the `_milliseconds` metric.

kevinfulkerson:
Thanks for being so responsive; I think I did what you asked, but let me know if I misinterpreted what you meant and I can fix it up.

exporter_test.go
```diff
@@ -177,15 +177,17 @@ func TestLatencySpike(t *testing.T) {
...
 
-			val := got.GetGauge().GetValue()
+			// The metric value is in seconds, but our sleep interval is specified
+			// in milliseconds, so we need to convert
+			val := got.GetGauge().GetValue() * 1000
 			// Because we're dealing with latency, there might be a slight delay
 			// even after sleeping for a specific amount of time so checking
 			// to see if we're between +-5 of our expected value
-			if val > float64(TimeToSleep)-5 && val < float64(TimeToSleep) {
+			if val > float64(TimeToSleep)+5 || val < float64(TimeToSleep)-5 {
```
oliver006: Minor nit: can you change this to `math.Abs(float64(TimeToSleep) - val) < 5`  ?

oliver006:
👍 
Thanks, will cut a new release in a bit.

kevinfulkerson:
Awesome! Thanks for accepting and being patient (as it is probably obvious I'm not very familiar with Go).

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.2.0

PR #317, opened 2019-10-01
Added alpine docker image tag

mdawar:
To get the latest scratch image we can use the `latest` tag but we can't specify the latest Alpine based image without specifying an exact version for example `v1.2.1-alpine`, so I have added the `alpine` tag which serves the same purpose of `latest` but for Alpine based images.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -144,7 +144,7 @@ steps:
   - name: release-docker-image-alpine-amd64
     image: plugins/docker
     settings:
-      tags: "${DRONE_TAG}-alpine"
+      tags: "alpine,${DRONE_TAG}-alpine"
       dockerfile: ./docker/Dockerfile.amd64
       repo: oliver006/redis_exporter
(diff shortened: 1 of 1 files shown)


oliver006:
Oh weird, I thought we already do that, nice catch.
What do you think about changing `alpine` to `latest-alpine` in your PR to make it clear it's the latest? Is there a convention around that?

mdawar:
@oliver006 I was just following the convention used by most public Docker images especially the official images, for example `golang:alpine`, `golang:stretch`, `python:alpine`, but if you want we can change that to `latest-alpine`, so what do you say?

oliver006:
Sounds good, then let's go with `alpine` as you have it right now.

oliver006:
Sorry for the delay, just released https://github.com/oliver006/redis_exporter/releases/tag/v1.2.2 - the new `alpine` tagged image is up on docker hub now.

PR #322, opened 2019-10-14
better to use kubernetes_pod_name as instanse

alireza1371:
better to use kubernetes_pod_name as instanse in setting variable.

### DIFF
--- a/contrib/grafana_prometheus_redis_dashboard.json
+++ b/contrib/grafana_prometheus_redis_dashboard.json
@@ -1,40 +1,4 @@
 {
-  "__inputs": [
-    {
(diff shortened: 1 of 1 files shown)


mohsenmottaghi:
It`s a good idea :+1:

oliver006:
This change (using kubernetes_pod_name instead of instance) would work for Kubernetes setups only and break for everyone else.
I think a better way to address this would be to either add a note to the docs regarding the dashboard or create a second, k8s-only dashboard (but then we'd need to keep two dashboards updated in case something changes).

oliver006:
Closing...

PR #324, opened 2019-10-25
Fix command calls/sec grafana dashboard panel hover tooltip

justabaka:
The hover tooltip is wrong as it sums up every value which it shouldn't be doing if we want to see a real command call counts. The tooltip type should be set to "individual" to fix the value stacking/summing up issue.

Wrong tooltip:
image

Correct tooltip:
image


oliver006:
Thank for the PR, looks good!

PR #332, opened 2019-11-14
Add -redis.file option back to support monitoring of multiple Redis dbs

### LINKED ISSUE #278: Scrape multiple redis instances with different passwords
**Describe the problem**

[ ] 0.3x.x
[ ] 1.x.x

**Running the exporter**

**Expected behavior**
What metrics are missing? What metrics are wrong? Is something missing that was present in an earlier version? 
Did you upgrade from 0.3x.x to 1.0 and are scraping multiple hosts? [Have a look here ](https://github.com/oliver006/redis_exporter#prometheus-configuration-to-scrape-multiple-redis-hosts) how the configuration changed.

**Screenshots**

**Additional context**

roman-vynar:
Fixes #278

This makes redis_exporter great again.

The changes made are fully backward-compatible with the current functionality. I mean it will still behave the same when you pass -redis.addr or try to access exporter using `/scrape?target=<dynamic host>`.

Again:
1. It does neither harm nor change the current behaviour/performance.
2. It allows you to monitor multiple instances w/o running multiple exporters.
3. A single password for multiple instances or no password, passing passwords from prometheus are all insecure approaches.
4. Maintaining prometheus discovery + redis_exporter mapping host<>password doubles the operational overhead, thus `-redis.file` is an old good option.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ and adjust the host name accordingly.
 
 ### Prometheus Configuration to Scrape Multiple Redis Hosts
 
-Run the exporter with the command line flag `--redis.addr=` so it won't try to access 
+Run the exporter with the command line flag `--redis.addr=` so it won't try to access
 the local instance every time the `/metrics` endpoint is scraped.
 
 ```yaml
@@ -63,7 +63,7 @@ scrape_configs:
         target_label: instance
       - target_label: __address__
         replacement: <<REDIS-EXPORTER-HOSTNAME>>:9121
-  
+
   ## config for scraping the exporter itself
   - job_name: 'redis_exporter'
     static_configs:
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR but I'm not going to bring back the functionality to monitor multiple redis instances by providing config options to the exporter (e.g. a file or cmd line params).

There are other ways to do this, among them the solution sketched out by Brian in the PR [here](https://github.com/oliver006/redis_exporter/pull/256#discussion_r283541885) which IMO is preferable.

roman-vynar:
No problem. 

As mentioned above, the solution suggested introduces more operational overhead and does not solve the password problem.

PR #344, opened 2020-01-27
fix(grafana_prometheus_redis_dashboard): Use alias for instance list

desaintmartin:
Proposition: would that be acceptable?

Of course, if it is acceptable, I would change all queries.

### DIFF
--- a/contrib/grafana_prometheus_redis_dashboard.json
+++ b/contrib/grafana_prometheus_redis_dashboard.json
@@ -122,7 +122,7 @@
       "tableColumn": "",
       "targets": [
         {
-          "expr": "max(max_over_time(redis_uptime_in_seconds{instance=~\"$instance\"}[$__interval]))",
+          "expr": "max(max_over_time(redis_uptime_in_seconds{job=~\"$job\"}[$__interval]))",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 2,
@@ -212,7 +212,7 @@
       "tableColumn": "",
       "targets": [
         {
-          "expr": "redis_connected_clients{instance=~\"$instance\"}",
+          "expr": "redis_connected_clients{job=~\"$job\"}",
           "format": "time_series",
           "intervalFactor": 2,
           "legendFormat": "",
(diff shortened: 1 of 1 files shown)


oliver006:
Why the change from `instance` to `job` ?

desaintmartin:
Using prometheus-operator on a kubernetes cluster with https://github.com/helm/charts/tree/master/stable/redis, "instance" is the Pod IP. Using service or job seems to do the trick, but I wonder if this is the only case where `instance` is the ip.
what are your thoughts on this subject?

oliver006:
That seems to be a problem/use case specific to Kubernetes so I'd rather not make the change to the dashboard.
One possibility would be to have a second json with the Kubernetes version of the dashboard (similar to how there is a 0.3x and 1.x exporter version of the dashboard).

naseemkullah:
Yes please! 

Maybe we can have one json but add instructions to replace `instance` with `job` using `sed`.

naseemkullah:
Actually `job` does not work for me as they are all collected via the `kubernetes-service-endpoints` job.

Perhaps the proper solution is to perform instance relabelling.

naseemkullah:
So I've been playing around with prometheus config and can confirm that instance relabelling works and as per @brian-brazil's advice. 

For example, if the metrics are being scraped via the [pod role](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#pod), one could add:

```
          - source_labels: [__meta_kubernetes_pod_name]
            action: replace
            target_label: instance ####
            regex: (.*redis.*)
```

as a relabel config to corresponding scrape config.

naseemkullah:
So I've been playing around with prometheus config and can confirm that instance relabelling works as per @brian-brazil's advice.

For example, if the metrics are being scraped via the pod role, one could add:

          - source_labels: [__meta_kubernetes_pod_name]
            action: replace
            target_label: instance ####
            regex: (.*redis.*)
as a relabel config to corresponding scrape config.

naseemkullah:
So I've been playing around with prometheus config and can confirm that instance relabelling works and as per @brian-brazil's advice.

For example, if the metrics are being scraped via the pod role, one could add:

```
          - source_labels: [__meta_kubernetes_pod_name]
            action: replace
            target_label: instance
            regex: (.*redis.*)
```
as a relabel config to the corresponding scrape config. As per the regex value, only pods with "redis" in their name will be relabelled as such.

naseemkullah:
So I've been playing around with prometheus config and can confirm that instance relabelling works as per @brian-brazil's advice.

For example, if the metrics are being scraped via the pod role, one could add:

```
          - source_labels: [__meta_kubernetes_pod_name]
            action: replace
            target_label: instance
            regex: (.*redis.*)
```
as a relabel config to the corresponding scrape config. As per the regex value, only pods with "redis" in their name will be relabelled as such.

oliver006:
Closing this as there hasn't been any activity in ~2mo.

PR #349, opened 2020-02-19
loop latency results

szook:
the `latency latest` redis command can return multiple results:

```
127.0.0.1:6379> latency latest
1) 1) "fork"
   2) (integer) 1582137796
   3) (integer) 244
   4) (integer) 318
2) 1) "expire-cycle"
   2) (integer) 1582134822
   3) (integer) 8
   4) (integer) 181
3) 1) "fast-command"
   2) (integer) 1582137383
   3) (integer) 58
   4) (integer) 567
4) 1) "command"
   2) (integer) 1582137381
   3) (integer) 231
   4) (integer) 1230
```

But the current redis_exporter code was only showing metrics on the first element of that list.

```
$ curl --silent localhost:9101/metrics | grep ^redis_latency
redis_latency_spike_duration_seconds{event_name="fork"} 0.244
redis_latency_spike_last{event_name="fork"} 1.582137796e+09
```

This change iterates that list and returns all the elements:

```
$ curl --silent localhost:9101/metrics | grep ^redis_latency
redis_latency_spike_duration_seconds{event_name="command"} 0.231
redis_latency_spike_duration_seconds{event_name="expire-cycle"} 0.008
redis_latency_spike_duration_seconds{event_name="fast-command"} 0.058
redis_latency_spike_duration_seconds{event_name="fork"} 0.244
redis_latency_spike_last{event_name="command"} 1.582137381e+09
redis_latency_spike_last{event_name="expire-cycle"} 1.582134822e+09
redis_latency_spike_last{event_name="fast-command"} 1.582137383e+09
redis_latency_spike_last{event_name="fork"} 1.582137796e+09

```

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -898,15 +898,16 @@ func (e *Exporter) extractSlowLogMetrics(ch chan<- prometheus.Metric, c redis.Co
 }
 
 func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
-	if reply, err := doRedisCmd(c, "LATENCY", "LATEST"); err == nil {
-		var eventName string
-		if tempVal, _ := reply.([]interface{}); len(tempVal) > 0 {
-			latencyResult := tempVal[0].([]interface{})
-			var spikeLast, spikeDuration, max int64
-			if _, err := redis.Scan(latencyResult, &eventName, &spikeLast, &spikeDuration, &max); err == nil {
-				spikeDurationSeconds := float64(spikeDuration) / 1e3
-				e.registerConstMetricGauge(ch, "latency_spike_last", float64(spikeLast), eventName)
-				e.registerConstMetricGauge(ch, "latency_spike_duration_seconds", spikeDurationSeconds, eventName)
+	if reply, err := redis.Values(doRedisCmd(c, "LATENCY", "LATEST")); err == nil {
+		for _, l := range reply {
+			if latencyResult, err := redis.Values(l, nil); err == nil {
+				var eventName string
+				var spikeLast, spikeDuration, max int64
+				if _, err := redis.Scan(latencyResult, &eventName, &spikeLast, &spikeDuration, &max); err == nil {
+					spikeDurationSeconds := float64(spikeDuration) / 1e3
+					e.registerConstMetricGauge(ch, "latency_spike_last", float64(spikeLast), eventName)
+					e.registerConstMetricGauge(ch, "latency_spike_duration_seconds", spikeDurationSeconds, eventName)
+				}
 			}
 		}
 	}


oliver006:
Oh nice, thanks for the PR!
I'm a little slammed with work right now but will get to it in the next couple of days.

exporter.go
```diff
@@ -901,12 +901,14 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
 	if reply, err := doRedisCmd(c, "LATENCY", "LATEST"); err == nil {
```
oliver006: make this `redis.Values(doRedisCmd( ... ` and then you won't need line 903 at all and can do the range over `reply` right away. 

While you're at it, you could improve two more things:
- use another call to `redis.Values()` (with err check) in line 905 instead of the current cast
- move eventName into the loop to reduce the scope

szook: I made the changes you suggested.  Is that what you were thinking?

exporter.go
```diff
@@ -898,13 +898,12 @@ func (e *Exporter) extractSlowLogMetrics(ch chan<- prometheus.Metric, c redis.Co
...
 
 func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
-	if reply, err := doRedisCmd(c, "LATENCY", "LATEST"); err == nil {
-		var eventName string
-		if tempVal, _ := reply.([]interface{}); len(tempVal) > 0 {
-			latencyResult := tempVal[0].([]interface{})
-			var spikeLast, spikeDuration, max int64
-			if _, err := redis.Scan(latencyResult, &eventName, &spikeLast, &spikeDuration, &max); err == nil {
-				spikeDurationSeconds := float64(spikeDuration) / 1e3
+	if reply, err := redis.Values(doRedisCmd(c, "LATENCY", "LATEST")); err == nil {
```
oliver006: 👍

exporter.go
```diff
@@ -898,13 +898,12 @@ func (e *Exporter) extractSlowLogMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+		for _, l := range reply {
+			if latencyFields, err := redis.Values(l.([]interface{}), nil); err == nil {
+				eventName := (string)(latencyFields[0].([]byte))
```
oliver006: I prefer the use of `redis.Scan()` over directly casting the variables as it error checks and is safer.

szook: I think I misunderstood what you meant by "use another call to redis.Values() (with err check) in line 905 instead of the current cast".  
Is this more of what you had in mind?

exporter.go
oliver006: Yeah, this looks good, thanks for changing it.

exporter.go
oliver006: Yup, this looks good, thanks!

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.4.0

PR #350, opened 2020-02-19
[dashboard] Add command related panels

naseemkullah:
Hello!

This PR adds panels for Average Time Spent by Command / sec, which takes the total time spent and divides it by the number of calls by command, as well as Total Time Spent by Command / sec, which calculates the  per second rate of total time by command. Furthermore the previously name Command Calls / sec panel is renamed to Total Commands / sec and its' stack graph which was previously cumulative is now individual which yields better results. Finally the dashboards tool tip has been changed to shared crosshair as to more easily correlate the various panels.

Please let me know what you think!

### DIFF
--- a/contrib/grafana_prometheus_redis_dashboard.json
+++ b/contrib/grafana_prometheus_redis_dashboard.json
@@ -53,7 +53,7 @@
   "gnetId": 763,
   "graphTooltip": 1,
   "id": null,
-  "iteration": 1582163289876,
+  "iteration": 1582163289879,
   "links": [],
   "panels": [
     {
@@ -335,7 +335,7 @@
       "datasource": "${DS_PROMETHEUS}",
       "editable": true,
       "error": false,
-      "fill": 1,
+      "fill": 8,
       "fillGradient": 0,
       "grid": {},
       "gridPos": {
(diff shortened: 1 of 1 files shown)


oliver006:
Nice, thanks for the PR!
Couple of quick questions:
- `Total Time Spent by Command / sec` - what's the use case of this panel?
- `Average Time Spent by Command / sec` - this is stacked, does that make sense? What's teh benefit of seeing the cumulative time spend on across all comands? Should this be a regular, non-stacked graph instead?
- I still see `Average Time Spent by Command / sec` on top of the dashboard and then the stacked `Total Commands / sec` - i think the stacked can (and should!) replace the one on top instead of having two, what do you think?

naseemkullah:
1. It's nice to see at a glance what requests are taking up the most time. Also if 1 sort of command is executed a billion times but it very quick, that could still be an issue. In this example exec dominates this panel, but is not really obvious in the other two:

image


2. Yes!

3. Yes! Should I move it up top or keep command related panels together?

oliver006:
`Should I move it up top` - yeah, move it up top, i think it's a very key metric that should be prominently displayed at the top.

naseemkullah:
Alright I've made the changes, as for the `Total Time Spent by Command / sec` panel, are you OK with keeping it? I get that it is essentially `Average Time Spent by Command / sec` * `Total Commands / s` but as per my pasted image up top, results can be surprising when command that takes up most time is not necessarily the most called nor the slowest on average.

naseemkullah:
Lastly, I've added a Blocked Clients line to the Redis Client graph panel, and renamed it accordingly. :)

oliver006:
This looks really good - nice work!

One last nit: the labels for the "Connected/Blocked Clients" panel should be "connected" and "blocked".

naseemkullah:
Thanks! ✅

naseemkullah:
Hey @oliver006 , on second thought... should we maybe hide the legend for this panel?

image

oliver006:
Yeah, probably. Showing that many labels makes the graph not very useful. Is there a way in Gaafana to limit the number of labels shown?

PR #351, opened 2020-02-20
Update readme

naseemkullah:
* Lint as per https://github.com/DavidAnson/markdownlint/blob/v0.19.0/doc/Rules.md
* Add section on using getting human readable names in drop down when using k8s SD


naseemkullah:
Thanks, done!

naseemkullah:
My pleasure!

PR #354, opened 2020-02-24
[Dashboard] Refactor to allow viewing of multiples Redises (Redi?) simultaneously.

### LINKED ISSUE #352: Dashboard question: Viewing overall sentinel cluster metrics?
Hi @oliver006 ,

Currently we have a few redis sentinels deployed where each redis runs alongside an exporter. In the dashboard today it is not possible to see all cluster members at the same time in the dashboard

Is having the drop down as a `multi` a possibility? 
👎 This does make the process of switching redises more more cumbersome (uncheck, check, uncheck, check)
 👍But allows us to see a cluster (check all members of cluster).

If not, would you have a suggested approach? Perhaps there is an exporter config/metrics I am unaware of that allows us to see performance of the cluster as whole in the dashboard.

naseemkullah:
To view a redis cluster as a whole, the drop down is converted to a multi value. 

Hi @oliver006, this is a first pass, let me know what you think and if you are still on board with this change? 

As per above, opted for metrics by instance for the most part, but cannot quite get Connected/Blocked Clients nor Average Time Spent by Command panels to display ANYTHING. Thoughts?

fixes #352

Oh I also went and hid the legend for Total Commands up top as per our other convo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,10 @@ Example [Grafana](http://grafana.org/) screenshots:
 
 Grafana dashboard is available on [grafana.com](https://grafana.com/dashboards/763) and/or [github.com](contrib/grafana_prometheus_redis_dashboard.json).
 
+### Viewing multiple Redis simultaneously
+
+If running [Redis Sentinel](https://redis.io/topics/sentinel), it may be desirable to view the metrics of the various cluster members simultaneously. For this reason the dashboard's drop down is of the multi-value type, allowing for the selection of multiple Redis. Please note that there is a  caveat; the single stat panels up top namely `uptime`, `total memory use` and `clients` do not function upon viewing multiple Redis.
+
 ## Communal effort
 
 Open an issue or PR if you have more suggestions, questions or ideas about what to add.
--- a/contrib/grafana_prometheus_redis_dashboard.json
+++ b/contrib/grafana_prometheus_redis_dashboard.json
@@ -53,7 +53,7 @@
   "gnetId": 763,
   "graphTooltip": 1,
   "id": null,
-  "iteration": 1583760176275,
+  "iteration": 1583850456553,
   "links": [],
   "panels": [
     {
@@ -79,7 +79,7 @@
       },
       "gridPos": {
         "h": 7,
-        "w": 2,
+        "w": 3,
         "x": 0,
         "y": 0
       },
@@ -133,7 +133,7 @@
         }
       ],
       "thresholds": "",
-      "title": "Uptime",
+      "title": "Max Uptime",
       "type": "singlestat",
       "valueFontSize": "70%",
       "valueMaps": [
@@ -169,7 +169,7 @@
       "gridPos": {
         "h": 7,
         "w": 2,
-        "x": 2,
+        "x": 3,
         "y": 0
       },
       "hideTimeOverride": true,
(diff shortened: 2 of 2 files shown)


oliver006:
Thanks for the PR - I'm a little swamped right now but hope to get to it in the next couple of days.

naseemkullah:
Sounds good, no rush and thanks for letting me know.

oliver006:
I just had a quick look at the dashboard, bunch of comments:
- in general I think you're better of to sum the stats across the redis instances instead of e.g. listing cmd usage per instance (that seems pretty useless and hard to read)
  for commands the query would look something like this:
  `topk( 10, sum(  irate(redis_commands_total{instance=~"$instance"} [1m])) by (cmd )  )`
- couple of metrics are not right, e.g. the query for connected clients should be `redis_connected_clients{instance=~"$instance"}` (use `=~` instead of `=` ) but probably should just use sum()
- the single stat panels error out, you could consider using min or max or something but i think there's always an issue as it's not clear what data is shown.

naseemkullah:
Good catches, thanks! Stay tuned for a commit addressing the above.

oliver006:
Where applicable, the number of digits of displayed numbers (e.g. items in a DB) should be zero, no need to show fractions.

naseemkullah:
> I just had a quick look at the dashboard, bunch of comments:
> 
> * in general I think you're better of to sum the stats across the redis instances instead of e.g. listing cmd usage per instance (that seems pretty useless and hard to read)
>   for commands the query would look something like this:
>   `topk( 10, sum(  irate(redis_commands_total{instance=~"$instance"} [1m])) by (cmd )  )`

* Done for redis commands.
* Done for Network I/O
* Hits/Misses/s and Memory used graphs seemed readable enough, but let me know if they should change.
* I added a `!= 0` for iterms per db to increase readability.

> * couple of metrics are not right, e.g. the query for connected clients should be `redis_connected_clients{instance=~"$instance"}` (use `=~` instead of `=` ) but probably should just use sum()
Thanks! I was having trouble figuring why they did not render.

> * the single stat panels error out, you could consider using min or max or something but i think there's always an issue as it's not clear what data is shown.

* Sum of all clients used for # of clients
* Max used for memory
* Max used for uptime


I'll go on to work on the values displayed as fractions issue.

naseemkullah:
> Where applicable, the number of digits of displayed numbers (e.g. items in a DB) should be zero, no need to show fractions.

I'm not sure I understand values of rates (hits/misses and expired/ecvicted keys) should be rounded?

DB Items look good as is.

oliver006:
Nice, i think you're almost there.

I think the "Total Commands/sec" panel query is not right, sometimes i see the graph looking funny and not 


image

while really it should be more like this:

image

I think the topk part is causing it to not be correct. Not 100% sure how it works, if it looks at every data point along the way but you could just drop it.
`sum(rate(redis_commands_total{instance=~"$instance"} [1m])) by (cmd)`

Can you remove the legend on the right side from the panel "total items per DB" ?

For the memory usage single value panel, you should be able to use the sum of all selected redis instances (and that value makes sense) so maybe do that?
And if you rename "Uptime" to "Max Uptime" then it's more clear it's the max value.

naseemkullah:
Thanks for the feedback, we will get there slowly but surely! I've applied those changed, I had to resize the single stats a little to have "Max Uptime" displayed. Please take another look and let me know what you think!

oliver006:
Looking good - nice work, thank you!

naseemkullah:
Thanks for guiding me through this as usual!

PR #384, opened 2020-05-11
add instance failover metrics

vacheli:
Currently, If one of master in Redis cluster failover,There is no way to configure alarms by metric.


oliver006:
Can you explain more why this is needed? 
What is it you're trying to do and how does the current set of metrics fall short? 

Also, there are a bunch of unrelated files checked in (.container and a Dockerfile)

Thanks!

PR #390, opened 2020-06-04
Add COUNT option for SCAN command

### LINKED ISSUE #381: How to config COUNT in the "check-keys" command line flag,just like the option in SCAN(SCAN cursor [MATCH pattern] [COUNT count])
**Describe the problem**
I have known the exporter support key patterns in the "check-keys" command line flag,but there is no way to config COUNT,just like the option in SCAN(SCAN cursor [MATCH pattern] [COUNT count]).How can I do it ?

[ ] 0.3x.x
[ ] 1.x.x

**Running the exporter**
Please include details about env variables, command line parameters, your orchestration setup, etc.

**Screenshots**
If applicable, add screenshots to help explain your question.

**Additional context**
Add any other context about the question here.

houstonheat:
Fixes #381 (Help wanted)
  * Add `COUNT` support and its number validation
  * Add `db` index validation


exporter.go
```diff
@@ -913,7 +940,7 @@ func (e *Exporter) extractCheckKeyMetrics(ch chan<- prometheus.Metric, c redis.C
 	log.Debugf("allKeys: %#v", allKeys)
 	for _, k := range allKeys {
 		if _, err := doRedisCmd(c, "SELECT", k.db); err != nil {
-			log.Debugf("Couldn't select database %#v when getting key info.", k.db)
+			log.Errorf("Couldn't select database %#v when getting key info.", k.db)
```
houstonheat, line 943: I think we should let users know about cases where their patterns won't be scaped because of the wrong database.

exporter.go
```diff
@@ -118,29 +120,53 @@ func (e *Exporter) scrapeHandler(w http.ResponseWriter, r *http.Request) {
...
 		}
 
-		keys = append(keys, dbKeyPair{db, key})
+		// We want to guarantee at the top level that invalid values will not fall into the final Redis call
+		dbInt, err := strconv.Atoi(db)
+		if (dbInt < 0) || (err != nil) {
+			return keys, fmt.Errorf("db index '%s' should be an non-negative integer value in '%s'", db, k)
+		}
+		countInt, err := strconv.Atoi(count)
+		if (countInt <= 0) || err != nil {
```
houstonheat, line 165: Should i rewrite those statements to separated blocks?

oliver006:
Thanks for the PR!
I'll try to get to it in the next couple of days as I'm a little slammed with work right now.

houstonheat:
Don't mean to rush anyone, but still waiting for the review 🌚 
Appreciate any updates from you @oliver006 ))

oliver006:
My apologies @houstonheat  - I didn't have a chance to get to this PR yet. 
I hope to have an in-depth look this week and give you feedback.

exporter.go
oliver006: Yeah, that makes sense.

exporter_test.go
```diff
@@ -1,8 +1,8 @@
 package main
 
 /*
-  to run the tests with redis running on anything but localhost:6379 use
-  $ go test   --redis.addr=<host>:<port>
+  to run the tests with running redis instance use
+  $ TEST_REDIS_URI=redis://<host>:<port> go test
```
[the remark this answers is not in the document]
oliver006: 👍

oliver006 (dismissed):
My apologies for having taken so long with the review, my last four weeks have been rather hectic.

As for the general implementation, I think you can simplify your approach by having one top-level command line parameter (e.g. --check-keys-scan-count=10) that lets you set the COUNT value for all the check-keys items. I don't think there's a need to have this configurable on a per check-keys item basis but could be wrong, curious to hear your opinion.
That approach would also simplify the syntax a lot.

houstonheat:
> ... I think you can simplify your approach by having one top-level command line parameter (e.g. --check-keys-scan-count=10) that lets you set the COUNT value for all the check-keys items. That approach would also simplify the syntax a lot.

Yeah, you are right. I've done it in the first implementation of this feature, but on development instance testing I've realized that in some cases it's ok to set different `COUNT` values.

I've thought that it's a more flexible solution to use different `COUNT` values and it's always based on your real production load. Let's say we have a couple of patterns: `web_socket:*` and `notification:*`
On our load, we can find several hundreds of sockets overall, but thousands of notifications at one single moment. To speed up this op, I've set  `COUNT 100` for sockets and `COUNT 5000` _(similar to average value)_ for notifications. The time difference for huge key patterns makes sense _(in milliseconds)_, but whole `/metrics` call still can be slow because of `TYPE` and `LLEN` on each key, so the call time does not decrease at all. And I can set `5000` to "default" `COUNT` value for the same result.

The two things I'm afraid of in this case it's **RAM limits** and **freeze**. Still Redis is a single-threaded app, huge default `COUNT` can affect production environment. But... it's similar to multiple huge COUNTs)))

As a result, I'm not sure about single option too, but now it's looks much easier and more useful, clear for me. 
What do @fengyinqiao think, as a requester?)

fengyinqiao:
> > ... I think you can simplify your approach by having one top-level command line parameter (e.g. --check-keys-scan-count=10) that lets you set the COUNT value for all the check-keys items. That approach would also simplify the syntax a lot.
> 
> Yeah, you are right. I've done it in the first implementation of this feature, but on development instance testing I've realized that in some cases it's ok to set different `COUNT` values.
> 
> I've thought that it's a more flexible solution to use different `COUNT` values and it's always based on your real production load. Let's say we have a couple of patterns: `web_socket:*` and `notification:*`
> On our load, we can find several hundreds of sockets overall, but thousands of notifications at one single moment. To speed up this op, I've set `COUNT 100` for sockets and `COUNT 5000` _(similar to average value)_ for notifications. The time difference for huge key patterns makes sense _(in milliseconds)_, but whole `/metrics` call still can be slow because of `TYPE` and `LLEN` on each key, so the call time does not decrease at all. And I can set `5000` to "default" `COUNT` value for the same result.
> 
> The two things I'm afraid of in this case it's **RAM limits** and **freeze**. Still Redis is a single-threaded app, huge default `COUNT` can affect production environment. But... it's similar to multiple huge COUNTs)))
> 
> As a result, I'm not sure about single option too, but now it's looks much easier and more useful, clear for me.
> What do @fengyinqiao think, as a requester?)
I agree with you.It's more flexible to use  different COUNT values for each SCAN based on user's real production load,but it's no need to set "default" COUNT value,it's up to the user themselves.

fengyinqiao:
> > ... I think you can simplify your approach by having one top-level command line parameter (e.g. --check-keys-scan-count=10) that lets you set the COUNT value for all the check-keys items. That approach would also simplify the syntax a lot.
> 
> Yeah, you are right. I've done it in the first implementation of this feature, but on development instance testing I've realized that in some cases it's ok to set different `COUNT` values.
> 
> I've thought that it's a more flexible solution to use different `COUNT` values and it's always based on your real production load. Let's say we have a couple of patterns: `web_socket:*` and `notification:*`
> On our load, we can find several hundreds of sockets overall, but thousands of notifications at one single moment. To speed up this op, I've set `COUNT 100` for sockets and `COUNT 5000` _(similar to average value)_ for notifications. The time difference for huge key patterns makes sense _(in milliseconds)_, but whole `/metrics` call still can be slow because of `TYPE` and `LLEN` on each key, so the call time does not decrease at all. And I can set `5000` to "default" `COUNT` value for the same result.
> 
> The two things I'm afraid of in this case it's **RAM limits** and **freeze**. Still Redis is a single-threaded app, huge default `COUNT` can affect production environment. But... it's similar to multiple huge COUNTs)))
> 
> As a result, I'm not sure about single option too, but now it's looks much easier and more useful, clear for me.
> What do @fengyinqiao think, as a requester?)

Sorry to comment on this question so late as a requester,because I am too busy with my work...

I agree with you.It's more flexible to use  different COUNT values for each SCAN based on user's real production load.But it's no need to set "default" COUNT value,it's up to the user themselves.Huge default COUNT value will block and occupy the single thread of Redis for long time,multiple scans can make user's request to Redis processed at least.

houstonheat:
What would the next step be?

houstonheat:
What would be the next step?
@oliver006 need your final opinion and I can continue with this issue (+ add CONT to new check-streams* options)

houstonheat:
I will try to update this in a couple of days to sync with updated code structure ⏳

oliver006:
Sorry for the very very late response, my apologies.

Before you go through the pains of rebasing this I'd suggest some changes.

I really don't like the three-values notation (`db0=key-pattern*=count50`)
I suggest going with one flag to set the COUNT value which will then be used for all patterns.
If something more nuanced is needed in the future we can think about adding it later or suggest using a lua script instead.

houstonheat:
🔝 According to build logs, there is no issue in my PR. Apparently, the `redis-sentinel` container didn't resolve `redis6` from time to time in the Drone-CI environment before: https://cloud.drone.io/oliver006/redis_exporter/1120/1/11
image

Local tests passed successfully, coverage increased too. Waiting for review.

exporter/exporter_test.go
```diff
@@ -277,6 +278,14 @@ func TestConnectionDurations(t *testing.T) {
 	}
 }
 
+// TODO Rework keys clean up.
+//      Clean all on startup or force `docker-compose down && up` each `make docker-test`
```
houstonheat, line 282: I wasn't sure if I had the right to edit the Makefile myself if this case.

main.go
```diff
@@ -65,7 +65,8 @@ func main() {
...
 		checkStreams            = flag.String("check-streams", getEnv("REDIS_EXPORTER_CHECK_STREAMS", ""), "Comma separated list of stream-patterns to export info about streams, groups and consumers, searched for with SCAN")
 		checkSingleStreams      = flag.String("check-single-streams", getEnv("REDIS_EXPORTER_CHECK_SINGLE_STREAMS", ""), "Comma separated list of single streams to export info about streams, groups and consumers")
-		countKeys               = flag.String("count-keys", getEnv("REDIS_EXPORTER_COUNT_KEYS", ""), "Comma separated list of patterns to count, eg: 'db3=sessions:*'. Warning: The exporter runs SCAN to count the keys.")
+		countKeys               = flag.String("count-keys", getEnv("REDIS_EXPORTER_COUNT_KEYS", ""), "Comma separated list of patterns to count (eg: 'db0=production_*,db3=sessions:*'), searched for with SCAN")
+		checkKeysBatchSize      = flag.Int64("check-keys-batch-size", getEnvInt64("REDIS_EXPORTER_CHECK_KEYS_BATCH_SIZE", 1000), "Approximate number of keys to process in each execution, larger value speeds up scanning.\nWARNING: Still Redis is a single-threaded app, huge COUNT can affect production environment.")
```
houstonheat, line 69: New application flag added.
I am interested in discussing adding a shared flag for `checkKeysBatchSize` and `checkKeyGroupsBatchSize`. It would be a breaking change, but we can set up a fallback rule for key-group to use the `checkKeysBatchSize` value if `checkKeyGroupsBatchSize` isn't specified.

Glad to hear @rmak-cpi as an author of `memory-usage metrics` functionality where the COUNT option was involved.

oliver006: I think one (shared) flag for both is fine.

exporter/key_groups.go
```diff
@@ -102,6 +102,7 @@ func (e *Exporter) gatherKeyGroupsMetricsForAllDatabases(c redis.Conn, dbCount i
 	if len(keyGroupsNoEmptyStrings) == 0 {
 		return allMetrics
 	}
+	log.Debugf("keyGroups: %s", keyGroupsNoEmptyStrings)
```
oliver006, line 105: Want to keep this?

exporter/keys.go
```diff
@@ -188,38 +139,54 @@ func (e *Exporter) extractCountKeysMetrics(ch chan<- prometheus.Metric, c redis.
...
+	return keysCount, nil
+}
 
-		if iter, _ = redis.Int(arr[0], nil); iter == 0 {
-			break
+// getKeysFromPatterns does a SCAN for a key if the key contains pattern characters
+func getKeysFromPatterns(c redis.Conn, keys []dbKeyPair, count int64) (expandedKeys []dbKeyPair, err error) {
+	expandedKeys = []dbKeyPair{}
+	for _, k := range keys {
+		if regexp.MustCompile(`[\?\*\[\]\^]+`).MatchString(k.key) {
```
oliver006, line 158: I know this is existing code but can you pull `regexp.MustCompile(`[\?\*\[\]\^]+`)` out of the loop and instantiate it once as a const/var right outside this function and then just call it here?

exporter/keys.go
```diff
@@ -235,7 +202,55 @@ func parseKeyArg(keysArgString string) (keys []dbKeyPair, err error) {
 			return keys, fmt.Errorf("couldn't parse db/key string: %s", k)
 		}
 
+		// We want to guarantee at the top level that invalid values
+		// will not fall into the final Redis call.
+		if db == "" && key == "" {
```
oliver006, line 207: I think you can simplify this and just check for `empty or empty` - the addtl granularity in the error message isn't really needed.

exporter/keys.go
```diff
@@ -235,7 +202,55 @@ func parseKeyArg(keysArgString string) (keys []dbKeyPair, err error) {
...
+			log.Debugf("parseKeyArg(): Got empty database and key, pair skipped")
+			continue
+		} else if db == "" || key == "" {
+			log.Errorf("parseKeyArg(): Empty value parsed in pair '%s=%s', skip", db, key)
+			continue
+		}
+
+		number, err := strconv.Atoi(db)
+		if err != nil {
```
oliver006, line 216: Similar here, just check for `err != nil || number < 0` and return `"invalid db index", that's good enough and simplifies things a little.

exporter/keys.go
```diff
@@ -235,7 +202,55 @@ func parseKeyArg(keysArgString string) (keys []dbKeyPair, err error) {
...
 	return keys, err
 }
+
+// scanForKeys returns a list of keys matching `pattern` by using `SCAN`, which is safer for production systems than using `KEYS`.
+// This function was adapted from: https://github.com/reisinger/examples-redigo
+func scanKeys(c redis.Conn, pattern string, count int64) ([]interface{}, error) {
+	iter := 0
+	var keys []interface{}
+
+	if pattern == "" {
```
oliver006, line 234: Good you're checking the inputs, do it right away as the first thing before declaring variables.

exporter/exporter_test.go
oliver006: Yup, you can edit the make file. 
You can handle this problem in two ways.
Either, you assume that the test Redis DB is always empty and clean to start with. 
Or you can run quick count before kicking off the real test and then you know what to expect and existing keys aren't a problem.
I don't have a strong preference for either way. In the end, the Drone CI test is what matters for PRs and both options will work for that.
Also, this is useful information and should either go to the top of the file or into a "Development" section in the README so it's easier to find.

oliver006 (commented):
Great, thanks! Overall approach is fine, just a couple of minor comments / nits.

oliver006:
> `TestExtractInfoMetricsSentinel` error

Ugh, that's a fun one, good catch. Looks like the sentinel comes up before the redis6 container started and then the sentinel fails, that's annoying.
A way to fix this could be to override the start comand for the sentinel container and prefix it with `sleep 10 && ` so it gices the redis6 service enough time to start. I'm ok with not fixing that in this PR, I can do this separately.
I just restarted the drone build and it succeeded so your tests look good!

exporter/key_groups.go
houstonheat: No, missed test debugging artifacts

README.md
```diff
@@ -228,6 +228,16 @@ Most items from the INFO command are exported,
 see [Redis documentation](https://redis.io/commands/info) for details.\
 In addition, for every database there are metrics for total keys, expiring keys and the average TTL for keys in the database.\
 You can also export values of keys if they're in numeric format by using the `-check-keys` flag. The exporter will also export the size (or, depending on the data type, the length) of the key. This can be used to export the number of elements in (sorted) sets, hashes, lists, streams, etc.
+Example:
```
oliver006, line 231: Was this added on purpose? What format is that yaml in?

contrib/docker-compose-for-tests.yml
```diff
@@ -50,7 +50,7 @@ services:
   redis-sentinel:
     image: docker.io/bitnami/redis-sentinel:6.0-debian-10
     environment:
-      -  REDIS_MASTER_HOST=redis6
```
oliver006, line 53: 👍

oliver006 (commented):
Looks good!
Just one question about the k8s yaml in the README and then we should be good to go.

README.md
houstonheat: This example remains from the time this request was created. After a while, I agree it confuses rather than helps.

We successfully use `redis_exporter` in a production environment but sometimes problems with the interpretation of flags surfaced (problems from the side of a non-native speaker). I believe that the documentation in the `README.md` file does not sufficiently reflect the syntax of some application flags, while they are sufficiently described in the `--help` message, for example:
```
help: 
-count-keys string
        Comma separated list of patterns to count (eg: 'db0=production_*,db3=sessions:*'), searched for with SCAN
README:
-count-keys
 Comma separated list of patterns to count, eg: `db3=sessions:*` will count all keys with prefix `sessions:` from db `3`. db defaults to `0` if omitted.


Our real production case:
Until I opened examples and tests in the repository in mid-2019, we naively believed that this line `-check-keys db0=queue1,queue2,queueN` allows us to monitor all keys in the specified database. Yes, DB 0 is used by default, but obviously, whole this string was a misconception (should be `-check-keys db0=queue1,db0=queueN`), but there was no explicit example with multiple keys at that time.

Comment (houstonheat) [reply README.md:231]: Returning to the comment, i will remove this example from PR, its ok to me

File: README.md [hunk 13]
```diff
@@ -306,6 +306,7 @@ You can bring up the Redis test instances first by running `make docker-env-up`
 Once you're done testing you can bring down the stack by running `make docker-env-down`.\
 Or you can bring up the stack, run the tests, and then tear down the stack, all in one shot, by running `make docker-all`.
 
+***Note.** Tests initialization can lead to unexpected results when using a persistent testing environment. When `make docker-env-up` is executed once and `make docker-test` is constantly run or stopped during execution, the number of keys in the database changes, which can lead to unexpected failures of tests. Use `make docker-env-down` periodacally to clean up as a workaround.*
```
oliver006, line 309: 👍

oliver006 (approved):
🚀

oliver006:
Thanks for your patience and persistence @houstonheat - I know it took me a long time to get this reviewed and merged, thanks for your patience.

Released as [v1.22.0](https://github.com/oliver006/redis_exporter/releases/tag/v1.22.0)

PR #392, opened 2020-06-11
Add support for user auth introduced in Redis 6.0

k3a:
Users can now optionally specify username for authentication with the new ACL feature of Redis.

That also changes redigo/redis dependency to v1.8.2 which is actually newer than 2.0.x tag and supports Redis ACL logins.


k3a:
Coverage decreased probably just because adding `redis.user` flag (with `REDIS_USER`) env which are not possible to test without adding new tests for flags...

oliver006:
Thanks for the PR, this is great!

exporter_test.go
```diff
@@ -1255,7 +1255,29 @@ func TestPasswordProtectedInstance(t *testing.T) {
...
+	}()
+
+	body := downloadURL(t, ts.URL+"/metrics")
+
+	if !strings.Contains(body, "test_up") {
+		t.Errorf("error, missing test_up")
+	}
+}
+
+func TestUserPasswordProtectedInstance(t *testing.T) {
```
oliver006, line 1276: Instead of a completely new test you can also do this in a table-driven way and have the two URIs in the table, up to you, I'm fine either way.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/oliver006/redis_exporter
 go 1.13
 
 require (
-	github.com/gomodule/redigo v2.0.1-0.20200211073029-7ac8ae1ada9f+incompatible
+	github.com/gomodule/redigo v1.8.2
```
oliver006, line 6: 👍

go.sum
```diff
@@ -29,6 +29,8 @@ github.com/golang/protobuf v1.4.0-rc.2/go.mod h1:LlEzMj4AhA7rCAGe4KMBDvJI+AwstrU
 github.com/golang/protobuf v1.4.0-rc.4.0.20200313231945-b860323f09d0/go.mod h1:WU3c8KckQ9AFe+yFwt9sWVRKCVIyN9cPHBJSNnbL67w=
 github.com/golang/protobuf v1.4.0 h1:oOuy+ugB+P/kBdUnG5QaMXSIyJ1q38wWSojYCb3z5VQ=
 github.com/golang/protobuf v1.4.0/go.mod h1:jodUvKwWbYaEsadDk5Fwe5c77LiNKVO9IDvqG2KuDX0=
+github.com/gomodule/redigo v1.8.2 h1:H5XSIre1MB5NbPYFp+i1NBbb5qN1W8Y8YAQoAYbkm8k=
+github.com/gomodule/redigo v1.8.2/go.mod h1:P9dn9mFrCBvWhGE1wpxx6fgq7BAeLBk+UUUzlpkBYO0=
```
oliver006, line 33: Can you try running `go mod tidy` and see if it gets rid of the 2.0.1 module here ?

oliver006:
Released as [v1.7.0](https://github.com/oliver006/redis_exporter/releases/tag/v1.7.0)

PR #393, opened 2020-06-16
Pass RootCAs to dialTLSConfig

### LINKED ISSUE #387: Cannot connect to redis using ssl because there is no option to pass ca cert
**Describe the problem**
Cannot connect to redis using ssl because there is no option to pass ca cert

 Redis Metrics Exporter v0.31.0

**Running the exporter**
Provided these environment variables in helm chart
REDIS_ADDR = rediss://localhost:6385
REDIS_PASSWORD = <password>
REDIS_EXPORTER_TLS_CLIENT_KEY_FILE = /path/to/client.key
REDIS_EXPORTER_TLS_CLIENT_CERT_FILE = /path/to/client.cert

**Additional context**
I think redis needs CA cert as well, and I don't see a paramter to pass that in redis-exporter

Logs on redis, when redis-exporter tries to connect:
May 2020 22:39:56.416 # Error accepting a client connection: error:14094412:SSL routines:ssl3_read_bytes:sslv3 alert bad certificate

SarjakSShah:
Fixes #387

Added paramater REDIS_EXPORTER_TLS_CA_CERT_FILE --> Name of the CA certificate file (including full path) if the server requires TLS client authentication

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"flag"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -108,6 +110,16 @@ func main() {
 		tlsClientCertificates = append(tlsClientCertificates, cert)
 	}
 
+	var tlsCaCertificates *x509.CertPool
+	if *tlsCaCertFile != "" {
+		caCert, err := ioutil.ReadFile(*tlsCaCertFile)
+		if err != nil {
+			log.Fatalf("Couldn't load TLS Ca certificate, err: %s", err)
+		}
+		tlsCaCertificates := x509.NewCertPool()
```
oliver006, line 119: The variable `tlsCaCertificates` is re-declared here and shadows the variable from line 113 which means you're never successfully picking up a CA cert.
Change `:=` to `=` and you should be fine.

SarjakSShah: Fixed

oliver006: Thanks, this looks good now.

PR #394, opened 2020-06-17
Add role label to keys and cmd stats

JanBerktold:
This adds a `role` label to the metrics in the command and keyspace categories. This allows us to monitor our Redis Clusters with "Total Commands/Sec By Role" and "Total Keys By Role" charts which give insight on how much unique commands we actually process and unique data we store.

The result looks like this:
```
# HELP redis_db_keys Total number of keys by DB
# TYPE redis_db_keys gauge
redis_db_keys{db="db0",role="master"} 0
redis_db_keys{db="db1",role="master"} 0
redis_db_keys{db="db10",role="master"} 0
redis_db_keys{db="db11",role="master"} 0
redis_db_keys{db="db12",role="master"} 0
redis_db_keys{db="db13",role="master"} 0
redis_db_keys{db="db14",role="master"} 0
redis_db_keys{db="db15",role="master"} 0
redis_db_keys{db="db2",role="master"} 0
redis_db_keys{db="db3",role="master"} 0
redis_db_keys{db="db4",role="master"} 0
redis_db_keys{db="db5",role="master"} 0
redis_db_keys{db="db6",role="master"} 0
redis_db_keys{db="db7",role="master"} 0
redis_db_keys{db="db8",role="master"} 0
redis_db_keys{db="db9",role="master"} 0
# HELP redis_db_keys_expiring Total number of expiring keys by DB
# TYPE redis_db_keys_expiring gauge
redis_db_keys_expiring{db="db0",role="master"} 0
redis_db_keys_expiring{db="db1",role="master"} 0
redis_db_keys_expiring{db="db10",role="master"} 0
redis_db_keys_expiring{db="db11",role="master"} 0
redis_db_keys_expiring{db="db12",role="master"} 0
redis_db_keys_expiring{db="db13",role="master"} 0
redis_db_keys_expiring{db="db14",role="master"} 0
redis_db_keys_expiring{db="db15",role="master"} 0
redis_db_keys_expiring{db="db2",role="master"} 0
redis_db_keys_expiring{db="db3",role="master"} 0
redis_db_keys_expiring{db="db4",role="master"} 0
redis_db_keys_expiring{db="db5",role="master"} 0
redis_db_keys_expiring{db="db6",role="master"} 0
redis_db_keys_expiring{db="db7",role="master"} 0
redis_db_keys_expiring{db="db8",role="master"} 0
redis_db_keys_expiring{db="db9",role="master"} 0
```

A note on the implementation:  We currently depend on the implementation detail that the Redis `INFO` command returns the `replication` section before the `keyspace` and `commands` sections. Stock Redis will always do this but custom implementations might not. We could also split the logic to do two passes over the output, looking for instance information in the first pass and keyspace/commands in the second but that'd make the change more complex. Let me know what you think.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -357,13 +357,13 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 		txt  string
 		lbls []string
 	}{
-		"commands_duration_seconds_total":      {txt: `Total amount of time in seconds spent per command`, lbls: []string{"cmd"}},
-		"commands_total":                       {txt: `Total number of calls per command`, lbls: []string{"cmd"}},
+		"commands_duration_seconds_total":      {txt: `Total amount of time in seconds spent per command`, lbls: []string{"cmd", "role"}},
+		"commands_total":                       {txt: `Total number of calls per command`, lbls: []string{"cmd", "role"}},
 		"connected_slave_lag_seconds":          {txt: "Lag of connected slave", lbls: []string{"slave_ip", "slave_port", "slave_state"}},
 		"connected_slave_offset_bytes":         {txt: "Offset of connected slave", lbls: []string{"slave_ip", "slave_port", "slave_state"}},
 		"db_avg_ttl_seconds":                   {txt: "Avg TTL in seconds", lbls: []string{"db"}},
-		"db_keys":                              {txt: "Total number of keys by DB", lbls: []string{"db"}},
-		"db_keys_expiring":                     {txt: "Total number of expiring keys by DB", lbls: []string{"db"}},
+		"db_keys":                              {txt: "Total number of keys by DB", lbls: []string{"db", "role"}},
+		"db_keys_expiring":                     {txt: "Total number of expiring keys by DB", lbls: []string{"db", "role"}},
 		"exporter_last_scrape_error":           {txt: "The last scrape error status.", lbls: []string{"err"}},
 		"instance_info":                        {txt: "Information about the Redis instance", lbls: []string{"role", "redis_version", "redis_build_id", "redis_mode", "os"}},
 		"key_size":                             {txt: `The length or size of "key"`, lbls: []string{"db", "key"}},
@@ -459,6 +459,7 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
 	if e.redisAddr != "" {
 		startTime := time.Now()
 		var up float64 = 1
+
 		if err := e.scrapeRedisHost(ch); err != nil {
 			up = 0
 			e.registerConstMetricGauge(ch, "exporter_last_scrape_error", 1.0, fmt.Sprintf("%s", err))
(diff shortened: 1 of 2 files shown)


oliver006:
Hmm, that's an interesting case.
Adding this to the cmd metrics is a little hacky.
What about adding it via relabeling (having the Prometheus server add a label to all your read replicas if you can identify the role in the server) or you filter in the Grafana Dashboard to only aggregate over a particular role?

JanBerktold:
Yea, doing this via relabeling rules would work for relatively static setups as instance roles are predictable and don't change too often there. 
However, we run an open-source Redis Cluster inside of a container scheduler (think something like K8) where the role of a given Redis instance is not predictable and can change quite often. To give an idea, we have clusters with 500+ Redis instances with ~40 failovers happening daily. I'm not really sure how we would get the instance role at the Prometheus level, suggestions are welcome!

oliver006:
Yeah, that's definitely a limitation.
What are you trying to use the additional label for? If it's just about querying then you could look into the Grafana variable/template option I mentioned before (have a variable for master/slave role, use that to filter which hosts to show and then the host list will let you filter which metrics to aggregate over).

JanBerktold:
My current usecase is purely for visualization using Grafana. I've tried setting up a chart using Grafana variables for masters/replicas but unfortunately that displays incorrect results because the variables will answer the question "Who are the masters right now?" as opposed to "Who were the masters from when we are displaying the data points?". This causes incorrect results to be displayed as the commands/sec from Redis instances that were a master previously but are now a replica/not existent anymore are not included in the metric. 

See this for an example: The top chart is backed by another metrics store (which we're working to move away from) and the lower one is backed by PromQL using the masters-in-variable approach. The lower one is missing a good chunk of the requests/sec because the cluster topology has now changed. You can also see it jump randomly when failovers happen.

image

This leads me to believe that the most viable way to solve this problem would be to embed the instance topology info into the metrics emitted by the redis_exporter itself (because it already has all of the required information while other parts of the stack would have to essentially reimplement parts of the exporter). What do you think?

oliver006:
Ah yes, good point regarding the "what are my roles now vs what are the roles at the time".

My apologies, I did not have time to look into this any further the last couple of days but one thing came to mind. 
Did you maybe try joining the `redis_instance_info` metric (and filter for the role as it was at the time) and join it (using `on(instance)` or something like that) with the cmd metricredis_instance_infos and hence be able to get the desired results?

jon-walton:
Hi @JanBerktold 

I needed something similar. Here's an example of evicted keys on the masters. A little background on the cluster

- Cluster with 2 shards
- Each shard having a master/slave
- Shard 1 = 15000 and 15001
- Shard 2 = 15002 and 15003
- There was a failover on shard 1 (instances 15000 and 15001) at 15:00

image


and ops / sec on the masters. The dip after the failover is due to load moving over to shard 2 and not really moving back... 🤔 

image

hopefully that helps

oliver006:
Thanks for posting that, @jon-walton ! That's the query that I was thinking of, using `on(instance) group_left(role)` to join the two time series.

oliver006:
Closing this

PR #398, opened 2020-06-28
add instantaneous_ops_per_sec gauge metric

naughtyGitCat:
add instantaneous_ops_per_sec,

the irate(commands_total) seems lower than instantaneous_ops_per_sec, the instantaneous_ops_per_sec is more accurate


oliver006:
The preferred Prometheus way to get the commands/sec is to not export the redis-calculated gauges but to to export the per-command metrics and do the calculation on the Prometheus side using rate/irate.

naughtyGitCat:
yes, calc by irate(commands_total) ether than  redis-calculated is more Prometheus way, but redis-calculated gauges value is more real and accurate than irate(commands_total) when pull interval becomes longer

Issue #252, opened 2019-04-23
export streams related metrics


devlounge:
Hi,

I don't see any support for streams, is it coming soon ?

oliver006:
Support to export the length of a stream was added in #237 - what kind of support are you looking for?

PR #411, opened 2020-08-19
Streams support: Add XINFO metrics

houstonheat:
Extends #252 
Stream keys matching implemented by reusing keys parse functions. [redigo](https://github.com/gomodule/redigo) provides `ScanStruct()` for values assignment,
so each element of _groups, consumers, XINFO_ want match it's `redis:"field_name"` struct field.
Such a fields like `last-generated-id` are useless in prometheus-metrics way, i think, so they are not implemented.

More info about streams: https://redis.io/topics/streams-intro


houstonheat:
Oh, need to add some missed metrics 🌚

Issue #405, opened 2020-07-23
Add scrape sentinel info


liuyk0109:
**What version of redis_exporter are you running?**
[√ ] 0.3x.x
[ ] 1.x.x

Do we have some plans to add scrape sentinel info like:
```
[root@server10 bin]# ./redis-cli -h 192.169.2.52 -p 26379 -i 1 info Sentinel 
# Sentinel
sentinel_masters:3
sentinel_tilt:0
sentinel_running_scripts:0
sentinel_scripts_queue_length:0
sentinel_simulate_failure_flags:0
master0:name=user03,status=sdown,address=192.169.2.52:6381,slaves=1,sentinels=5
master1:name=user02,status=ok,address=192.169.2.54:6380,slaves=1,sentinels=5
master2:name=user01,status=ok,address=192.169.2.52:6379,slaves=1,sentinels=5
```
The `status` is very useful for me. Hope it can be added to the exporter.

oliver006:
Agreed, that would definitely be helpful.
I don't use Redis with sentinel and am very short of time right now but I'll gladly welcome a PR to add this.

oliver006:
Closed by #418 - thx @houstonheat !

PR #418, opened 2020-08-27
Issue #405 | Add sentinel info metrics

houstonheat:
Implements #405 
The division into three metrics (`sentinel_master_up`, `sentinel_master_sdown`, `sentinel_master_odown`) is deliberately done so that users have the opportunity to get both the master status _(up or not)_ and a specific state in case of an error: `sdown` or `odown`.

Info and descriptions from:
- https://redis.io/topics/sentinel
- https://github.com/redis/redis/blob/6.0/src/sentinel.c#L3406

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -427,8 +427,16 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 		"master_link_up":                         {txt: "Master link status on Redis slave", lbls: []string{"master_host", "master_port"}},
 		"master_sync_in_progress":                {txt: "Master sync in progress", lbls: []string{"master_host", "master_port"}},
 		"master_last_io_seconds_ago":             {txt: "Master last io seconds ago", lbls: []string{"master_host", "master_port"}},
-		"slave_repl_offset":                      {txt: "Slave replication offset", lbls: []string{"master_host", "master_port"}},
 		"script_values":                          {txt: "Values returned by the collect script", lbls: []string{"key"}},
+		"sentinel_tilt":                          {txt: "Sentinel is in TILT mode"},
+		"sentinel_masters":                       {txt: "The number of masters this sentinel is watching"},
+		"sentinel_running_scripts":               {txt: "Number of scripts in execution right now"},
+		"sentinel_scripts_queue_length":          {txt: "Queue of user scripts to execute"},
+		"sentinel_simulate_failure_flags":        {txt: "Failures simulations"},
+		"sentinel_master_status":                 {txt: "Master status on Sentinel", lbls: []string{"master_name", "master_address", "master_status"}},
+		"sentinel_master_slaves":                 {txt: "The number of slaves of the master", lbls: []string{"master_name", "master_address"}},
+		"sentinel_master_sentinels":              {txt: "The number of sentinels monitoring this master", lbls: []string{"master_name", "master_address"}},
+		"slave_repl_offset":                      {txt: "Slave replication offset", lbls: []string{"master_host", "master_port"}},
 		"slave_info":                             {txt: "Information about the Redis slave", lbls: []string{"master_host", "master_port", "read_only"}},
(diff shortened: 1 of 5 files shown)


exporter.go
```diff
@@ -692,6 +702,44 @@ func parseConnectedSlaveString(slaveName string, slaveInfo string) (offset float
...
+	if err != nil {
+		log.Debugf("parseSentinelMasterString(): couldn't parse slaves value, got: %s, err: %s", matchedMasterInfo["slaves"], err)
+		return
+	}
+	masterSentinels, err = strconv.ParseFloat(matchedMasterInfo["sentinels"], 64)
+	if err != nil {
+		log.Debugf("parseSentinelMasterString(): couldn't parse sentinels value, got: %s, err: %s", matchedMasterInfo["sentinels"], err)
+		return
+	}
+	ok = true
```
houstonheat, line 738: We can assign `ok` to `true` earlier, but I'm not sure is it really ok :)

exporter.go
```diff
@@ -1462,7 +1549,7 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 
 	infoAll, err := redis.String(doRedisCmd(c, "INFO", "ALL"))
 	if err != nil {
-		log.Debugf("Redis INFO err: %s", err)
+		log.Debugf("Redis INFO ALL err: %s", err)
```
oliver006, line 1552: :+1:  - good catch, thx

exporter.go
oliver006: I think this is fine.

houstonheat:
> ... have one metric `sentinel_master_status` that has a label `status` = `up|sdown|odown` ?

Yeah, I saw this example with `slaveState` var, but it's not a true way with Prometheus I think. [According to the documentation](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels):
_"Every time series is uniquely identified by its **metric name** and optional **key-value pairs called labels.** ... Changing any **label value**, including adding or removing a label, **will create a new time series**."_
And for the [libraries](https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels):
_"Metrics from custom collectors should almost always have consistent label names."_

So in this example, we have a consistent **metric name**, but changing its label value from `ok` to `down` will cause creating new time series, and I'm trying to avoid this scenario. Having different metrics for each state is more clear for end user, but looks a bit ugly, I agree.

oliver006:
> _"Every time series is uniquely identified by its **metric name** and optional **key-value pairs called labels.** ... Changing any **label value**, including adding or removing a label, **will create a new time series**."_

You're right, each label-metric combination in reality is a new time series so in reality, the number of timeseries are the same but your approach adds two more metrics.
We shouldn't setting unused metrics to zero (e.g. set odown and sdown to zero if a node is up). Plus you're right, the other metric uses a state label so I'd say let's use one metric `sentinel_master_status` and add a label `status` that indices up, odown or sdown.

houstonheat:
>... so I'd say let's use one metric `sentinel_master_status` and add a label `status` that indices up, odown or sdown.
Ok, couple of minutes 👌

exporter.go
```diff
@@ -810,6 +856,31 @@ func (e *Exporter) handleMetricsReplication(ch chan<- prometheus.Metric, masterH
...
+		e.registerConstMetricGauge(ch, fieldKey, float64(val))
+		return true
+	}
+
+	if masterName, masterStatus, masterAddress, masterSlaves, masterSentinels, ok := parseSentinelMasterString(fieldKey, fieldValue); ok {
+		if masterStatus == "ok" {
+			e.registerConstMetricGauge(ch, "sentinel_master_status", 1, masterName, masterAddress, masterStatus)
+		} else {
+			e.registerConstMetricGauge(ch, "sentinel_master_status", 0, masterName, masterAddress, masterStatus)
+		}
```
oliver006, line 874: 👍 

Thanks!

PR #428, opened 2020-10-03
Introduce count-keys

felixb:
Allows exposing metrics for the number of keys maching a pattern like
`session:*`.


exporter.go
```diff
@@ -421,6 +432,7 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 		"instance_info":                          {txt: "Information about the Redis instance", lbls: []string{"role", "redis_version", "redis_build_id", "redis_mode", "os", "maxmemory_policy"}},
 		"key_size":                               {txt: `The length or size of "key"`, lbls: []string{"db", "key"}},
 		"key_value":                              {txt: `The value of "key"`, lbls: []string{"db", "key"}},
+		"keys_size":                              {txt: `Count of  keys`, lbls: []string{"db", "key"}},
```
oliver006, line 435: `Count of  keys` --> `Count of keys` (remove extra space char)

oliver006: And maybe name the metric `keys_count` ?

oliver006:
Thanks for the PR.

I'm a little nervous about this as something like `KEYS session:*` has the potential for blocking a Redis server with large number of keys (see the warning on the KEYS docs page [here](https://redis.io/commands/keys))

Thoughts?

felixb:
How about implementing the same logic with SCAN instead? This leads to more interactions with redis but should not block n large DBs.

oliver006:
> How about implementing the same logic with SCAN instead? This leads to more interactions with redis but should not block on large DBs.

That's a better idea, let me re-review.

README.md
```diff
@@ -148,6 +148,7 @@ check-keys             | REDIS_EXPORTER_CHECK_KEYS            | Comma separated
 check-single-keys      | REDIS_EXPORTER_CHECK_SINGLE_KEYS     | Comma separated list of keys to export value and length/size, eg: `db3=user_count` will export key `user_count` from db `3`. db defaults to `0` if omitted.  The keys specified with this flag will be looked up directly without any glob pattern matching.  Use this option if you don't need glob pattern matching;  it is faster than `check-keys`.
 check-streams          | REDIS_EXPORTER_CHECK_STREAMS         | Comma separated list of stream-patterns to export info about streams, groups and consumers. Syntax is the same as `check-keys`.
 check-single-streams   | REDIS_EXPORTER_CHECK_SINGLE_STREAMS  | Comma separated list of streams to export info about streams, groups and consumers. The streams specified with this flag will be looked up directly without any glob pattern matching.  Use this option if you don't need glob pattern matching;  it is faster than `check-streams`.
+count-keys             | REDIS_EXPORTER_COUNT_KEYS            | Comma separated list of patterns to count, eg: `db3=sessions:*` will count all keys with prefix `sessions:` from db `3`. db defaults to `0` if omitted.
```
oliver006, line 151: Can you add a brief comment here that it uses SCAN to count the keys? so users are aware of the methodology? Maybe also a warning that using this option can still add addtl load for large key sets (all matching key values are shipped to the exporter, right?)

exporter.go
```diff
@@ -400,6 +405,12 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 		log.Debugf("singleStreams: %#v", singleStreams)
 	}
 
+	if countKeys, err := parseKeyArg(opts.CountKeys); err != nil {
+		return nil, fmt.Errorf("couldn't parse count-keys: %s", err)
+	} else {
+		log.Debugf("countKeus: %#v", countKeys)
```
oliver006, line 411: `countKeus` --> countKeys

oliver006 (commented):
Two minor issues, this is almost ready to go!

oliver006 (approved):
Looks good, thanks!

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.12.1

PR #430, opened 2020-10-12
Add redis mixin

dafydd-t:
This proposes the addition of a [monitoring mixin](https://monitoring.mixins.dev/) to the exporter. The dashboard is based on the more recent dashboard in the [contrib](https://github.com/oliver006/redis_exporter/tree/master/contrib) directory. I've also added a sample rule and some basic alert configuration.

The mixin is closely related to the exporter, so it seems natural to place the mixin within this repository. What do you think?

This initial version is very basic and does not utilise Jsonnet for the mixin config. It is intended as a starting template.

Outstanding
- [x] Template the datasource in the dashboard


oliver006:
Thanks for the PR!
I'm personally not familiar with mixins but i like the idea of adding this so other people can find it and befit from it.
Can you move the `redis-mixin` folder inside the `contrib` folder to make clear this is not part of the actual exporter (but still useful and related)?

dafydd-t:
Done! Glad you're happy to have the mixin in the repo.

PR #431, opened 2020-10-12
fix k8s deployment example with proper api version and selector added

pmmalinov01:
The example now should be simple apply, no need for local fixes.


oliver006:
Thanks for the PR!

PR #434, opened 2020-10-16
Exporter as lib

dafydd-t:
These changes define the exporter logic as a library, so that it can be used in in other applications.

Summary of changes:
- Move exporter to its own directory. I called it `lib/exporter`
- Avoid use of global vars `BuildVersion`, `BuildCommitSha`, `BuildDate`. I modified `NewRedisExporter` to take a `exporter.BuildInfo` struct to pass this information around. 

Outstanding:
- [ ] Capture recent count keys changes

### DIFF
--- a/lib/exporter/exporter.go
+++ b/lib/exporter/exporter.go
@@ -1,4 +1,4 @@
-package main
+package exporter
 
 import (
 	"crypto/tls"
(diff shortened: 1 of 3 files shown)


dafydd-t:
Closed in favour of #435 .

PR #435, opened 2020-10-16
Make exporter lib

dafydd-t:
These changes define the exporter logic as a library, so that it can be used in in other applications.

Summary of changes:
- Move exporter to its own directory. I called it `lib/exporter`
- Avoid use of global vars `BuildVersion`, `BuildCommitSha`, `BuildDate`. I modified `NewRedisExporter` to take a `exporter.BuildInfo` struct to pass this information around. 

The change of note in this PR is the change of the function signature for `NewRedisExporter`. Before this change, we couldn't use this function externally because it was in the `main` package. However, this change might be contentious so I'd like to hear your opinions on it. Alternatives or additional options are:
- Move the logic to export a build info metric into a function that you would call off the exporter. Adding to the public API, but not changing what already exists. This would remove the ability to expose build info metrics for anyone using the old API (which I think is impossible)
- Have two `NewRedisExporter` functions, each with a different signature. This would result in an addition to the API and you would not be able to export build info metrics by default.


dafydd-t:
> This PR is still in draft mode - let me know when you're ready for comments.

Thanks - I was just waiting to see if some related work I was doing required further changes here. Good to go now.

lib/exporter/exporter.go
```diff
@@ -191,14 +199,16 @@ func newMetricDescr(namespace string, metricName string, docString string, label
 }
 
 // NewRedisExporter returns a new exporter of Redis metrics.
-func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
+func NewRedisExporter(redisURI string, opts Options, bi BuildInfo) (*Exporter, error) {
```
oliver006, line 202: Instead of changing the signature - what do you think about adding the BuildInfo fields to the Options struct?

oliver006 (commented):
Change looks pretty good  - one request: can you make the package just `exporter` instead of `lib/exporter`?

lib/exporter/exporter.go
dafydd-t: Yes, that's a much better idea. I'll do that.

dafydd-t:
@oliver006 I have:
- moved lib from `lib/exporter` to `exporter` directory
- Moved `BuildInfo` into `Options`. Much cleaner!

exporter/exporter_test.go
```diff
@@ -1453,6 +1453,26 @@ func TestHTTPHTMLPages(t *testing.T) {
 	}
 }
 
+func TestRedisMetricsOnly(t *testing.T) {
```
oliver006, line 1456: 👍

oliver006 (approved):
Looks real good - nice work!

PR #439, opened 2020-10-21
Fix CGO_ENABLED flag

### LINKED ISSUE #438: Unable to use on alpine based images
**Describe the problem**
Since version 1.11.0 I'm unable do use redis_exporter on alpine based images.

```
$ ./redis_exporter 
sh: ./redis_exporter: not found
```

1.11.0 +

----

To "fix" the problem, I added the follow line on my init script:

```
ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
```

The root of the problem is that `CGO_ENABLED` are not visible to the child processes on the Makefile. The generated binary has a dynamically linked library:

```
$ file .build/redis_exporter-.linux-amd64/redis_exporter 
.build/redis_exporter-.linux-amd64/redis_exporter: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, Go BuildID=O3EaMhGscnElvmdYdgbe/2GIsgvVzSK7fuFCfsWkv/KJoyUQxG6dnBnzOYhASG/KTn9SP-vh6Y0-K05gXWu, stripped
```

laubstein:
This fixes #438

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ build-binaries:
 
 	rm -rf .build | true
 
-	CGO_ENABLED=0 ; \
+	export CGO_ENABLED=0 ; \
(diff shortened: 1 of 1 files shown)


oliver006:
Thanks for the PR!

PR #443, opened 2020-10-23
update redis-mixin templated datasource to use templated variable

dafydd-t:
This minor update fixes a bug where the datasource in a template variable was hard coded instead of using a variable. This means the variable will work when the datasource is not called `prometheus`.


dafydd-t:
Screenshot 2020-10-23 at 16 06 58

oliver006:
Nice, thanks!

PR #449, opened 2020-11-02
Fix drone docker releases

rayanimesh:
Post merging of #435, drone docker releases are broken. Check https://cloud.drone.io/oliver006/redis_exporter/922/1/13 for reference. 
This commit fixes it.


oliver006:
Thanks for the quick fix!

oliver006:
Note to self: I should a dry-run docker build to the tests so this won't happen again in the future.

PR #457, opened 2020-12-01
Support for memory usage aggregation by key groups

rmak-cpi:
Hello redis_exporter maintainer,

I am hoping to upstream an enhancement via this PR that I have been using for my local environments for tracking Redis memory usage by key groups as determined through a user-defined list of (LUA) regular expressions. I understand that such capability is available to some extent through the check-keys, check-single-keys, and count-keys parameters, so I have included a somewhat detailed explanation of how the enhancement work in the README which will hopefully convince the redis_exporter community that this is a useful enhancement to be incorporated into the official redis_exporter.

I have certainly enjoyed using redis_exporter, and I hope that I can give back a bit to the community through this PR.

### DIFF
--- a/exporter/key_groups.go
+++ b/exporter/key_groups.go
@@ -1,0 +1,236 @@
+package exporter
+
+import (
+	"encoding/csv"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/gomodule/redigo/redis"
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+)
+
+type keyGroupMetrics struct {
+	keyGroup    string
+	count       int64
+	memoryUsage int64
+}
+
+type overflowedKeyGroupMetrics struct {
+	topMemoryUsageKeyGroups   []*keyGroupMetrics
+	overflowKeyGroupAggregate keyGroupMetrics
+	keyGroupsCount            int64
+}
+
+type keyGroupsScrapeResult struct {
+	duration          time.Duration
+	metrics           []map[string]*keyGroupMetrics
+	overflowedMetrics []*overflowedKeyGroupMetrics
+}
+
+func (e *Exporter) extractKeyGroupMetrics(ch chan<- prometheus.Metric, c redis.Conn, dbCount int) {
+	allDbKeyGroupMetrics := e.gatherKeyGroupsMetricsForAllDatabases(c, dbCount)
+	if allDbKeyGroupMetrics == nil {
+		return
+	}
+	for db, dbKeyGroupMetrics := range allDbKeyGroupMetrics.metrics {
+		dbLabel := fmt.Sprintf("db%d", db)
+		registerKeyGroupMetrics := func(metrics *keyGroupMetrics) {
+			e.registerConstMetricGauge(
+				ch,
+				"key_group_count",
+				float64(metrics.count),
+				dbLabel,
+				metrics.keyGroup,
+			)
+			e.registerConstMetricGauge(
+				ch,
+				"key_group_memory_usage_bytes",
+				float64(metrics.memoryUsage),
+				dbLabel,
+				metrics.keyGroup,
+			)
+		}
+		if allDbKeyGroupMetrics.overflowedMetrics[db] != nil {
+			overflowedMetrics := allDbKeyGroupMetrics.overflowedMetrics[db]
+			for _, metrics := range overflowedMetrics.topMemoryUsageKeyGroups {
+				registerKeyGroupMetrics(metrics)
+			}
+			registerKeyGroupMetrics(&overflowedMetrics.overflowKeyGroupAggregate)
+			e.registerConstMetricGauge(ch, "number_of_distinct_key_groups", float64(overflowedMetrics.keyGroupsCount), dbLabel)
+		} else if dbKeyGroupMetrics != nil {
+			for _, metrics := range dbKeyGroupMetrics {
+				registerKeyGroupMetrics(metrics)
+			}
+			e.registerConstMetricGauge(ch, "number_of_distinct_key_groups", float64(len(dbKeyGroupMetrics)), dbLabel)
+		}
+	}
+	e.registerConstMetricGauge(ch, "last_key_groups_scrape_duration_milliseconds", float64(allDbKeyGroupMetrics.duration.Milliseconds()))
+}
+
+func (e *Exporter) gatherKeyGroupsMetricsForAllDatabases(c redis.Conn, dbCount int) *keyGroupsScrapeResult {
+	start := time.Now()
+	allMetrics := &keyGroupsScrapeResult{
+		metrics:           make([]map[string]*keyGroupMetrics, dbCount, dbCount),
+		overflowedMetrics: make([]*overflowedKeyGroupMetrics, dbCount, dbCount),
+	}
+	defer func() {
+		allMetrics.duration = time.Since(start)
+	}()
+	if strings.TrimSpace(e.options.CheckKeyGroups) == "" {
+		return allMetrics
+	}
+	keyGroups, err := csv.NewReader(
+		strings.NewReader(e.options.CheckKeyGroups),
+	).Read()
+	if err != nil {
+		log.Errorf("Failed to parse key groups as csv: %s", err)
+		return allMetrics
+	}
+	for i, v := range keyGroups {
+		keyGroups[i] = strings.TrimSpace(v)
+	}
+
+	keyGroupsNoEmptyStrings := make([]string, 0)
+	for _, v := range keyGroups {
+		if len(v) > 0 {
+			keyGroupsNoEmptyStrings = append(keyGroupsNoEmptyStrings, v)
+		}
+	}
+	if len(keyGroupsNoEmptyStrings) == 0 {
+		return allMetrics
+	}
+	for db := 0; db < dbCount; db++ {
+		if _, err := doRedisCmd(c, "SELECT", db); err != nil {
+			log.Errorf("Couldn't select database %d when getting key info.", db)
+			continue
+		}
+		allGroups, err := gatherKeyGroupMetrics(c, e.options.CheckKeyGroupsBatchSize, keyGroupsNoEmptyStrings)
+		if err != nil {
+			log.Error(err)
+			continue
+		}
+		allMetrics.metrics[db] = allGroups
+		if int64(len(allGroups)) > e.options.MaxDistinctKeyGroups {
+			metricsSlice := make([]*keyGroupMetrics, 0, len(allGroups))
+			for _, v := range allGroups {
+				metricsSlice = append(metricsSlice, v)
+			}
+			sort.Slice(metricsSlice, func(i, j int) bool {
+				if metricsSlice[i].memoryUsage == metricsSlice[j].memoryUsage {
+					if metricsSlice[i].count == metricsSlice[j].count {
+						return metricsSlice[i].keyGroup < metricsSlice[j].keyGroup
+					}
+					return metricsSlice[i].count < metricsSlice[j].count
+				}
+				return metricsSlice[i].memoryUsage > metricsSlice[j].memoryUsage
+			})
+			var overflowedCount, overflowedMemoryUsage int64
+			for _, v := range metricsSlice[e.options.MaxDistinctKeyGroups:] {
+				overflowedCount += v.count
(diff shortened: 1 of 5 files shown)


oliver006:
Thanks for the PR!
I'll try to get to reviewing it this week or next.

oliver006:
Sorry for taking so long with getting back to you.

This is very intersting and I'm sure it's really helpful information you're being able to extract from this.
My biggest concern is adding such a rather specialized piece of code to the exporter as I think your objectives can (mostly?) be achieved using existing functionality like check keys and Lua scripting.

I like the idea of tracking memory instead of just "size" (as you can with check keys) - but I wonder if size isn't "good enough" to approximate what keys use the most space in your Redis instance.

Similarly, I like the idea of not having to ship all keys to the redis_exporter when running queries but I wonder if using a existing Lua script functionality allows for exactly that.
Your concern about long-running queries is valid but IMO it's still an issue if you block for e.g. 5 seconds, even if you do it only once an hour.

This got me thinking though that it's currently possible to only run one Lua script which would be an issue if extending the usage of scripts.
It might be helpful to expand that to more scripts (or pass a directory with all the scripts) to allow for more powerful usage of Lua scripts but not sure, no one has asked for that yet 🤷‍♀️ .

Let me know what you think.

rmak-cpi:
@oliver006, thank you for taking the time to come up with the feedbacks, let me try my best to address your concerns.

> I like the idea of tracking memory instead of just "size" (as you can with check keys) - but I wonder if size isn't "good enough" to approximate what keys use the most space in your Redis instance.

I think size can certainly provide a good enough estimate of space usage, but 'MEMORY USAGE' does seem to include all the overheads (internal fragmentation e.g.) for storing both key and value which seems to be better while not exactly more expensive to process on the Redis server side(?)

> Similarly, I like the idea of not having to ship all keys to the redis_exporter when running queries but I wonder if using a existing Lua script functionality allows for exactly that.
> Your concern about long-running queries is valid but IMO it's still an issue if you block for e.g. 5 seconds, even if you do it only once an hour.
> 
> This got me thinking though that it's currently possible to only run one Lua script which would be an issue if extending the usage of scripts.
> It might be helpful to expand that to more scripts (or pass a directory with all the scripts) to allow for more powerful usage of Lua scripts but not sure, no one has asked for that yet 🤷‍♀️ .
> 

I had a feeling that the use of a purpose made LUA script is going to be a little difficult to explain, but let me try to do it in a slightly different manner: Execution(s) of the LUA script (batch-level aggregation) are actually interleaved with the client-side (redis_exporter) db-level aggregation logic. So while it may take a a full 5s to execute a full SCAN of the entire key space in my example, the redis processing thread is not "blocked" for the entire 5s. Instead, the processing thread will only be blocked for how long it takes to execute individual batches through the LUA script (< 40ms in my case according to redis exporter), and the batch size can be tuned to achieve stricter requirements. As you can probably imagine, there is inherent trade-offs between the number of network round trips and batch size/latency bubble in Redis processing thread. As far as I can tell, this seems to follow the same logic behind the use of a SCAN cursor for the existing check-keys / count-keys functionality. 

Unfortunately, the interleaving of server-side LUA script and client-side aggregation logic precludes the direct use of the existing LUA script functionality in redis_exporter. On one hand, having a single LUA script execution aggregating an entire database will likely introduce unacceptable stalls in the Redis process thread. On the other hand, a pure client-side implementation that interrogates memory usage info on a key-by-key basis will likely lead to excessive network traffic (which could be batched with redis pipeline). Overall, I do feel that the middle-ground batch-oriented approach implemented in this PR provides performance characteristics much more compelling than those two extremes. 

It may be worthwhile to generalize this "batch-oriented LUA script execution" approach to support other scenarios, but I am having a bit of trouble imagining what other scenarios are out there similar enough to what I have implemented here to make a good crack at making the appropriate generalizations.

And to go out on a limb for a bit, I do feel that the flexibility of LUA regex (with capture groups) holds substantial advantage over Redis glob patterns used by the existing check-keys/count-keys functionality. With LUA regex, you can classify all keys that follow a ":"-separated namespace hierarchy with the last component being the unique id using a single `^(.*):[^:]+$` regex into separate prefix buckets. However (I could be wrong here), someone needs to spell out all possible prefixes as glob patterns when using check-/count-keys for the same scenario and each glob pattern corresponds to a full SCAN of the Redis keyspace (only the matching keys will be returned, but Redis server would have to match every key to the pattern). That said, evaluating LUA regex could indeed be a lot more expensive than evaluating glob patterns.

I do feel that some of these points are a little on the subtle side, but I hope I made a good case explaining myself. That said, please feel free to post follow-up request for further discussions/clarifications. 

> Let me know what you think.

oliver006:
You made some good points and I think I got a better understanding of the motivations behind the PR and I'm ok with moving this forward.
I'll give this a proper code review in the coming days but I'd like to simplify this a little if possible.

In particular, do you think you could get rid of the async aggregation in a goroutine? I think this could be handled via Prometheus scrape interval and, if needed, let people run a second exporter if there's something they want to scrape just once a minute or once an hour.

And secondly, could you move your new code into a separate file? (this made me realize that exporter.go has become a bit too big and I want to start splitting it up into several files and we could start right here)

rmak-cpi:
> You made some good points and I think I got a better understanding of the motivations behind the PR and I'm ok with moving this forward.
> I'll give this a proper code review in the coming days but I'd like to simplify this a little if possible.
> 
> In particular, do you think you could get rid of the async aggregation in a goroutine? I think this could be handled via Prometheus scrape interval and, if needed, let people run a second exporter if there's something they want to scrape just once a minute or once an hour.

While the use of a gorountine for asynchronous scraping of the key groups metrics is in direct violation of the recommendation at https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling, I do want to try to make the case that the asynchronous adaptive scheduling of key group metrics scraping does free me from having to think about tailoring the prometheus scrape intervals to different redis instances with widely different sizes; and there is no on-going management of the scrape intervals as the redis instances grow or shrink in the future. 

Without adaptive asynchronous scheduling, decoupling the key groups metrics scraping schedule from the schedule of other redis metrics for larger redis instances would likely mean additional administrative overhead of configuring multiple instances of redis exporter with the potential of duplicating the common redis metrics. We can separate out the key group metrics into a completely separate endpoint or create a separate exporter altogether, but the additional administrative overhead remains.

We could also implement "adaptive" caching of key groups metrics scrape results (as per suggestion from https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling) by tracking the next key group metrics scrape timestamp in the exporter struct and a synchronous prometheus scrape will only perform a key group metrics scrape if the current timestamp exceeds the precomputed scheduled time, and the next scheduled timestamp can be computed from the duration of the current key group metrics scrape. This approach does eliminate the use of a goroutine and the additional administrative overhead that I mentioned, but the intervals between prometheus scrapes are no longer consistent which can cause gaps in the time series data, and a key groups metrics scrape can be long enough to require tuning of the prometheus scrape timeout.

Having said all that, I have removed the adaptive asynchronous key groups metrics scraping logic from this PR due to its controversial nature, but I am actually keeping a local branch with the functionality since I am actively using that in my environment. I am more than happy to create a different PR for that in the future should folks deem it necessary. 

> 
> And secondly, could you move your new code into a separate file? (this made me realize that exporter.go has become a bit too big and I want to start splitting it up into several files and we could start right here)

Breaking my code into separate code modules sounds good, and I have done it in this PR albeit in a somewhat naive manner.

oliver006:
> Breaking my code into separate code modules sounds good, and I have done it in this PR albeit in a somewhat naive manner.

Thank you. I'll try to get to a more detailed code review in the next couple of days when I can find a block of free time.

oliver006:
PS: thanks for the thorough explanation of the benefits adaptive asynchronous scheduling of the key group metrics, appreciate it.
If there's enough interest in the future to add it then we can revisit this issue.

exporter/key_groups.go
```diff
@@ -0,0 +1,240 @@
...
+		if int64(len(allGroups)) > e.options.MaxDistinctKeyGroups {
+			metricsSlice := make([]*keyGroupMetrics, 0, len(allGroups))
+			for _, v := range allGroups {
+				metricsSlice = append(metricsSlice, v)
+			}
+			sort.Slice(metricsSlice, func(i, j int) bool {
+				if metricsSlice[i].memoryUsage == metricsSlice[j].memoryUsage {
+					if metricsSlice[i].count == metricsSlice[j].count {
+						return metricsSlice[i].keyGroup < metricsSlice[j].keyGroup
+					} else {
```
oliver006, line 127: no need for an else

exporter/key_groups.go
```diff
@@ -0,0 +1,240 @@
...
+						return metricsSlice[i].count < metricsSlice[j].count
+					}
+				} else {
+					return metricsSlice[i].memoryUsage > metricsSlice[j].memoryUsage
```
oliver006, line 131: Same

exporter/key_groups.go
```diff
@@ -0,0 +1,240 @@
...
+				)
+			}
+			if allDbKeyGroupMetrics.overflowedMetrics[db] != nil {
+				overflowedMetrics := allDbKeyGroupMetrics.overflowedMetrics[db]
+				for _, metrics := range overflowedMetrics.topMemoryUsageKeyGroups {
+					registerKeyGroupMetrics(metrics)
+				}
+				registerKeyGroupMetrics(&overflowedMetrics.overflowKeyGroupAggregate)
+				e.registerConstMetricGauge(ch, "number_of_distinct_key_groups", float64(overflowedMetrics.keyGroupsCount), dbLabel)
+			} else if dbKeyGroupMetrics != nil {
```
oliver006, line 61: Not sure I follow - why the if/else ? can't you have both, overflow and dbKeyGroupMetrics, for a db ?

exporter/key_groups.go
```diff
@@ -0,0 +1,240 @@
...
+
+type keyGroupsScrapeResult struct {
+	duration          time.Duration
+	metrics           []map[string]*keyGroupMetrics
+	overflowedMetrics []*overflowedKeyGroupMetrics
+}
+
+func (e *Exporter) extractKeyGroupMetrics(ch chan<- prometheus.Metric, c redis.Conn, dbCount int) {
+	allDbKeyGroupMetrics := e.gatherKeyGroupsMetricsForAllDatabases(c, dbCount)
+	if allDbKeyGroupMetrics != nil {
```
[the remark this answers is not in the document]
rmak-cpi: Sounds good, done.

exporter/key_groups.go
```diff
@@ -0,0 +1,240 @@
...
+		if len(v) > 0 {
+			keyGroupsNoEmptyStrings = append(keyGroupsNoEmptyStrings, v)
+		}
+	}
+	if len(keyGroupsNoEmptyStrings) == 0 {
+		return allMetrics
+	}
+	for db := 0; db < dbCount; db++ {
+		if _, err := doRedisCmd(c, "SELECT", db); err != nil {
+			log.Debugf("Couldn't select database %d when getting key info.", db)
```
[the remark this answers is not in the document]
rmak-cpi: Sounds good, done.

exporter/key_groups.go
rmak-cpi: Fixed.

exporter/key_groups.go
rmak-cpi: Fixed.

exporter/key_groups.go
rmak-cpi: Fixed

exporter/key_groups.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	keyGroups, err := csv.NewReader(
+		strings.NewReader(e.options.CheckKeyGroups),
+	).Read()
+	if err != nil {
+		log.Errorf("Failed to parse key groups as csv: %s", err)
+		return allMetrics
+	}
+
+	strings.Split(e.options.CheckKeyGroups, ",")
```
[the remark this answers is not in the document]
rmak-cpi: Thanks for the catch, it was probably there for debugging some issue and I forgot to take it out afterwards. Removed.

exporter/key_groups.go
oliver006: I'm fine with the check, i'm just wondering if the `else` is needed.

exporter/key_groups.go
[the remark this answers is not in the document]
oliver006: 👍

PR #465, opened 2020-12-16
Add scrape password file support

Ansen:
password file format:
```json
{
"redis://<<HOSTNAME>>:<<PORT>>":"<<PASSWORD>>"
}
```


oliver006:
Thanks for the PR!
I'm super busy this week but I'll try to get to reviewing it as soon as possible.

exporter/redisPwd.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+
+	log "github.com/sirupsen/logrus"
+)
+
+var RedisPwd map[string]string
+
+func LoadPwdFile(pwdFile string) error {
```
oliver006, line 12: This shouldn't set a global variable but return the passwords read from the file.

You could declare a type called PasswordMap and then return this here along with the error.

exporter/redisPwd.go
```diff
@@ -0,0 +1,25 @@
...
+	log.Debugf("start load password file: %s", pwdFile)
+	bytes, err := ioutil.ReadFile(pwdFile)
+	if err != nil {
+		log.Debugf("load password file failed: %s", err)
+		return err
+	}
+	err = json.Unmarshal(bytes, &RedisPwd)
+	if err != nil {
+		log.Debugf("password file format error: %s", err)
```
oliver006, line 21: This should probably be log.Errorf() so users are aware if there json file is malformed.

pwd.json
```diff
@@ -0,0 +1,5 @@
+{
```
oliver006, line 1: Can you put this file into contrib and please call it "sample-pwd-file.json" ?

exporter/redisPwd.go
```diff
@@ -0,0 +1,25 @@
+package exporter
```
oliver006, line 1: Thanks for putting this in a separate file!
Can you rename it to `pwd_file.go` please?

exporter/redisPwd_test.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func TestLoadPwdFile(*testing.T) {
+	confFile := "../pwd.json"
+	err := LoadPwdFile(confFile)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	fmt.Println(len(RedisPwd))
```
oliver006, line 15: Instead of the Println() - can you please try checking the actual values of the map that's returned from loadPwdFile() and fail if they're not present?

main.go
```diff
@@ -48,6 +48,7 @@ func main() {
 		redisAddr           = flag.String("redis.addr", getEnv("REDIS_ADDR", "redis://localhost:6379"), "Address of the Redis instance to scrape")
 		redisUser           = flag.String("redis.user", getEnv("REDIS_USER", ""), "User name to use for authentication (Redis ACL for Redis 6.0 and newer)")
 		redisPwd            = flag.String("redis.password", getEnv("REDIS_PASSWORD", ""), "Password of the Redis instance to scrape")
+		redisPwdFile        = flag.String("redis.password-file", getEnv("REDIS_PASSWORD_FILE", ""), "Password file of the Redis instance to scrape")
```
oliver006, line 51: This also needs to be added to the README.

oliver006 (commented):
This is great, thank you for adding this functionality.

Main issue is that this shouldn't use a global variable but make the redis password map a field in the Exporter struct and populate and use it from there. I can give you further guidance with that change if you need it, just let me know.

README.md
```diff
@@ -136,6 +136,7 @@ Name                   | Environment Variable Name            | Description
 redis.addr             | REDIS_ADDR                           | Address of the Redis instance, defaults to `redis://localhost:6379`.
 redis.user             | REDIS_USER                           | User name to use for authentication (Redis ACL for Redis 6.0 and newer).
 redis.password         | REDIS_PASSWORD                       | Password of the Redis instance, defaults to `""` (no password).
+redis.password-file         | REDIS_PASSWORD_FILE                       | Password file of the Redis instance to scrape, defaults to `""` (no password file)，only takes effect when `redis.password == ""` .
```
oliver006, line 139: Can you fix the formatting here please?

exporter/exporter.go
```diff
@@ -1557,6 +1558,11 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
 		uri = "redis://" + uri
 	}
 
+	if len(e.options.PasswordMap) > 0 {
+		log.Debugf("%s pwd : ", e.redisAddr, e.options.PasswordMap[uri])
```
oliver006, line 1562: See the security warning, I think it's better to not log the full password map.

exporter/pwd_file_test.go
```diff
@@ -0,0 +1,16 @@
+package exporter
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestLoadPwdFile(t *testing.T) {
```
oliver006, line 8: The test still needs a little polishing. See what the other tests do in terms of using t.Fatalf() if expected data wasn't found in the result of `LoadPwdFile()`

exporter/pwd_file.go
```diff
@@ -0,0 +1,28 @@
...
+
+type PasswordMap struct {
+	RedisPwd map[string]string
+}
+
+func NewPasswordMap() *PasswordMap {
+	return &PasswordMap{}
+}
+
+func (p *PasswordMap) LoadPwdFile(pwdFile string) {
```
oliver006, line 18: Thinking about this more, it's probably a bit more clear flow if this returns as error and the caller logs the error and calls log.Fatalf()

README.md
oliver006: The README could also use a short description on how to use the file, maybe further down in the "Authenticating..." section.

exporter/exporter.go
```diff
@@ -1557,6 +1558,11 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
...
+		options = append(options, redis.DialPassword(e.options.PasswordMap[uri]))
```
oliver006, line 1563: Should this only be done when the uri is found? Does it work when using an empty password?

sample-pwd-file.json
```diff
@@ -0,0 +1,5 @@
+{
```
oliver006, line 1: Move into `contrib/` please

oliver006 (commented):
This is starting to look good!

I left a couple of comments. Please also add a test to exporter_test.go to test that the scrpae endpoint loads and works with the redis uri file, e.g. set the PasswordMap for os var `TEST_PWD_REDIS_URI` and make a successful http request to /scrape and confirm you get good data.

You can use `TestHTTPScrapeMetricsEndpoints` as a blueprint but I think you can simplify it a lot.

exporter/exporter.go
Ansen: Thank you for your suggestion, I have modified the code.

It works when using an empty password.

exporter/pwd_file.go
```diff
@@ -0,0 +1,31 @@
...
+	"io/ioutil"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type PasswordMap struct {
+	RedisPwd map[string]string
+}
+
+func NewPasswordMap() *PasswordMap {
```
oliver006, line 14: No need for this extra function, instead the caller can simply do `&PasswordMap{}`

exporter/pwd_file.go
```diff
@@ -0,0 +1,31 @@
...
+	return &PasswordMap{}
+}
+
+func (p *PasswordMap) LoadPwdFile(pwdFile string) error {
```
oliver006, line 18: I think it made more sense what you had before. Don't think there's a need to make `LoadPwdFile` a member function of PasswordMap, just make it return `(*PasswordMap, error)`.

exporter/exporter_test.go
```diff
@@ -2244,3 +2244,112 @@ func TestProcessSentinelSlaves(t *testing.T) {
 		})
 	}
 }
+
+func TestHTTPScrapeWithPasswordFile(t *testing.T) {
```
oliver006, line 2248: This is a great test, thanks for adding that.

Two comments:
1) you can shorten the list of expected values for success to maybe just `test_up 1` and `uptime_in_seconds`, that should be enough. You can also remove all the csk and cntk stuff, 

2) can you add a test that missing the password from the password file fails? 
There's a second password protected Redis instance in the test setup, you can try to connect to that and expect the exporter to return `test_up 0`.

exporter/pwd_file_test.go
```diff
@@ -0,0 +1,19 @@
...
+	confFile := "../sample-pwd-file.json"
+	p := NewPasswordMap()
+	err := p.LoadPwdFile(confFile)
+	if err != nil {
+		t.Fatalf("Test Failed, error: %v", err)
+	}
+
+	fmt.Println(len(p.RedisPwd))
```
oliver006, line 16: This still needs improvement.

exporter/exporter_test.go
```diff
@@ -2244,3 +2244,94 @@ func TestProcessSentinelSlaves(t *testing.T) {
...
+			Namespace: "test",
+			Password:  pwd,
+			LuaScript: []byte(`return {"a", "11", "b", "12", "c", "13"}`),
+			Registry:  prometheus.NewRegistry(),
+		}
+		httpScrapeWithPasswordRun(t, tstName, target, options, wants, tstWantStatusCode)
+	}
+
+	tstName = "scrape-pwd-file-error-password"
+	wants = []string{
```
oliver006, line 2323: Instead of redeclaring `wants` here, just make it part of a tests array and then iterate over it.
See [here](https://github.com/oliver006/redis_exporter/blob/master/exporter/exporter_test.go#L1544-L1563) for an example.
This also means you won't need `httpScrapeWithPasswordRun`.

exporter/pwd_file_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func TestLoadPwdFile(t *testing.T) {
+	confFile := "../contrib/sample-pwd-file.json"
+	passwordMap, err := LoadPwdFile(confFile)
+	if err != nil {
+		t.Fatalf("Test Failed, error: %v", err)
+	}
+
+	redisHostList := []string{
```
oliver006, line 15: Similar to my other comment, see [here](https://github.com/oliver006/redis_exporter/blob/master/exporter/exporter_test.go#L1544-L1563) what a test layout looks like.
Define a struct with some inputs and expected outputs, then iterate over it and fail if outputs don't match the expected results.

main.go
```diff
@@ -140,11 +141,20 @@ func main() {
 		registry = prometheus.DefaultRegisterer.(*prometheus.Registry)
 	}
 
+	passwordMap := make(map[string]string)
+	if *redisPwd == "" && *redisPwdFile != "" {
+		passwordMap, err = exporter.LoadPwdFile(*redisPwdFile)
+		if err != nil {
+			log.Fatalf("Use empty redis password map.")
```
oliver006, line 148: You could improve this error messgae by saying something like:

`log.Fatalf("Error loading redis passwords from file %s, err: %s", *redisPwdFile, err)`

exporter/pwd_file_test.go
Ansen: I think here only needs to test whether the contents of the password file can be read correctly.

exporter/exporter_test.go
```diff
@@ -2244,3 +2244,80 @@ func TestProcessSentinelSlaves(t *testing.T) {
...
+		}{
+			{name: "scrape-pwd-file", password: pwd, wants: []string{
+				"uptime_in_seconds",
+				"test_up 1",
+			}},
+			{name: "scrape-pwd-file-error-password", password: "error_password", wants: []string{
+				"test_up 0",
+			}},
+		} {
+			options := Options{
```
oliver006, line 2276: move this into the block down below, e.g. after line 2282

exporter/exporter.go
```diff
@@ -1563,6 +1564,10 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
 		uri = "redis://" + uri
 	}
 
+	if e.options.PasswordMap[uri] != "" {
+		options = append(options, redis.DialPassword(e.options.PasswordMap[uri]))
```
oliver006, line 1568: This is not being hit by a test:
https://coveralls.io/builds/35963780/source?filename=exporter%2Fexporter.go#L1568

Can you make sure this is really being triggered by the test below?

exporter/pwd_file.go
```diff
@@ -0,0 +1,26 @@
+package exporter
+
+import (
+	"encoding/json"
+	"io/ioutil"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// LoadPwdFile Read the redis password file and return the password map
```
oliver006, line 10: -=> `LoadPwdFile reads the redis password file and returns the password map`

exporter/pwd_file_test.go
```diff
@@ -0,0 +1,29 @@
...
+	passwordMap, err := LoadPwdFile(os.Getenv("TEST_REDIS_PWD_FILE"))
+	if err != nil {
+		t.Fatalf("Test Failed, error: %v", err)
+	}
+
+	if len(passwordMap) == 0 {
+		t.Skipf("Password map is empty -skipping")
+	}
+
+	for host, password := range passwordMap {
```
oliver006, line 22: Don't iterate over the map that you load but iterate of a given set of tests where you define the inputs and outputs.

See here for more information and examples: https://github.com/golang/go/wiki/TableDrivenTests

exporter/exporter_test.go
```diff
@@ -2244,3 +2244,80 @@ func TestProcessSentinelSlaves(t *testing.T) {
...
+
+	passwordMap, err := LoadPwdFile(os.Getenv("TEST_REDIS_PWD_FILE"))
+	if err != nil {
+		t.Fatalf("Test Failed, error: %v", err)
+	}
+
+	if len(passwordMap) == 0 {
+		t.Skipf("Password map is empty -skipping")
+	}
+	for target, pwd := range passwordMap {
```
oliver006, line 2261: Don't iterate over the map that you load but iterate of a given set of tests where you define the inputs and outputs.
For instance, spell out each of the test cases with one `for` loop instead of using `passwordMap` and the double `for` loop`.

See here for more information and examples: https://github.com/golang/go/wiki/TableDrivenTests

exporter/exporter_test.go
```diff
@@ -2244,3 +2244,80 @@ func TestProcessSentinelSlaves(t *testing.T) {
 		})
 	}
 }
+
+func TestHTTPScrapeWithPasswordFile(t *testing.T) {
+	if os.Getenv("TEST_REDIS_PWD_FILE") == "" {
```
oliver006, line 2249: This env variable is not set so the test is being skipped in the CI pipeline.

exporter/exporter_test.go
oliver006: In particular, this should fail in the CI/CD runs as the IP addresses are invalid.

See [here](https://github.com/oliver006/redis_exporter/blob/master/exporter/exporter_test.go#L1073) for an example how to include target as part of the test and pull it from the env variables.

Ansen:
Thank you very much for your patience and suggestions. As a rookie, I have learned a lot from it.

oliver006:
We're getting there, thanks for your patience :-)

Biggest problem is still your use of the `TEST_REDIS_PWD_FILE` variable. There's no need for that, you can hardcode the filename in the tests.
But if you require the env variable then you need to set it in the Makefile otherwise the tests will be skipped.
Once you got that addressed you can start working on making the `TestHTTPScrapeWithPasswordFile` tests pass, they should fail. It's because you're using hardcoded IPs in the json password file.

Ansen:
> We're getting there, thanks for your patience :-)
> 
> Biggest problem is still your use of the `TEST_REDIS_PWD_FILE` variable. There's no need for that, you can hardcode the filename in the tests.
> But if you require the env variable then you need to set it in the Makefile otherwise the tests will be skipped.
> Once you got that addressed you can start working on making the `TestHTTPScrapeWithPasswordFile` tests pass, they should fail. It's because you're using hardcoded IPs in the json password file.

Yes, I found this problem yesterday, I will try to fix it next

oliver006 (approved):
👍  - this looks good, thanks for your efforts and patience!

Ansen:
> 👍 - this looks good, thanks for your efforts and patience!

Thank you for your patient guidance, this is very important to me.

oliver006:
This has been released as 1.16.0

oldthreefeng:
Is there a good installation experiment tutorial for mutli redis by password file On k8s?  @Ansen

Ansen:
> 
> 
> Is there a good installation experiment tutorial for mutli redis by password file On k8s? @Ansen

I don't have time to do it at the moment, if you finish it, please let me know.

Ansen:
@oliver006 Please add an environment variable named REDIS_PASSWORD_FILE in docker, I didn't find where to add it. 

Thx.

Docker doc: [environment-variables](https://hub.docker.com/r/oliver006/redis_exporter/#environment-variables)

oliver006:
Oh, the docker description is way outdated, let me see that I can update that, thanks for pointing that out.

PR #467, opened 2020-12-18
[400] Add redis cluster support

ashikjm:


oliver006:
This is still in Draft mode (and tests are failing - ping me when this is ready to review or if you have any questions.

ashikjm:
I have patched the code and most of the tests are passing now. Few tests are still failing. I am checking those and will submit once ready for review.

ashikjm:
@oliver006  Need help with one test case. All tests are now passing except the TestClusterSlave. The test says test_slave_info and test_master_last_io_seconds keys are missing. I don't think these are related to the changes in this PR. I see that the TestClusterSlave explicitly tests for the presence of these keys and the /metrics doesn't expose it.

oliver006:
Hmm, that's odd.
Can you try something and change this line [here](https://github.com/oliver006/redis_exporter/blob/32742238acf18b73e08729b13d9a4a142eabbf63/.drone.yml#L147) to `- build-binaries` ? 

This should give the test cluster a bit more time to fully start up and it might resolve the issue of the missing metrics.

oliver006:
In addition, can you also add `LOG_LEVEL=debug` [somewhere here](https://github.com/oliver006/redis_exporter/blob/32742238acf18b73e08729b13d9a4a142eabbf63/Makefile#L25-L37)?

oliver006:
I have a suspicion what the problem is.
If you look at the debug output of the failed test you see this:
`test_instance_info{maxmemory_policy="noeviction",os="Linux 4.15.0-50-generic x86_64",redis_build_id="c9a9e32323d0b10f",redis_mode="cluster",redis_version="6.0.9",role="master"} 1`

Note towards the end: `role="master"`.

It looks like this information is not from the replica. Could it be that the new connection library always routes the requests to a master node? 

Note to self: we should include more information about the current Redis instance like tcp_port, run_id, etc etc

ashikjm:
oh yes. The current library does a cluster.refresh which fetches all nodes of the cluster. Then it contacts any of the nodes with the query and handles the redirect error if we dont hit the right node. 

The TestClusterSlave test expects to connect to slave and get the metrics.

I think we should use the cluster library only if key specific flags are enabled. All other requests should fall back to the old library. Thoughts?

oliver006:
Ugh, that's annoying. Is there no way to direct the library to use a particular host? 
I see there's a method to Bind() to a certain host but it expects a key as the parameter (so the slot can be calculated), that's not helpful as we don't have a key.

I'd prefer to not have multiple connections established but maybe there's no way around it.

A suggestion would be to use a "normal" connection for everything but retrieving the keys.
And then for the keys you have I guess there are two ways how to handle that:
1) introduce a flag and if "is-cluster-mode" (or whatever it's called) is set then use the cluster connection for retrieving key data, otherwise use the already established connection.
**Pro**: could re-use the connection if it's not a cluster. **Con**: extra config flag needed
2) always use a cluster connection for getting key data.
**Pro**: less config, easier for the user **Con**: an extra connection attempt will be made

I think I prefer 1) - what do you think?

ashikjm:
I also think that we should go with 1.
- extra connections are never good.
- it is cleaner.
- the flag will have a default value and hence will be transparent to the existing users. I will update the commit with the change.

ashikjm:
I am not able to resolve the master. When I rebase the master, I get the the conflicts. I fix the conflicts and continue rebase, it pulls another change and shows conflict. 

> ✗ git rebase --continue
Applying: [400] Add redis cluster support
Applying: Separating connection object for cluster and standalone instances.
Using index info to reconstruct a base tree...
M	exporter/exporter.go
Falling back to patching base and 3-way merge...
Auto-merging exporter/exporter.go
CONFLICT (content): Merge conflict in exporter/exporter.go
error: Failed to merge in the changes.
Patch failed at 0002 Separating connection object for cluster and standalone instances.
hint: Use 'git am --show-current-patch' to see the failed patch
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".

When I fix that, I get 

>  git rebase --continue
Applying: Separating connection object for cluster and standalone instances.
No changes - did you forget to use 'git add'?
If there is nothing left to stage, chances are that something else
already introduced the same changes; you might want to skip this patch.
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".

Could you try the resolve conflicts from github ui to see whether that works?

ashikjm:
Closing this as lot of merge conflicts as there was a major revamp in the repo. Opened #481  for the same.

oliver006:
> Closing this as lot of merge conflicts as there was a major revamp in the repo. Opened #481 for the same.

Ah yes, my bad. I merged #474 and that's probably hard to reconcile with this PR, good call to start over.

PR #475, opened 2021-01-14
Collect length of redis lists growing up.

jango89:
Later we can create alerts or dashboards for this.
Specifically for old Redis versions < 3.1 where MEMORY commands are not available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,12 @@ Here is a list of additional metrics that will be exposed when memory usage aggr
 |redis_number_of_distinct_key_groups               |db          |Number of distinct key groups in a Redis database when the `overflow` group is fully expanded
 |redis_last_key_groups_scrape_duration_milliseconds|            |Duration of the last memory usage aggregation by key groups in milliseconds
 
+### Script to collect Redis lists and respective sizes.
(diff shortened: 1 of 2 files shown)


oliver006:
Thanks for the PR - do you have a little more context what this is about?

jango89:
**Motivation for the PR**
We use Redis 3.2 version up and running in Prod. Most of the helpful metrics which we needed to gather based on length or memory is not possible with this version. 

**How does this PR help?**
This PR has a script which helps to collects the length of Redis lists. With the help of this count, we can take the following actions.
Create alerts or dashboards for this in grafana or any other tools with these Prometheus metrics.

oliver006:
Thanks for explaining!

I think it would help to add this to the README (maybe inbetween `Memory Usage Aggregation by Key Groups` and `Development`?) with a link to the script in the contrib folder, in that way other users have an easier time finding it.

jango89:
> Thanks for explaining!
> 
> I think it would help to add this to the README (maybe inbetween `Memory Usage Aggregation by Key Groups` and `Development`?) with a link to the script in the contrib folder, in that way other users have an easier time finding it.

I have updated and thanks for the message :)

oliver006:
Thanks!

PR #476, opened 2021-01-16
add redis endpoint to connect error message

smcavallo:
by default logger does not run in debug mode.
showing the redis endpoint in the exception message at info level and above is helpful.
the connection failure reasons are pretty standard so without the stack trace the info logs can identify connect failures and frequency which can be tracked down at the target level.

Signed-off-by: S.Cavallo <smcavallo@hotmail.com>


oliver006:
This redis instance address is omitted on purpose as it might include a password and we don't want that to show up in the logs.
If you do need to see the address you can run the exporter with the debug logging enabled.

PR #477, opened 2021-01-16
go 1.15

smcavallo:
updating the version of go in go.mod to match version in Dockerfiles
Signed-off-by: S.Cavallo <smcavallo@hotmail.com>


oliver006:
I think the go.mod file describes the required Go version and it can be an older version, no need to force Go 1.15 here so I'd prefer to leave it as is (unless I'm missing something and there's a good reason to update this).

PR #481, opened 2021-02-02
[400] Adding redis cluster support.

ashikjm:
- Introducing isCluster flag
- Using redisc module to implement cluster support

All discussions happened in #467


oliver006:
I think there was a misunderstanding in #467 with regards to what I suggested.

Your change means that, if the isCluster variable is set, you will always use the cluster-object connection for all the operations. This will cause the same error with regards to e.g. making the INFO call (it'll possibly get routed to another node, see discussion in previous).
I think you to open the regular Redis connection as before, directly to the node that the exporter is querying.
When the [key info is being retrieved here](https://github.com/oliver006/redis_exporter/blob/master/exporter/keys.go#L118) that's where you need to create and use a cluster connection so the requests are routed correctly. And only do that if the flag is set, otherwise use the connection we already have.

Let me know if that makes sense.

ashikjm:
@oliver006  oh yes. That was what we agreed on. Somehow it slipped my mind. Let me make the changes and patch the PR.

ashikjm:
@oliver006  Hope you are doing good and apologies for the long wait to get back on this PR. I have patched the PR to create and use the cluster object only if the cluster flag is set and for specific key operations only. Tests are all passing in my local. Here I see a awaiting workflow and make command failing. Could you quickly check when you get time ? 

Thank you.

oliver006:
Thanks for the update! I'll try to review this in the next couple of days and leave feedback.

oliver006:
Code looks good.
Last thing that's needed is a test that runs the exporter with isCluster=true

Have a look at the other tests, you can probably use a test like `TestKeyValuesAndSizes` as the blueprint and retrofit it so it'll set isCluster to true and hit the cluster that's running in drone (use `TEST_REDIS_CLUSTER_MASTER_URI` for the redis instance address).

oliver006:
You should check what happened here but it looks like something went wrong and you pulled in a whole bunch of already merged commits.
Once that's cleaned up I'll review it. Thanks!

ashikjm:
Cleaning this was getting messy and time consuming. So picked the changes and opened a new PR #531 . Closing this PR

PR #484, opened 2021-02-05
support cmdstats with multi underscore

tanapoln:
commandstats such as `georadius_ro` which run on slave node is missing in the exporter.

This fix the logic in splitting the command parts.

### DIFF
--- a/exporter/info.go
+++ b/exporter/info.go
@@ -279,8 +279,9 @@ func (e *Exporter) handleMetricsCommandStats(ch chan<- prometheus.Metric, fieldK
 		cmdstat_get:calls=21,usec=175,usec_per_call=8.33
 		cmdstat_set:calls=61,usec=3139,usec_per_call=51.46
 		cmdstat_setex:calls=75,usec=1260,usec_per_call=16.80
+		cmdstat_georadius_ro:calls=75,usec=1260,usec_per_call=16.80
 	*/
-	splitKey := strings.Split(fieldKey, "_")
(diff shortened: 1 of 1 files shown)


oliver-amzn:
Thanks for the PR!

Looking at the code and cmdstat_ examples now, I think the parsing could/should be done a little differently.

Instead of the first split by `_` I'd suggest splitting it by `:`, that gives you `cmdstat_<< cmd >>` and the metrics.
You can use strings.TrimPrefix() to get rid of the `cmdstat_` prefix and continue with splitting the metrics by `,` as done right now. 

What do you think ?

PS: bonus points for breaking out the splitting logic into it's own function and adding unit tests but if that's too much I'm ok with at least just adding a check somewhere in the current tests that cmds with multiple `_` are parsed fine.

oliver006:
Thanks for the PR!

Looking at the code and cmdstat_ examples now, I think the parsing could/should be done a little differently.

Instead of the first split by _ I'd suggest splitting it by :, that gives you cmdstat_<< cmd >> and the metrics.
You can use strings.TrimPrefix() to get rid of the cmdstat_ prefix and continue with splitting the metrics by , as done right now.

What do you think ?

PS: bonus points for breaking out the splitting logic into it's own function and adding unit tests but if that's too much I'm ok with at least just adding a check somewhere in the current tests that cmds with multiple _ are parsed fine.

oliver006:
Fixed by #505

Issue #496, opened 2021-02-26
age label in redis_connected_clients_details spams prometheus server


h3isenbug:
**Describe the problem**
age label of redis_connected_clients_details gets incremented every second and it makes a ton of metrics.

**What version of redis_exporter are you running?**
v1.17.1-amd64

oliver006:
Ohhh, good catch. That's probably something we should replace with a timestamp instead. Something like `now() - age` so it'll stay constant and not constantly create new metrics.

If you want to submit a PR please go ahead and I can review it. Otherwise I'll try to get to it some time in the next couple of days/weeks.

h3isenbug:
Yeah, that would do the trick. I'll submit my PR in 2 days tops.

oliver006:
That'd be great, thanks!

h3isenbug:
@oliver006 PR is ready, but the test environment seems to have problems with `latency` command.

oliver006:
Closed by #534 - released as v1.25.0

PR #497, opened 2021-03-01
fix: convert age and idle fields to timestamps

h3isenbug:
Fix for https://github.com/oliver006/redis_exporter/issues/496


h3isenbug:
@oliver006

exporter/clients.go
```diff
@@ -1,12 +1,13 @@
 package exporter
 
 import (
-	"regexp"
-	"strings"
-
 	"github.com/gomodule/redigo/redis"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
+	"regexp"
```
oliver006, line 7: Can you fix the sorting again? stdlibs first, then a blank line, then 3rd party imports.

exporter/clients.go
```diff
@@ -28,15 +29,27 @@ func parseClientListString(clientInfo string) ([]string, bool) {
...
 	hostPortString := strings.Split(connectedClient["addr"], ":")
 	if len(hostPortString) != 2 {
 		return nil, false
 	}
 
 	return []string{
 		connectedClient["name"],
-		connectedClient["age"],
-		connectedClient["idle"],
+		strconv.FormatInt(createdAt, 10),
```
oliver006, line 51: Can you use `strconv.Itoa()` ? And call that in the `durationFieldToTimestamp()` helper function?

exporter/clients.go
```diff
@@ -48,6 +61,16 @@ func parseClientListString(clientInfo string) ([]string, bool) {
 
 }
 
+func durationFieldToTimestamp(field string) (int64, error) {
+	parsed, err := strconv.ParseInt(field, 10, 64)
```
oliver006, line 65: strconv.Atoi() is the more commonly used function, can you use that?

exporter/clients_test.go
```diff
@@ -1,16 +1,24 @@
...
 import (
+	"strconv"
 	"strings"
 	"testing"
 
 	"github.com/prometheus/client_golang/prometheus"
 )
 
 func TestParseClientListString(t *testing.T) {
+	createdAt6321, _ := durationFieldToTimestamp("6321")
```
oliver006, line 12: Instead of having these called one-off here can you add a new `Test...` function that loops through a bunch of test cases and directly calls and tests `durationFieldToTimestamp()` ?

exporter/exporter.go
```diff
@@ -306,7 +306,7 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 
 	e.metricDescriptions = map[string]*prometheus.Desc{}
 
-	connectedClientsLabels := []string{"name", "age", "idle", "flags", "db", "omem", "cmd", "host"}
+	connectedClientsLabels := []string{"name", "created_at", "idle_since", "flags", "db", "omem", "cmd", "host"}
```
oliver006, line 309: 👍

exporter/clients_test.go
oliver006: (That would/should also address the comments left by the bot about the lines not covered by tests)

oliver006:
@h3isenbug  - just a friendly nudge, would love to see this merged as this could significantly reduce load on a prometheus server.

h3isenbug:
@oliver006 Sorry for the delay. I'm very busy at the moment(getting married 😅).
I'll fix these as soon as I get a chance.

oliver006:
Congratulations on getting married 🥳 !
Take your time and resume this whenever you're ready - there's no rush.

oliver006:
@h3isenbug - are you still working on this?

h3isenbug:
@oliver006 I will fix these and rebase it on master today 😊

exporter/clients.go
h3isenbug: Itoa accepts int, not int64 which is required to prevent int32 overflow in the years to follow.

h3isenbug: purpose of `durationFieldToTimestamp` is to convert duration to timestamp, not to string. It makes sense to do FormatInt in the parseClientListString, because the logic of createdAt and idleSince being string is related to `parseClientListString`, not `durationFieldToTimestamp`

exporter/clients.go
h3isenbug: as described before, FormatInt and ParseInt are better fits for this purpose.

exporter/clients_test.go
h3isenbug: proper testing of `durationFieldToTimestamp` would require frozen time. I think this function is small enough to ignore it being not tested.
I improved `TestParseClientListString` as much as possible.

h3isenbug:
@oliver006 I answered all of the discussions on this pr

exporter/clients.go
oliver006: Ah right, int64 compatibility is a good call. Still, can you do the call to `strconv.FormatInt()` inside durationFieldToTimestamp() and return a string instead of int? In that way you can also get rid of the helper function in the tests.

exporter/clients_test.go
```diff
@@ -1,16 +1,28 @@
...
 import (
+	"strconv"
 	"strings"
 	"testing"
 
 	"github.com/prometheus/client_golang/prometheus"
 )
 
 func TestParseClientListString(t *testing.T) {
+	convertDurationToTimestampString := func(duration string) string {
```
oliver006, line 12: See above, If you make `durationFieldToTimestamp()` return a string then we won't need this function.

oliver006 (commented):
Good point about int64 - that makes sense.
I think you can convert the int64 result to a string and return that in `durationFieldToTimestamp()` - otherwise this is good to go!

exporter/clients_test.go
h3isenbug: returning string is not purpose of `durationFieldToTimestamp`. This would violate SRP.

oliver006: Not sure what SRP is.
But you always convert the int that's returned from `durationFieldToTimestamp()` to a string so instead of doing that in reach place I prefer you'd just have the function return a string and then the caller can use it right away.

exporter/clients_test.go
oliver006: What I mean is using table driven tests which are common in Go.
Add tests for all the corner cases to improve test coverage.

See e.g. this one here in the std lib: https://golang.org/src/time/time_test.go#L1003

and this longer explanation: https://mj-go.in/golang/table-driven-tests-in-go

exporter/clients_test.go
zuiderkwast: SRP = single-responsibility principle

(That's what I assume it means here.)

Maybe just rename durationFieldToTimestamp to durationFieldToTimestampString and make it responsible for the whole conversion? (It's still a single responsibility.)

test12345123456: @zuiderkwast That would not make it comply with SRP, but I don't have enough time to convince everyone on this matter. I'll do it as soon as I get the time to do so. If anyone had some spare time to fix this before I do, go ahead.

h3isenbug: @zuiderkwast That would not make it comply with SRP, but I don't have enough time to convince everyone on this matter. I'll do it as soon as I get the time to do so. If anyone had some spare time to fix this before I do, go ahead.

oliver006:
@h3isenbug - gentle nudge, are you still working on this? If not then please let me know and I'll try to salvage what I can from this PR. Thanks!

h3isenbug:
@oliver006 I've been very busy lately and I think it would be great if you continued stuff done on this pr. Thanks.

oliver006:
Finishing this in #534 - thanks for getting this started @h3isenbug !

PR #504, opened 2021-03-19
Run Docker image as non-root

eumel8:
I'v struggled with some issue to get the exporter running by the Helm chart. Pod Security Policy was created but enforced AppArmor profile will prefent the POD start:

```
Status:         Pending
Reason:         AppArmor
Message:        Cannot enforce AppArmor: AppArmor is not enabled on the host
```

The default PSP on my cluster doesn't allow to run root images. Normaly the is a USER defintion in the Dockerfile which is here only in the builder, but not the finally image. Added the USER definition there and got the exporter running.

### DIFF
--- a/docker/Dockerfile.amd64
+++ b/docker/Dockerfile.amd64
@@ -39,6 +39,7 @@ FROM alpine as alpine
 COPY --from=builder /redis_exporter /redis_exporter
 COPY --from=builder /etc/ssl/certs /etc/ssl/certs
 
+# Run as non-root user for secure environments
 USER 59000:59000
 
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR, this is great!
Can you add this to the two ARM docker files as well? Thanks!

eumel8:
@oliver006 done

oliver006 (approved):
Looks good, thanks!

oliver006:
Released as [v1.20.0](https://github.com/oliver006/redis_exporter/releases/tag/v1.20.0)

PR #526, opened 2021-06-07
Add support for TLS certificate hot-reload

### LINKED ISSUE #525: Support TLS certificate hot-reload
**Describe the problem**
When TLS certificates are renewed there is a need to trigger `redis_exporter` to use the new certs.
The current procedure is to restart the process.

The proposal is to enable hot-reload of the certs, either by lazy-loading the configured files during the accept or connect,
or by using something like `fsnotify` to trigger a reload when changes occur.

This would be applicable for both the metrics endpoint server and the connection to Redis.

v1.23.1 / master

bjosv:
This changes the handling of TLS cert files from pre-loading to lazy loading when `Prometheus` connects,
or when `redis_exporter` connects to a `Redis` instance.

This enables certificate rotation in systems that require increased security. A setup example would be to mount the cert-files into the `redis_exporter` container and let other systems update these files. The update process can run without involving `redis_exporter` and there is no need know when and why we need to restart the `redis_exporter` process for cert-reloading.
Previously the certs could expire in redis or redis-client and the only way to see this was that metrics was missing.

The scape interval is usually in seconds so I have not seen any significant  performance penalty for this, even with lower intervals.

Equivalent handling exists in other metrics exporters/scrapers, like [elasticsearch_exporter](https://github.com/prometheus-community/elasticsearch_exporter)

Resolves #525

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"crypto/tls"
-	"crypto/x509"
 	"flag"
 	"io/ioutil"
 	"net/http"
@@ -120,28 +119,6 @@ func main() {
 		log.Fatalf("Couldn't parse connection timeout duration, err: %s", err)
 	}
 
-	var tlsClientCertificates []tls.Certificate
-	if (*tlsClientKeyFile != "") != (*tlsClientCertFile != "") {
-		log.Fatal("TLS client key file and cert file should both be present")
-	}
-	if *tlsClientKeyFile != "" && *tlsClientCertFile != "" {
-		cert, err := tls.LoadX509KeyPair(*tlsClientCertFile, *tlsClientKeyFile)
-		if err != nil {
-			log.Fatalf("Couldn't load TLS client key pair, err: %s", err)
-		}
-		tlsClientCertificates = append(tlsClientCertificates, cert)
-	}
-
-	var tlsCaCertificates *x509.CertPool
-	if *tlsCaCertFile != "" {
-		caCert, err := ioutil.ReadFile(*tlsCaCertFile)
-		if err != nil {
-			log.Fatalf("Couldn't load TLS Ca certificate, err: %s", err)
-		}
-		tlsCaCertificates = x509.NewCertPool()
-		tlsCaCertificates.AppendCertsFromPEM(caCert)
-	}
-
 	passwordMap := make(map[string]string)
 	if *redisPwd == "" && *redisPwdFile != "" {
 		passwordMap, err = exporter.LoadPwdFile(*redisPwdFile)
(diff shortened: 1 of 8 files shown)


bjosv:
Are the sentinel tests flaky or have I just been lucky multiple times?
After checking them it looks like my changes are not in that code path... but still

oliver006:
The sentinel tests can be a little flaky, note to self: find a way to fix that.
I restarted the tests and it passed.

And thanks for opening the PR, this makes a lot more sense now. I misunderstood your issue when you opened it and thought this is about TLS for the https endpoint only but now I see this includes the certs needed for connecting to a Redis instance. This indeed is a very useful feature.

main.go
```diff
@@ -203,11 +181,30 @@ func main() {
...
+		log.Fatal(err)
+	}
+
 	log.Infof("Providing metrics at %s%s", *listenAddress, *metricPath)
 	log.Debugf("Configured redis addr: %#v", *redisAddr)
 	if *tlsServerCertFile != "" && *tlsServerKeyFile != "" {
 		log.Debugf("Bind as TLS using cert %s and key %s", *tlsServerCertFile, *tlsServerKeyFile)
-		log.Fatal(http.ListenAndServeTLS(*listenAddress, *tlsServerCertFile, *tlsServerKeyFile, exp))
+
+		// Verify that the initial key pair is accepted
```
oliver006, line 198: 👍

oliver006 (commented):
This looks really good, thanks.
Bonus points if we could add a test that successfully loads the certs, both in tls.go:31-33 and down in LoadKeyPair().
You could either add test certs in the contrib folder (and make them valid for 10yrs) or dynamically create one for the test by calling `openssl req ...`
But I'm fine without it too, just let me know what you want to do.

bjosv:
> Bonus points if we could add a test that successfully loads the certs, both in tls.go:31-33 and down in LoadKeyPair().
> You could either add test certs in the contrib folder (and make them valid for 10yrs) or dynamically create one for the test by calling `openssl req ...`

I'll give it a go, the coverage was lowered a bit to much for my own pride :)
After a quick look my plan is to add an addition Redis 6 instance to `docker-compose-for-tests.yml` that uses TLS,
and add a connection test in tls_test.go. I'll add the cert-generation script and certs to the contrib folder to start with,
then the decision to require openssl when running tests can be taken later.

bjosv:
It looks like we use drone.io to setup Redis "services" in CI, but I couldn't find a way to mount files/folders like its done in docker-composer. Do you know if it can be done, or have some other suggestions how to start the Redis instance with TLS configs in CI?

oliver006:
> Ignore above, I found out that the repo already was available in service containers..

Yeah, the repo base dir should be available to the service.

Fyi, just opened #527 to fix the flaky sentinel tests (hopefully for good). Going to merge this in a minute and then you can pull this in to avoid future issues with your PR.

bjosv:
> Fyi, just opened #527 to fix the flaky sentinel tests (hopefully for good). Going to merge this in a minute and then you can pull this in to avoid future issues with your PR.

Worked nicely, got hit by a "misbehaving Go proxy" instead but got through that.

I added a cert generation script that creates the certs during the `test` build target, works for both docker-composer and drone runs. Also improved tests for tls.go so it covers 100%.

I first had thoughts to run towards a Redis instance using TLS, but without running code from main.go I couldn't see much gain. It would be more like testing handling in redigo, and maybe it would require to have certs in the repo to avoid timing problems since the drone containers seems started in parallel..

oliver006:
Nice work, thanks for the quick turn-around!

PR #531, opened 2021-07-07
[400] Adding redis cluster support

ashikjm:
- Added tests for cluster support


ashikjm:
@oliver006  I closed the other PR and pulled the changes here. Please review.

oliver006:
Thanks for cleaning this up! I'll try to review this by the end of this week.

exporter/exporter_test.go
```diff
@@ -117,6 +118,65 @@ func setupDBKeys(t *testing.T, uri string) error {
...
+	}
+
+	c, err := redisc.RetryConn(conn, 10, 100*time.Millisecond)
+	if err != nil {
+		log.Printf("RetryConn failed: %v", err)
+	}
+
+	defer c.Close()
+
+	if _, err := c.Do("SELECT", "0"); err != nil {
```
oliver006, line 144: Instead of repeating this all here, put the setupKeys() code in a separate function and call it from both setupDBKeys() and setupDBKeysCluster() and pass the connection object as a parameter.

exporter/exporter_test.go
```diff
@@ -148,6 +208,49 @@ func deleteKeysFromDB(t *testing.T, addr string) error {
 	return nil
 }
 
+func deleteKeysFromDBCluster(t *testing.T, addr string) error {
```
oliver006, line 211: Same as setupDBKeys(), try to reduce code repetition.

exporter/redis.go
```diff
@@ -52,6 +55,70 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
...
+	} else {
+		if frags := strings.Split(uri, ":"); len(frags) != 2 {
+			uri = uri + ":6379"
+		}
+	}
+	log.Debugf("Trying DialURL(): %s", uri)
+
+	if e.options.PasswordMap[uri] != "" {
+		options = append(options, redis.DialPassword(e.options.PasswordMap[uri]))
+	}
```
oliver006, line 96: There's a lot of copy-paste from `connectToRedis()` - consider moving lines 19-43 into a separate function and reuse it here and in connecttoRedis()

moish83:
@ashikjm, great job, thank you. 
when do you think this will be merged?
Another question, how will the exposed metrics differentiate between shards? Will we get another Prometheus label for shard number or something like that?

exporter/exporter_test.go
```diff
@@ -59,21 +60,44 @@ func getTestExporterWithOptions(opt Options) *Exporter {
...
+		log.Debugf("Cluster refresh failed: %v", err)
+		return nil, err
+	}
+
+	log.Printf("Creating redis connection object")
+	conn, err := cluster.Dial()
 	if err != nil {
-		t.Errorf("couldn't setup redis for uri %s, err: %s ", uri, err)
-		return err
+		log.Printf("Dial failed: %v", err)
```
oliver006, line 78: Can you keep this is as `Errorf()` please?

exporter/exporter_test.go
```diff
@@ -112,19 +136,11 @@ func setupDBKeys(t *testing.T, uri string) error {
 	c.Do("XREADGROUP", "GROUP", "test_group_1", "test_consumer_2", "COUNT", "1", "STREAMS", TestStreamName, ">")
 	c.Do("XREADGROUP", "GROUP", "test_group_2", "test_consumer_1", "COUNT", "1", "STREAMS", TestStreamName, "0")
 
-	time.Sleep(time.Millisecond * 50)
-
 	return nil
+
```
oliver006, line 140: Remove the blank line plz.

exporter/exporter_test.go
```diff
@@ -145,6 +161,73 @@ func deleteKeysFromDB(t *testing.T, addr string) error {
 
 	c.Do("DEL", TestSetName)
 	c.Do("DEL", TestStreamName)
+
```
oliver006, line 164: No blank line plz.

exporter/exporter_test.go
```diff
@@ -145,6 +161,73 @@ func deleteKeysFromDB(t *testing.T, addr string) error {
...
+		return err
+	}
+
+	time.Sleep(time.Millisecond * 50)
+
+	return nil
+}
+
+func setupDBKeysCluster(t *testing.T, uri string) error {
+
```
oliver006, line 187: no blank line plz

exporter/exporter_test.go
```diff
@@ -59,21 +60,44 @@ func getTestExporterWithOptions(opt Options) *Exporter {
...
+func setupKeys(t *testing.T, c redis.Conn, dbNumStr string) error {
 
 	if _, err := c.Do("SELECT", dbNumStr); err != nil {
 		log.Printf("setupDBKeys() - couldn't setup redis, err: %s ", err)
 		// not failing on this one - cluster doesn't allow for SELECT so we log and ignore the error
 	}
 
 	for _, key := range keys {
-		_, err = c.Do("SET", key, TestValue)
+		_, err := c.Do("SET", key, TestValue)
```
oliver006, line 100: Good catch. 
You could even change this to `if _, err := c.Do("SET", key, TestValue); err != nil { ` to make sure the scoping is as tight as possible.

exporter/redis.go
oliver006: This still applies - I think you could consolidate a bit here.

exporter/redis.go
```diff
@@ -52,6 +64,50 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
 	return c, err
 }
 
+func (e *Exporter) connectToRedisCluster() (redis.Conn, error) {
+	uri := e.redisAddr
+	if strings.Contains(uri, "://") {
```
houstonheat, line 69: Since you have already copied this code into `TestClusterKeyValuesAndSizes()`, I propose to move the stage of determining the address into a separate method. In this case, it can also be tested and reused during unit testing, e.g. `getClusterAddr()`:
```go
// redis.go
func (e *Exporter) getClusterAddr() string {
	clusterAddr := e.redisAddr
	if strings.Contains(clusterAddr, "://") {
		...
	} else {
		...
	}

	return clusterAddr
}
```

exporter/exporter.go
```diff
@@ -559,7 +560,16 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 
 	e.extractLatencyMetrics(ch, c)
 
-	e.extractCheckKeyMetrics(ch, c)
+	if e.options.IsCluster {
+		clusterClient, err := e.connectToRedisCluster()
```
houstonheat, line 564: 1.  Don't forget to call `defer()` when creating any connection:
```go
defer clusterClient.Close()
```
2. Is it ok to use 2 separate clients/connections in the single `scareHost()`?

exporter/redis.go
```diff
@@ -52,6 +64,50 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
...
+		}
+	}
+
+	options, err := configureOptions(e, uri)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Debugf("Creating cluster object")
+	cluster := redisc.Cluster{
```
houstonheat, line 88: If the `connectToRedisCluster()` returns `c` object _(which is `redisc.RetryConn()`, should be deferred on the top level)_, how can we `defer cluster.Close()`? `cluster` will be created for each  call of  the `scrapeHost() -> connectToRedisCluster()` and I'm not sure about possible _"memory"_ or _"connection leak"_ without explicit closing.

I apologize for such non-obvious questions, but it really worries me from the side of the stability of this monitoring project, servicing the need of thousands of operations engineers.

houstonheat (changes requested):

exporter/exporter.go
oliver006: 1. that is very good catch, adding the `defer close()` after line 568 should do the trick.

2. I think that's fine.

exporter/keys.go
```diff
@@ -93,6 +93,11 @@ func (e *Exporter) extractCheckKeyMetrics(ch chan<- prometheus.Metric, c redis.C
 
 	log.Debugf("allKeys: %#v", allKeys)
 	for _, k := range allKeys {
+		if e.options.IsCluster {
+			//Cluster mode only has one db
+			k.db = "0"
```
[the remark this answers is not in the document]
oliver006: Agreed, best to just put the wrap the select statement with an `if !isCluster` check.

exporter/redis.go
oliver006: That's an interesting question.
I think you're right, you're supposed to call `cluster.Close()` when done to release all the cluster resources. And we don't make `cluster` available to the caller so it's not possible.

But

If you look at the code for Close() [here](https://github.com/mna/redisc/blob/v1.2.1/cluster.go#L397-L411) then you notice that it's only closing all the pools and I think we don't use pools as we use `Dial()`.

From [the redisc docs](https://pkg.go.dev/github.com/mna/redisc#hdr-Cluster): 
`The Dial method, on the other hand, guarantees that the returned connection will not be managed by a pool` 

I'm not loving that we rely on that but I think I'm ok with it.

exporter/exporter.go
ashikjm: I have added the defer in the function. Good catch @houstonheat . Thank you.

exporter/keys.go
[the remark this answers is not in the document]
ashikjm: i have moved this to the else section of the `if isCluster`.

ashikjm:
@oliver006  I had to rebase to push changes to this branch. It was refusing to accept the push saying the branch was behind. But rebase has pulled a lot of changes from the master branch. Do you know a easy way to clean this? Last time I had to abandon the whole commit and raise a new PR.

oliver006:
Hmm, not sure what's the easiest way to fix this. My normal flow for this is to first do a `git fetch` or `git pull` and then `git rebase -i origin/master` and then you replay your changes on top of master (and you also squash a few commits).

I thing you could just leave this as is. I'll squash merge the PR when it's done and then the extra commits will go away. I'll try to get this reviewed by the end of today.

houstonheat:
`git pull --rebase origin master` for short :)

exporter/keys_test.go
```diff
@@ -52,6 +52,51 @@ func TestKeyValuesAndSizes(t *testing.T) {
...
+			uri = url.Host + ":6379"
+		} else {
+			uri = url.Host
+		}
+	} else {
+		if frags := strings.Split(uri, ":"); len(frags) != 2 {
+			uri = uri + ":6379"
+		}
+	}
+
```
oliver006, line 74: I don't think you need all of this, you can expect a correctly formed Uri for the tests.

exporter/redis.go
```diff
@@ -28,13 +31,22 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
 		options = append(options, redis.DialPassword(e.options.Password))
 	}
 
+	if e.options.PasswordMap[uri] != "" {
```
oliver006, line 34: Don't pass `uri` if this is all you need it for, instead pass `password`. Move the password check into the caller (and check both, `e.options.password` and the map and then pass it to configureOptions).

You should use `e.redisAddr` for the password lookup from the map, in that way it's the exact same format as provided by the user and can match what they put in the password file. Makes it easier to keep those two consistent.

Also, change the function signature to `func (e *Exporter) configureOptions(uri string) ...`

exporter/redis.go
oliver006: I'd suggest to get rid of the copy in `TestClusterKeyValuesAndSizes()` - we control what uri is being passed to the tests and we shouldn't have to sanitize the input there.

exporter/redis.go
```diff
@@ -52,6 +64,50 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
...
+	}
+
+	log.Debugf("Creating cluster object")
+	cluster := redisc.Cluster{
+		StartupNodes: []string{uri},
+		DialOptions:  options,
+	}
+	log.Debugf("Running refresh on cluster object")
+	if err := cluster.Refresh(); err != nil {
+		log.Debugf("Cluster refresh failed: %v", err)
```
oliver006, line 94: Errorf

exporter/redis.go
```diff
@@ -52,6 +64,50 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
...
+	}
+
+	log.Debugf("Creating redis connection object")
+	conn, err := cluster.Dial()
+	if err != nil {
+		log.Debugf("Dial failed: %v", err)
```
oliver006, line 100: Errorf

exporter/redis.go
```diff
@@ -52,6 +64,50 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
...
+	}
+
+	c, err := redisc.RetryConn(conn, 10, 100*time.Millisecond)
+	if err != nil {
+		log.Debugf("RetryConn failed: %v", err)
```
oliver006, line 105: Errorf

exporter/exporter_test.go
```diff
@@ -59,40 +60,57 @@ func getTestExporterWithOptions(opt Options) *Exporter {
 	return e
 }
 
-func setupDBKeys(t *testing.T, uri string) error {
-	c, err := redis.DialURL(uri)
+func createClusterClient(uri string) (redis.Conn, error) {
```
oliver006, line 63: Why a repeat all the logic here and not use `connectToRedisCluster()` instead? 

All you need to do is set `redisAddr`, do something like this:

```
e := Exporter{redisAddr: uri}
c, err :=  e.connectToRedisCluster()
...
```

ashikjm:
> I see you pushed up some new commits (and also some old ones), I'll try to review this in the next couple of days.

I only rebase from the master before pushing `git pull --rebase origin master`. But then my commits get rejected saying 'Updates were rejected because the tip of your current branch is behind its remote counterpart.'. Then I have to git pull and then all these get pulled in.

oliver006:
This looks good now @ashikjm - thanks for your patience and persistence!

ashikjm:
> This looks good now @ashikjm - thanks for your patience and persistence!

Thank you for bearing with me @oliver006  :)

PR #538, opened 2021-08-20
Add condition to check if key is of type string before running GET to increase performance

kerossin:
This adds a condition so that GET is only ran on types of keys that it works on - strings.

Before it was ran on every key and you could see errors with debug mode:
running command: GET [my-key]
WRONGTYPE Operation against a key holding the wrong kind of value

This also reduces the number commands ran on Redis from about 4 to 3 per key and should improve performance (unless all of your keys are strings).

I also thought of switching places of PFCOUNT and STRLEN which are ran on string types on the premise that regular strings are more common than HyperLogLog ones and STRLEN being the first conditional could slightly improve performance.

The problem is that STRLEN also works on HLL strings but returns the wrong (or at least not the one we want) value so I left it as is and added a comment for the reasoning behind the order.

### DIFF
--- a/exporter/keys.go
+++ b/exporter/keys.go
@@ -33,6 +33,8 @@ func getKeyInfo(c redis.Conn, key string) (info keyInfo, err error) {
 	case "none":
 		return info, errKeyTypeNotFound
 	case "string":
+        // Use PFCOUNT first because STRLEN on HyperLogLog strings returns the wrong length
+        // while PFCOUNT only works on HLL strings and returns an error on regular strings.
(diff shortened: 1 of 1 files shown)


oliver006:
I fixed the CI issue, rebase your branch to pull in the changes from #539, that should  fix the issue.

oliver006:
This PR looks good but would be good to see the tests, etc succeed.

kerossin:
Hey, pulled in your change. Tests are looking good.

oliver006:
Thank you!

oliver006:
Released as v1.26.0
https://github.com/oliver006/redis_exporter/releases/tag/v1.26.0

Issue #533, opened 2021-08-02
Multiplatform Docker Images


bartenew:
**Describe the problem**
Today the images for amd64 and arm64 are in separate tags. There is a way to unite them under the same tag using docker manifest or docker buildx utility. My use case is that I have multiple Kubernetes nodes where some of them are ARM64 and some are AMD64. Having a docker manifest will allow using the same tag for all platforms

**What version of redis_exporter are you running?**
Please run `redis_exporter --version` if you're not sure what version you're running.
[ ] 0.3x.x
[x] 1.x.x

oliver006:
Thanks for raising this issue - do you have a link with more information? If there's a better way to organize/set up the Docker files then I'd be interested to see how.

garrygerber:
Would love to see this supported too!

you will find all the information you need in: https://docs.docker.com/buildx/working-with-buildx/
but basically, you need your Dockerfile to know to differentiate between the architectures and install appropriate files to each arch, and then use the following command to build and push it:
`docker buildx build --platform linux/amd64,linux/arm64 --push -t latest .`

hope this helps

oliver006:
Thanks Garry, that's helpful.
I'd need to look into how to make `buildx` work with the drone plugin and then rework the Dockerfile accordingly but I'm a little short of time right now.
If anyone is interested in picking this up and submitting a PR that'd be great!

oliver006:
Fixed by #548 - thanks @garrygerber !

riconnon:
@oliver006 I think the changes you've made are a huge step in the right direction, but I think there's still a little left to be desired.
At the moment, from what I can tell, only the `alpine` tag is a multi-platform manifest, and the other tags, (eg. `latest`, `v1.31.0` or `v1.31.0-alpine`) are still single platform

oliver006:
Thanks for double-checking our work @riconnon !

I pushed a change and released v1.31.1 and I now see this on [Docker Hub](https://hub.docker.com/r/oliver006/redis_exporter/tags):
<img width="1266" alt="image" src="https://user-images.githubusercontent.com/1222339/141215422-c1951907-e488-4dad-be0c-0d372522c98f.png">
 
Does that look correct to you?

garrygerber:
there is a small issue with the ARM tags, they are built for arm64/v8:
repo:
`docker run oliver006/redis_exporter:v1.31.1-arm64  (master)redis_exporter
[...]

PR #548, opened 2021-09-27
change to single docker file for multiple arch's

garrygerber:
Dockerfile now supports all arch's, control on the arch is made by the variable "TARGETPLATFORM"

using the following plugin [drone-docker-buildx](https://drone-plugin-index.geekdocs.de/plugins/drone-docker-buildx/) which uses the "docker buildx" command, you can set a flag "platforms" to the plugin and pass with comma-separation, list of the arch's you want
I tried to update the drone.yml but can't test it, hope this is enough to allow you to fix issue #533


oliver006:
Thanks for the PR - I'll try to get to reviewing it as soon as I can, hopefully in the next 1-2 days.

oliver006:
Sorry it took me a while to get to this.
The general approach makes sense, two things we should look into:
- something is wrong with your yaml file, you can see the error here: https://cloud.drone.io/oliver006/redis_exporter/1244 - I think line 221 is not indented right
- I'd like to avoid using a 3rd party plugin from an unknown contributor (thegeeklab/drone-docker-buildx). Let me look into that plugin and what it does differently compared to the official plugin and maybe there's an alternate way to make this work with the official plugin.

oliver006:
I think the issue is that the docker build step needs to run in privileged mode. Can you try adding an attribute `privileged: true` and see if that helps? It might not be possible to run the geeklabs image in privileged mode, in that case you could try using the regular docker image and directly invoke the build command via the `commands` parameter.

garrygerber:
this doesn't work with the plugins/docker image.

another way:
can you add to the DRONE_RUNNER_PRIVILEGED_IMAGES env var - "thegeeklab/drone-docker-buildx"
according to https://docs.drone.io/runner/docker/configuration/reference/drone-runner-privileged-images/
you can set the default containers that start with privileged mode

oliver006:
Unfortunately  I don't own or control the cloud.drone.io instance and don't have a way to set the `DRONE_RUNNER_PRIVILEGED_IMAGES` variable.

garrygerber:
using the docker_pass and docker_user helped, it passed and built the manifest!
its working and ready for merge, looking forward to your review
Thanks!

oliver006:
Thanks for figuring this out, great work!
I'll try to review as soon as possible, hopefully in the next few days.

.drone.yml
```diff
@@ -313,24 +331,51 @@ steps:
...
-      dockerfile: ./docker/Dockerfile.amd64
+      dockerfile: ./docker/Dockerfile
       target: scratch
       build_args:
         - 'TAG=${DRONE_TAG}'
         - 'SHA1=${DRONE_COMMIT_SHA}'
         - 'GOARCH=amd64'
       config:
         from_secret: docker_config_json
+      username:
```
oliver006, line 375: This shouldn't be necessary for quay, they're for docker hub only..

.drone.yml
```diff
@@ -340,18 +385,23 @@ steps:
...
+      tags: "latest-arm64,${DRONE_TAG}-arm64,alpine-arm64"
+      dockerfile: ./docker/Dockerfile
+      target: alpine
       build_args:
         - 'TAG=${DRONE_TAG}'
         - 'SHA1=${DRONE_COMMIT_SHA}'
         - 'GOARCH=arm64'
       config:
         from_secret: docker_config_json
+      username:
```
oliver006, line 401: Same

.drone.yml
```diff
@@ -361,18 +411,23 @@ steps:
...
+      tags: "latest-arm,${DRONE_TAG}-arm,alpine-arm"
+      dockerfile: ./docker/Dockerfile
+      target: alpine
       build_args:
         - 'TAG=${DRONE_TAG}'
         - 'SHA1=${DRONE_COMMIT_SHA}'
         - 'GOARCH=arm'
       config:
         from_secret: docker_config_json
+      username:
```
oliver006, line 427: Same

.drone.yml
```diff
@@ -382,27 +437,56 @@ steps:
...
+      tags: "${DRONE_TAG}-alpine,alpine-amd64"
+      dockerfile: ./docker/Dockerfile
       target: alpine
       build_args:
         - 'TAG=${DRONE_TAG}'
         - 'SHA1=${DRONE_COMMIT_SHA}'
         - 'GOARCH=amd64'
       config:
         from_secret: docker_config_json
+      username:
```
oliver006, line 453: Same

.drone.yml
```diff
@@ -382,27 +437,56 @@ steps:
...
         - tag
     depends_on:
       - tests
 
+
+  - name: manifest-quay-alpine
+    image: plugins/manifest
+    pull: if-not-exists
+    settings:
+      username:
```
oliver006, line 468: This should use the json config, not the docker username and password.

oliver006 (commented):
Left a few comments.
Also, something is trimming the end-of-file line feeds of files you touch - maybe you can revert that and turn that off.
Thanks!

garrygerber:
@oliver006 Thanks for the review, I've fixed all I think
can you review it again?

oliver006:
Thanks @garrygerber for seeing this through to the end - I think this looks good, let's hope it all works once I merge and tag a release!

garrygerber:
Thanks @oliver006 glad we made it work.
looking forward to your tag and release!

oliver006:
Had to push one minor tweak but I think it worked!
Released as v1.31.0

PR #549, opened 2021-10-01
Use $__rate_interval for rate queries in monitoring mixin

rgeyer:
This PR changes to using [$__rate_interval](https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/) for promql queries containing a `rate` function call.

The hardcoded default of `1m` only works when the tsdb has data with a scrape interval of 15s or less. Some users may extend this to 30s or 60s in an effort to reduce time series data storage costs, which would render these queries ineffective.

A second commit on this PR also reduces the default timerange from 12h to 30m. This should help improve responsiveness of the first load, and reduce strain on the tsdb when the dashboard is used regularly.

### DIFF
--- a/contrib/redis-mixin/dashboards/redis-overview.json
+++ b/contrib/redis-mixin/dashboards/redis-overview.json
@@ -89,7 +89,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "avg(irate(redis_commands_total{instance=~\"$instance\"} [1m])) by (cmd)",
+          "expr": "avg(irate(redis_commands_total{instance=~\"$instance\"} [$__rate_interval])) by (cmd)",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 2,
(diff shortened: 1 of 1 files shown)


oliver006:
Thanks for the PR, Ryan. This looks good but I'd prefer to keep the default time range a little longer, at least 23h or so, can't really see much what's going on with a 30mins range.

rgeyer:
> Thanks for the PR, Ryan. This looks good but I'd prefer to keep the default time range a little longer, at least 23h or so, can't really see much what's going on with a 30mins range. 

Fair enough, that was a bit overly opinionated. I'll revert the time range change by Monday. :)

oliver006:
I think your branch is out of date (still uses golang 1.15 images in drone) - can you please rebase?

rgeyer:
> I think your branch is out of date (still uses golang 1.15 images in drone) - can you please rebase?

Ohhfff.. Quite right, comedy of errors! Thanks for your patience, I *think* I got it right this time.

oliver006:
👍

PR #577, opened 2021-11-02
Expose io-threads configuration metric

### LINKED ISSUE #576: Expose metric for `io-threads` config param.
**Describe the problem**

Redis 6+ added support for multi threaded IO. This is configurable via `io-threads` option.

```
127.0.0.1:6379> CONFIG GET io-threads
1) "io-threads"
2) "1"
```

It will be useful to have this eposed via the exporter as well

I can PR this , if it will be an acceptable change
Thanks

shamil:
Fixes #576


shamil:
@oliver006 as discussed in #576 adding `io-threads` configuration metric. Let me know if any changes needed.
Thanks

oliver006:
Thank you, the PR looks good.
Can you add a small test to this, in http_test.go around line 146 to get this covered? Thanks!

shamil:
> Can you add a small test to this, in http_test.go around line 146 to get this covered? Thanks!

Hi @oliver006 I've added test, but I had to make it only for `redis` 6+, otherwise tests will fail, since `io-threads` supported only in `redis` 6 and up

oliver006:
The test instance should run Redis 6 so I don't think you need all the checks.

shamil:
> The test instance should run Redis 6 so I don't think you need all the checks.

Currently there are 2 tests that use `TEST_PWD_REDIS_URI`, which is redis 5 according to [Makefile](https://github.com/oliver006/redis_exporter/blob/master/Makefile#L32). Do you think we should change that to 6?

oliver006:
Ugh, you're right. In that case, just remove the test, not worth pulling in all that extra stuff.

shamil:
> Ugh, you're right. In that case, just remove the test, not worth pulling in all that extra stuff.

Personally I don't think this is too much, maybe we should keep this, it's good ground for future uses in case there will be a need to run tests in this file based on redis version. If you think otherwise, I will  remove....

oliver006:
Let's just remove it.
If we want to check both then I'd think that the tests should be structured quite differently, like have a list of "wants" per host that is being scraped instead of the one-size-fits-all approach that we have right now.

shamil:
Alright, removed in https://github.com/oliver006/redis_exporter/pull/577/commits/3e98fda8a91cf911e8e789c36b1b3186594328e5

oliver006 (approved):
Looks good!

oliver006:
Released as [v1.30.0](https://github.com/oliver006/redis_exporter/releases/tag/v1.30.0)

Issue #585, opened 2021-11-23
Redis Stream item ids


Lupunita:
**Describe the problem** 
The exporter does not provide information on Redis Stream item ids:
 
- [last-generated-id ](https://redis.io/commands/xinfo-stream) for stream
- [last-delivered-id ](https://redis.io/commands/xinfo-groups)  for stream consumer group

These two metrics can estimate how much are services keeping up with the traffic. And for instance, it would be possible to build a scaling logic based on that. 

The ID consists of a timestamp and a number of items added/delivered under some timestamp. I propose to parse just the timestamp as a unique ID.  `1538385846314-2` -> `1538385846314`

The custom ID would be ignored and metrics wouldn't return anything. 

Is it possible to add this? I have it prepared locally for PR if that would be ok? 

Thanks!

oliver006:
Thanks for raising this issue.
Would you export these as metrics (a gauge) or as label values? 
For label values this could be problematic as this would generate a very high cardinality for the labels, something that Prometheus doesn't handle well.

Lupunita:
No, definitively metrics. I am currently exporting these as `stream_last_generated_id` and `stream_group_last_delivered_id` metrics.

oliver006:
Sounds good - I'd be happy to review and merge that PR.

Lupunita:
Could you please give me access rights? 
Thanks!

oliver006:
Fork the repo, create a branch in your repo, open a PR.

oliver006:
Closed by #588 - thanks @Lupunita !

PR #588, opened 2021-11-28
Exteding redis stream metrics for item ID's

Lupunita:
PR for https://github.com/oliver006/redis_exporter/issues/585

@oliver006


oliver006:
Thanks for the PR - I'll try to review it later today.

exporter/streams.go
```diff
@@ -4,22 +4,26 @@ import (
 	"github.com/gomodule/redigo/redis"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
+	"strconv"
```
oliver006, line 7: Can you group and sort the stdlib imports separately please? 
[This is a good example ](https://github.com/oliver006/redis_exporter/blob/master/main.go#L3-L17)if what it should look like.

Note to self: the linter should catch this, not sure why it didn't, I need to check what's up and fix it.

exporter/streams.go
```diff
@@ -4,22 +4,26 @@ import (
...
+	LastGeneratedId  string `redis:"last-generated-id"`
+	Groups           int64  `redis:"groups"`
 	StreamGroupsInfo []streamGroupsInfo
 }
 
 type streamGroupsInfo struct {
 	Name                     string `redis:"name"`
 	Consumers                int64  `redis:"consumers"`
 	Pending                  int64  `redis:"pending"`
+	LasDeliveredtId          string `redis:"last-delivered-id"`
```
oliver006, line 26: `LasDeliveredtId` --> `LastDeliveredId` ?

exporter/streams.go
```diff
@@ -111,6 +115,14 @@ func scanStreamGroupConsumers(c redis.Conn, stream string, group string) ([]stre
 	return result, nil
 }
 
+func ParseId(id string) float64 {
+	parsed_id, err := strconv.ParseFloat(strings.Split(id, "-")[0], 64)
```
oliver006, line 119: You might want to break out the `strings.Split(id, "-")[0]` and check that you actually have >= 1 resu;ts from the `split()`.

exporter/streams.go
```diff
@@ -111,6 +115,14 @@ func scanStreamGroupConsumers(c redis.Conn, stream string, group string) ([]stre
 	return result, nil
 }
 
+func ParseId(id string) float64 {
```
oliver006, line 118: Maybe give this a little more descriptive name like `parseStreamId`?

exporter/streams_test.go
```diff
@@ -1,13 +1,12 @@
...
 
 import (
-	"os"
-	"strings"
-	"testing"
-
 	"github.com/gomodule/redigo/redis"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
+	"os"
```
oliver006, line 7: Same as above.

exporter/streams.go
Lupunita: Sure, this wasn't changed intentionally. It made my local linter edit on save. My first GO project so I still working on tuning my dev env.

exporter/streams.go
Lupunita: Right. changed to `parseStreamItemID` in my last commit.

exporter/streams.go
Lupunita: changed to `parseStreamItemID` in 70664ce

exporter/streams_test.go
Lupunita: fixed in 70664ce

exporter/streams.go
```diff
@@ -9,17 +12,19 @@ import (
...
+	LastGeneratedId  string `redis:"last-generated-id"`
+	Groups           int64  `redis:"groups"`
 	StreamGroupsInfo []streamGroupsInfo
 }
 
 type streamGroupsInfo struct {
 	Name                     string `redis:"name"`
 	Consumers                int64  `redis:"consumers"`
 	Pending                  int64  `redis:"pending"`
+	LasDeliveredId           string `redis:"last-delivered-id"`
```
oliver006, line 27: LasDeliveredId --> LastDeliveredId

exporter/streams.go
```diff
@@ -111,6 +116,14 @@ func scanStreamGroupConsumers(c redis.Conn, stream string, group string) ([]stre
 	return result, nil
 }
 
+func parseStreamItemID(id string) float64 {
```
oliver006, line 119: parseStreamItemID --> parseStreamItemId

exporter/streams.go
oliver006: I think you have to check for this, otherwise this has the ability to crash the exporter.

oliver006 (commented):
Looks good, two minor nits, otherwise it's good to go.

exporter/streams.go
Lupunita: I am looking at it, but cannot figure out the case in which it can crash the exporter. The item ID must be int and it cannot be 0. The func  strconv.ParseFloa returns 0 on strings. In which case it can crash the exporter?

oliver006: Oh you're right! Split() should always return at least one item here so you're fine, my bad.

oliver006:
Looks good, thanks!

PR #593, opened 2021-12-15
Add support of TLS with mutual authentication on the metrics endpoint

bjosv:
This PR adds support of TLS with mutual authentication on the metrics endpoint, which in practice adds authentication of a connecting client. This is sometimes loosely called "mTLS" or "mutual TLS" and is often required in enterprise setups.
The connection towards Redis is already using mutual authentication, which is enabled by default in Redis.

The proposal for enabling this is to use a similar concept as how to enable TLS; by configuring the file path to a CA cert file.
When the (new) config: `tls-server-ca-cert-file`
or the (new) environment variable: `REDIS_EXPORTER_TLS_SERVER_CA_CERT_FILE`
is configured a client connecting to the metrics endpoint will be authenticated.

The CA file is also hot-reloaded in the same manner as the server cert and keys.

There is a need to update the docs, but I can do that after feedback regarding the configuration names or in general. There might be a need to clarify that K8s Liveness/Readinessprobes does not support TLS with mutual authentication directly but it can be handled by probing differently, like using `curl`

Example commands for manual tests:
```
# This command will now not be accepted, works in legacy:
curl -vvv --key /tls-data/curl.key --cacert /tls-data/ca.crt https://$PODIP:9121/metrics

# The --cert flag needs to be given for it to work, like:
curl -vvv --cert /tls-data/curl.crt --key /tls-data/curl.key --cacert /tls-data/ca.crt https://$PODIP:9121/metrics
```


oliver006:
Thanks for the PR! I'm a little short with time right now but will try to get to this in the next few days.

oliver006:
This looks good. I think once you added the `REDIS_EXPORTER_TLS_SERVER_CA_CERT_FILE` flag to the README we should be good to go.

oliver006 (approved):
🚀

PR #600, opened 2022-01-12
Added support for ERRORSTATS and COMMANDSTATS fields introduced on v6.2

filipecosta90:
This PR pushes forward the observability on overall error statistics and command statistics within the prometheus redis exporter. The features discussed on this PR were introduced in Redis v6.2 ( specifically after PR https://github.com/redis/redis/pull/8217 )
It:
- Includes the extension of `INFO COMMANDSTATS` to parse:
  - failed_calls in - so we can keep track of errors that happen from the command itself, broken by command.
  - rejected_calls - so we can keep track of errors that were triggered outside the command processing per se ( example of wrong arity, oom, etc... )
- Parses a new section to INFO, named `ERRORSTATS` that enables keeping track of the different errors that occur within redis ( within processCommand and call ) based upon the reply Error Prefix ( The first word after the "-", up to the first space ).


.drone.yml
```diff
@@ -23,6 +23,14 @@ services:
     ports:
       - 6379
 
+  - name: redis6-2
```
oliver006, line 26: I think you can just change the current `redis6` redis instance to use redis 6.2, no need for a separate test instance.

Makefile
```diff
@@ -35,7 +36,7 @@ test:
 	TEST_REDIS_CLUSTER_SLAVE_URI="redis://redis-cluster:7005" \
 	TEST_TILE38_URI="redis://tile38:9851" \
 	TEST_REDIS_SENTINEL_URI="redis://redis-sentinel:26379" \
-	go test -v -covermode=atomic -cover -race -coverprofile=coverage.txt -p 1 ./...
+	go test -covermode=atomic -cover -race -coverprofile=coverage.txt -p 1 ./...
```
oliver006, line 39: Can you please leave `-v` to make it easier to find issues with tests?

exporter/info.go
```diff
@@ -274,47 +278,111 @@ func (e *Exporter) handleMetricsServer(ch chan<- prometheus.Metric, fieldKey str
...
+
+	failedCalls, err = extractVal(splitValue[4])
+	if err != nil {
+		errorOut = errors.New("Invalid splitValue[4]")
+		return
 	}
+	extendedStats = true
+	return
+}
+func parseMetricsErrorStats(fieldKey string, fieldValue string) (errorPrefix string, count float64, errorOut error) {
```
oliver006, line 347: `errorPrefix` - better name might be `errorType` or `errorName`

exporter/info.go
```diff
@@ -274,47 +278,111 @@ func (e *Exporter) handleMetricsServer(ch chan<- prometheus.Metric, fieldKey str
...
-	return cmd, calls, usecTotal, nil
+		broken up like this:
+			fieldKey  = errorstat_ERR
+			fieldValue= count=3
+	*/
+
+	const prefix = "errorstat_"
+
+	if !strings.HasPrefix(fieldKey, prefix) {
+		errorOut = errors.New("Invalid fieldKey")
```
oliver006, line 361: Both the errors.New() here and in line 367 should include what the invalid field was so the error message is more useful.

oliver006: Same for all the errors.New() in parseMetricsCommandStats()

exporter/info.go
```diff
@@ -274,47 +278,111 @@ func (e *Exporter) handleMetricsServer(ch chan<- prometheus.Metric, fieldKey str
...
 	splitValue := strings.Split(fieldValue, ",")
-	if len(splitValue) < 3 {
-		return "", 0.0, 0.0, errors.New("Invalid fieldValue")
+	splitLen := len(splitValue)
+	if splitLen < 3 {
+		errorOut = errors.New("Invalid fieldValue")
+		return
 	}
 
 	calls, err := extractVal(splitValue[0])
```
oliver006, line 316: Don;'t you need to change this to `calls, err =` (instead of `:=`) so you don't shadow the `calls` variable?

exporter/info.go
```diff
@@ -274,47 +278,111 @@ func (e *Exporter) handleMetricsServer(ch chan<- prometheus.Metric, fieldKey str
...
+	}
+
+	usecTotal, err = extractVal(splitValue[1])
+	if err != nil {
+		errorOut = errors.New("Invalid splitValue[1]")
+		return
 	}
 
-	usecTotal, err := extractVal(splitValue[1])
+	// If we're parsing pre v6.2 output
```
oliver006, line 328: This comment could be a bit more clear.
Something like "pre 6.2 did not include rejected/failed calls stats so if we have less than 5 tokens we're done here"

exporter/info.go
```diff
@@ -274,47 +278,111 @@ func (e *Exporter) handleMetricsServer(ch chan<- prometheus.Metric, fieldKey str
...
+	}
+
+	failedCalls, err = extractVal(splitValue[4])
+	if err != nil {
+		errorOut = errors.New("Invalid splitValue[4]")
+		return
 	}
+	extendedStats = true
+	return
+}
```
oliver006, line 346: Missing blank line after end of function.

go.sum
```diff
@@ -45,6 +45,8 @@ github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.5 h1:Khx7svrCpmxxtHBq5j2mp/xVjsi8hQMfNLvJFAlrGgU=
 github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/gofuzz v1.0.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
+github.com/hashicorp/go-version v1.0.0 h1:21MVWPKDphxa7ineQQTrCU5brh7OuVVAzGOCnnCPtE8=
```
oliver006, line 48: Don't think this or the next block (gox) is needed.

oliver006 (commented):
Left a bunch of minor comments but overall this PR looks good.

filipecosta90:
> Left a bunch of minor comments but overall this PR looks good.

@oliver006 thank you for the review. Addressed the comments in the latest PR. can you please review again?

oliver006 (approved):
Nice work! 🚀

PR #604, opened 2022-01-25
Make the redis connection threshold a variable for alerting

ioanav104:
Allow mixin users more flexibility to set the threshold

### DIFF
--- a/contrib/redis-mixin/alerts/redis.libsonnet
+++ b/contrib/redis-mixin/alerts/redis.libsonnet
@@ -30,7 +30,7 @@
           },
           {
             alert: 'RedisTooManyConnections',
-            expr: 'redis_connected_clients > 100',
+            expr: 'redis_connected_clients > %(redisConnectionsThreshold)' % $._config,
             'for': '5m',
             labels: {
               severity: 'warning',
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR! The linter complained here: 
https://cloud.drone.io/oliver006/redis_exporter/1442/1/16

Can you make sense of that error? I'm not really familiar with the mixin stuff...

ioanav104:
I'm having a look

contrib/redis-mixin/alerts/redis.libsonnet
```diff
@@ -30,7 +30,7 @@
           },
           {
             alert: 'RedisTooManyConnections',
-            expr: 'redis_connected_clients > 100',
+            expr: 'redis_connected_clients > %(redisConnectionsThreshold)' % $._config,
```
kragniz, line 33:
```suggestion
            expr: 'redis_connected_clients > %(redisConnectionsThreshold)s' % $._config,
```

kragniz: jsonnet uses the same mini-language as python for string formatting, so you need to specify it's a string

https://docs.python.org/2/library/stdtypes.html#string-formatting

oliver006:
Thanks for getting this sorted out!

PR #619, opened 2022-03-09
Add main_thread metrics added in 6.2

artificial-aidan:
Main thread cpu time was added in 6.2

https://raw.githubusercontent.com/antirez/redis/6.2/00-RELEASENOTES


oliver006 (approved):
👍 

Thanks for the PR!

Issue #620, opened 2022-03-11
[Chore] Add new metrics to redis exporter


oliver006:
When running `INFO` I see quite a few new metrics like 
```
unexpected_error_replies:0
total_error_replies:0
dump_payload_sanitizations:0
total_reads_processed:2
total_writes_processed:1
io_threaded_reads_processed:0
io_threaded_writes_processed:0
```
and probably quite a few more.

Should add these to the exporter.

oliver006:
Absolutely, all help is appreciated, I'll have a look at your PR soon, thanks!

oliver006:
Closed by #621

PR #621, opened 2022-03-14
Add new metrics of info cmd to redis exporter

jjz921024:
#620


oliver006:
Thanks for opening the PR!
Are there any other metrics we can add or just the seven you added?

oliver006:
Thanks for opening the PR!
Are there any other metrics we can add or just the seven you added?

jjz921024:
> Thanks for opening the PR! Are there any other metrics we can add or just the seven you added?

maybe is there a way for users to configure the metrics they're interested in?
It's just an idea, i don't know if it is available.

oliver006:
Eh, don't make anything optional but there are a few more metrics you could add.
Like these for example:
```
io_threads_active
clients_in_timeout_table
number_of_cached_scripts
tracking_total_keys
tracking_total_items
tracking_total_prefixes
total_reads_processed
total_writes_processed
```

oliver006:
Nice work - thanks!

PR #634, opened 2022-04-27
Allow the ability to customize redis exporter selector

danlenar:
Most mixins allow to customize selectors as such as seen in the Customising the mixin section in
https://monitoring.mixins.dev/


oliver006:
Sorry I'm a little late here. Thanks for the PR!

Issue #638, opened 2022-05-05
Add Redis address to scrapeRedisHost error message


neomantra:
If a Redis host is not [available to scrape](https://github.com/oliver006/redis_exporter/blob/a1c28b775760f2f00fce07a24db7fd4e83c26b9f/exporter/exporter.go#L530), this error is logged:
```
Couldn't connect to redis instance
```

Looking through issues here, it seems this is often because of `--redis.addr=` and the suggestion is to enable `--debug` to check.   However, that generates a *ton* of output (which is great when one needs it). 

 In my case, we had a misconfigured static target and simply having the host in the message would have helped.

Is there any reason (security?) the message cannot be like:
```
log.Errorf("Couldn't connect to redis instance %s", e.redisAddr)
```

**What version of redis_exporter are you running?**
Please run `redis_exporter --version` if you're not sure what version you're running.
[ ] 0.3x.x
[X] 1.x.x
docker image v1.24.0

Thanks!

oliver006:
Reason is that you can put the password in the URI and logging the full URI would leak the password to the logs.

An improvement could be to log the URI with the password redacted. Happy to review/merge a PR for that but I'm currently short of time to implement something myself.

neomantra:
That makes sense -- I didn't want to proceed with a PR without understanding it...  and now I made the PR :)

oliver006:
Closed by #643

PR #643, opened 2022-05-14
Add redacted addr to scrapeRedisHost error message (#638)

neomantra:
When `scrapeRedisHost` fails to connect to a Redis instance, the code now tries to redact the password of the Redis address (URL).  If it can parse and redact it, it will print the address in the error message.   Otherwise, we note that `-debug` is needed to see the address.   It does not log any URL parsing errors, just in case that may leak any info.

Testing:
```
> ./redis_exporter  -redis.addr "redis://user:peace@foobar.com:9999"
INFO[0000] Redis Metrics Exporter <<< filled in by build >>>    build date: <<< filled in by build >>>    sha1: <<< filled in by build >>>    Go: go1.18.1    GOOS: darwin    GOARCH: arm64 
INFO[0000] Providing metrics at :9121/metrics           
ERRO[0018] Couldn't connect to redis instance (redis://user:xxxxx@foobar.com:9999)
```

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -3,6 +3,7 @@ package exporter
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


exporter/exporter.go
```diff
@@ -527,7 +528,13 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 	e.registerConstMetricGauge(ch, "exporter_last_scrape_connect_time_seconds", connectTookSeconds)
 
 	if err != nil {
-		log.Errorf("Couldn't connect to redis instance")
+		var redactedAddr string
+		if redisURL, _ := url.Parse(e.redisAddr); redisURL != nil {
```
oliver006, line 532: Maybe catch the error here instead of checking the result for nil? You could then output the error in the else block in log level debug to add addtl information.

oliver006 (commented):
Looks good, just one comment

oliver006:
@neomantra  - do you still plan to make changes on this? I think surfacing the error would be good.

neomantra:
I’ll incorporate that feedback today, thanks.

neomantra:
I updated this per your comment and rebased to latest master.

oliver006 (approved):
🎉

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.39.0

PR #644, opened 2022-05-16
Add /scrape-all endpoint

roman-vynar:
Hello,

A bit history: in the past redis_exporter was capable to scrape multiple endpoints at once. Later it was removed and a tradeoff has been made by implementing `-redis.password-file` option + putting all the efforts on Prometheus to scrape multiple targets by using /scrape endpoint with `target` query arg.

Here I propose a very minimal PR which basically re-uses the existing `-redis.password-file` and adds a new endpoint /scape-all in order to scrape all redis endpoints listed in the password file:

* minimalistic PR
* no changes to the existing functionality, nothing breaks
* whoever wants to scrape multiple instances at once can do this with the least efforts
* helps in performance as we don't need to add a ton of targets into Prometheus (we run 30 redis dbs per host) 

When you want scrape all defined endpoints, simply do this by going into /scrape-all. Because metrics need to be unique per endpoint we add `addr` label with the actual endpoint url. Optionally, you can add other custom labels via query arg, e.g. the file entry `{"redis://127.0.0.1:36379?db=foobar": "mypassword"}` will produce metrics with labels `{addr="redis://127.0.0.1:36379", db="foobar", ...}`

Thanks!

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -3,6 +3,7 @@ package exporter
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 	"runtime"
 	"strconv"
 	"strings"
@@ -27,6 +28,8 @@ type Exporter struct {
 
 	redisAddr string
 	namespace string
+
+	constLabels prometheus.Labels
 
 	totalScrapes              prometheus.Counter
 	scrapeDuration            prometheus.Summary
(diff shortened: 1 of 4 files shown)


roman-vynar:
Any feedback on this?

Please note that adding /scrape-all doesn't make this feature anyhow wrong or not following "the right guidelines" because there is an existing password file option for setting of multiple addresses and passwords and this serves already a special purpose so why not to re-use it for getting everything scraped especially when it is optional.

PR #651, opened 2022-05-22
Update README.md

stobbe99:
Hello Oliver,

Just added one line to make live a little easier for the next newbee,

KR Henk


oliver006:
Thank you!

PR #656, opened 2022-05-31
fix #650, add db_keys_cached gauge.

### LINKED ISSUE #650: KeyDB get information about keys per DB
**Describe the problem**
The `redis_db_keys` and `redis_db_keys_expiring` always zero.

```
INFO[0000] Redis Metrics Exporter v1.37.0    build date: 2022-03-18-01:20:01    sha1: a1c28b775760f2f00fce07a24db7fd4e83c26b9f    Go: go1.17.8    GOOS: linux    GOARCH: amd64
```

**Running the exporter**
No any args and env variables.

**Screenshots**
```
172.31.130.70:6379> INFO Keyspace
# Keyspace
db0:keys=12867,expires=12833,avg_ttl=3465683577997863,cached_keys=12867
```

**Additional context**
KeyDB server:

```
KeyDB server v=6.3.0 sha=cbc7a895:1
```

vixns:


exporter/info_test.go
```diff
@@ -17,7 +17,7 @@ func TestKeyspaceStringParser(t *testing.T) {
 	tsts := []struct {
 		db                        string
 		stats                     string
-		keysTotal, keysEx, avgTTL float64
+		keysTotal, keysEx, keysCached, avgTTL float64
```
oliver006, line 20: Indentation here seems off. Not sure why the linter didn't catch.

oliver006:
Thanks for the PR @vixns !
Can you please fix the linting issue and see if you can bump the test coverage a bit? Thanks!

exporter/info_test.go
vixns: fixed

exporter/info.go
```diff
@@ -100,11 +100,12 @@ func (e *Exporter) extractInfoMetrics(ch chan<- prometheus.Metric, info string,
 			continue
 
 		case "Keyspace":
-			if keysTotal, keysEx, avgTTL, ok := parseDBKeyspaceString(fieldKey, fieldValue); ok {
+			if keysTotal, keysEx, avgTTL, keysCached, ok := parseDBKeyspaceString(fieldKey, fieldValue); ok {
 				dbName := fieldKey
 
 				e.registerConstMetricGauge(ch, "db_keys", keysTotal, dbName)
 				e.registerConstMetricGauge(ch, "db_keys_expiring", keysEx, dbName)
+				e.registerConstMetricGauge(ch, "db_keys_cached", keysCached, dbName)
```
oliver006, line 108: Sorry, I missed this earlier.
This will start registering a new metric for all types of targets, even if it's not a KeyDB instance (and vanilla Redis doesn't have cached keys, right?).
Can you modify this to only emit the metric if it was found? 
Could be as simple as initializing `keysCachedTotal = -1` in line 224 and then add a check around line 108 to only record the metric if `keysCached > -1`

vixns: Done, but the test coverage is decreasing.

oliver006: I think this is fine, thanks for updating the code!

oliver006:
There's a coverage check warning here: https://github.com/oliver006/redis_exporter/pull/656/files#diff-01814cc33b1a90b54048d89aaf89507891e8dc68a953a859cdb3b7985b3e906bR110

Which means that none of the tests emit the cached_key metrics.
Did you verify this is working by running it locally? One of the tests should be hitting a KeyDB instance so not clear why this is not covered.
Just want to make sure this is working.

vixns:
> There's a coverage check warning here: https://github.com/oliver006/redis_exporter/pull/656/files#diff-01814cc33b1a90b54048d89aaf89507891e8dc68a953a859cdb3b7985b3e906bR110
> 
> Which means that none of the tests emit the cached_key metrics. Did you verify this is working by running it locally? One of the tests should be hitting a KeyDB instance so not clear why this is not covered. Just want to make sure this is working.

Tested with keydb 6.3.1,

oliver006 (commented):
🚀

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.40.0

PR #660, opened 2022-06-02
expose instantaneous_ops_per_sec gauge

naughtyGitCat:
the Redis info metric instantaneous_ops_per_sec is a instant guage shows current ops per sec, more accurate than irate(redis_commands_processed_total[range]}

refs:
-  https://github.com/redis/redis/issues/3746
- `instantaneous_ops_per_sec`: Number of commands processed per second https://redis.io/commands/info/


oliver006:
Thank for the PR!
I don't think this is needed as you can calculate the ops per second from other exported metrics, 
for instance query `rate(redis_commands_processed_total[1m])`.
Any particular reason you need this metric?

naughtyGitCat:
Thanks for your attention,
1. high definition metrics may lead to pressure on the Redis server, due to some hard time, when we slow down the scrape, the redis_commands_processed_total does not accurate
2.  neither rate nor irate can be more accurate than the instantaneous_ops_per_sec, in our production environment, the value between irate and instantaneous_ops_per_sec can not be ignored

after the weekend, I will post the different grafana timeseries which calc by irate and  instantaneous_ops_per_sec

PR #662, opened 2022-06-14
print version to stdout when -version flag is used

### LINKED ISSUE #646: The output for version flag is printed on stderr
Was facing an issue with the version check for the exporter. But found a similar issue on node_exporter issues which is already fixed. Linking the same here https://github.com/prometheus/node_exporter/issues/1271

Can we do the same fix here?

ethanmdavidson:
otherwise, print it to stderr as normal.

fixes #646

### DIFF
--- a/main.go
+++ b/main.go
@@ -100,21 +100,23 @@ func main() {
 	default:
 		log.SetFormatter(&log.TextFormatter{})
 	}
+	if *showVersion {
+		log.SetOutput(os.Stdout)
+	}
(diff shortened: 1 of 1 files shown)


oliver006:
Thank you!

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.41.0

nkn1111:
Thank you so much @ethanmdavidson @oliver006

PR #666, opened 2022-06-17
Add configuration to set minimum TLS version accepted by the metrics server port.

bjosv:
This adds the configuration `tls-server-min-version` to be able to set the minimum TLS version that will be accepted by the metrics server endpoint.

Since TLS 1.0 and 1.1 have reached end-of-life they are normally not allowed within some organisations, but Go still [enables](https://github.com/golang/go/blob/6c25ba624fd032a20dcfa94f9f9f0ae32c57c54b/src/crypto/tls/common.go#L686-L688) it by default on TLS servers.
Due to the [deprecation](https://datatracker.ietf.org/doc/rfc8996/) this PR also changes the default behavior and rejects clients that request to only use the old less secure TLS versions. This can be opened again by the configuration (setting `tls-server-min-version` to `TLS1.0`).

This PR also contains a commit to fix the local testruns using the makefile.
Maybe we should add a run of `make docker-all` in CI?


oliver006 (approved):
Thanks for the PR, change looks good and thx for fixing the docker-compose file.

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.43.0

bjosv:
> Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.43.0

You are the best. Thanks!

PR #677, opened 2022-07-15
Adding "instantaneous_ops_per_sec" from `info stats`

samuelclay:
A handy metric for gauging load on Redis.


oliver006:
Thanks for opening the PR. 
The `instantaneous_ops_per_sec` metric can be derived by querying Prometheus with `rate(redis_commands_processed_total[1m])` and we shouldn't record derived metrics but have one source of truth.

See https://github.com/oliver006/redis_exporter/pull/660 for a previous discussion. There seems to be a concern that `instantaneous_ops_per_sec` is more accurate than querying Prometheus but I haven't really looked into that.

PR #691, opened 2022-08-15
feat: add Redis Sentinel metrics ckquorum

opan:
Expose new metrics for Redis Sentinel from command `SENTINEL CKQUORUM <master-name>`

### DIFF
--- a/exporter/sentinels.go
+++ b/exporter/sentinels.go
@@ -68,10 +68,10 @@ func (e *Exporter) extractSentinelMetrics(ch chan<- prometheus.Metric, c redis.C
 
 		masterCkquorum, err := redis.String(doRedisCmd(c, "SENTINEL", "CKQUORUM", masterName))
 		log.Debugf("Sentinel ckquorum status for master %s: %s %s", masterName, masterCkquorum, err)
-		masterCkquorumStatus := 0
+		masterCkquorumStatus := 1
 		masterCkquorumMsg := masterCkquorum
 		if err != nil {
-			masterCkquorumStatus = 1
+			masterCkquorumStatus = 0
 			masterCkquorumMsg = err.Error()
 		}
 		e.registerConstMetricGauge(ch, "sentinel_master_ckquorum_status", float64(masterCkquorumStatus), masterName, masterCkquorumMsg)
--- a/exporter/sentinels.go
+++ b/exporter/sentinels.go
@@ -66,10 +66,9 @@ func (e *Exporter) extractSentinelMetrics(ch chan<- prometheus.Metric, c redis.C
 		}
 		masterAddr := masterIp + ":" + masterPort
 
-		masterCkquorum, err := redis.String(doRedisCmd(c, "SENTINEL", "CKQUORUM", masterName))
-		log.Debugf("Sentinel ckquorum status for master %s: %s %s", masterName, masterCkquorum, err)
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR! I'll try to get to reviewing it in the next few days.

opan:
@oliver006 Thanks!

exporter/sentinels.go
```diff
@@ -66,6 +66,16 @@ func (e *Exporter) extractSentinelMetrics(ch chan<- prometheus.Metric, c redis.C
 		}
 		masterAddr := masterIp + ":" + masterPort
 
+		masterCkquorum, err := redis.String(doRedisCmd(c, "SENTINEL", "CKQUORUM", masterName))
+		log.Debugf("Sentinel ckquorum status for master %s", masterName)
```
oliver006, line 70: I think this log message is in the wrong place or not capturing enough info. Should this output `masterCkquorum` and `err` as well?

exporter/sentinels.go
```diff
@@ -66,6 +66,16 @@ func (e *Exporter) extractSentinelMetrics(ch chan<- prometheus.Metric, c redis.C
...
+		masterCkquorumStatus := 0
+		masterCkquorumMsg := masterCkquorum
+		if err != nil {
+			masterCkquorumStatus = 1
```
oliver006, line 74: What's the intention with setting this to zero and one depending on the error? 

From a cursory look t the docs, looks like if we have `ckquorum` then things are ok, so make it `1`?

exporter/sentinels.go
opan: Hi, sorry for the late response. Oh yes, I think you're right. I will update it soon

exporter/sentinels.go
opan: I was planning to use this metric as a boolean value. If it's 0 meaning the ckquorum result is okay (meaning `err` is empty), otherwise (`err` value is not empty), the master is not okay or can't be failover

oliver006: I see. Would it make more sense to set it to "1" if things are up and ok? Just like the "up" metric?

opan:
@oliver006 I just push a new commit contains fix for the log debug for `masterCkquorum` metric

exporter/sentinels.go
opan: I think you're right. Also forgot to lookup other metrics with a similar setup. I will update this now.

opan: Updated

exporter/sentinels.go
```diff
@@ -66,6 +66,16 @@ func (e *Exporter) extractSentinelMetrics(ch chan<- prometheus.Metric, c redis.C
 		}
 		masterAddr := masterIp + ":" + masterPort
 
+		masterCkquorum, err := redis.String(doRedisCmd(c, "SENTINEL", "CKQUORUM", masterName))
```
oliver006, line 69: Make this `masterCkquorumMsg, err := ` and remove line 72? (keep 75)

oliver006 (commented):
Looks good. One last nit and then this should be good to go.

exporter/sentinels.go
opan: Make sense. On it

oliver006: 🎉

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.44.0

PR #696, opened 2022-08-25
Update README.md

gberche-orange:
Fix small typo on tls-server-ca-cert-file arg which is used for client to validate server cert, and not for client auth (this is instead the tls-ca-cert-file flag)


oliver006 (approved):
🎉

oliver006:
Thank you!

PR #710, opened 2022-10-25
keydb multimaster support

mihaiush:
Add support for info keys like master_[0-9]_ which are produced by keydb in case there are more than 1 master.
```
redis_master_link_up{master_host="m0.server.local",master_port="6377"} 1
redis_master_link_up{master_host="m1.server.local",master_port="6377"} 1
```

### DIFF
--- a/exporter/info.go
+++ b/exporter/info.go
@@ -67,11 +67,21 @@ func (e *Exporter) extractInfoMetrics(ch chan<- prometheus.Metric, info string, 
 
 		keyValues[fieldKey] = fieldValue
 
-		if fieldKey == "master_host" {
+        // keydb multimaster
+        // master_host:kdb0.server.local
+        // master_port:6377
+        // master_1_host:kdb1.server.local
+        // master_1_port:6377        
+
+        r := false
+        r,_ = regexp.MatchString("^master(_[0-9]+)?_host", fieldKey)
+		if r {
 			masterHost = fieldValue
 		}
 
-		if fieldKey == "master_port" {
+        r = false
+        r,_ = regexp.MatchString("^master(_[0-9]+)?_port", fieldKey)
+		if r {
 			masterPort = fieldValue
 		}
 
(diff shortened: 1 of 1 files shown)


oliver006 (commented):
Thanks for the PR! Looks good but a couple of smaller issues:

- There are a couple of comments in the linter that make sense (see here: https://cloud.drone.io/oliver006/redis_exporter/1709/1/15)

- Can you also please run "go fmt" on your change to make sure it's properly formatted?

exporter/info.go
```diff
@@ -279,17 +303,22 @@ func parseConnectedSlaveString(slaveName string, keyValues string) (offset float
...
 		}
 		return true
 	}
-	switch fieldKey {
 
-	case "master_last_io_seconds_ago", "slave_repl_offset", "master_sync_in_progress":
+	if reMasterDirect.MatchString(fieldKey) {
+		if strings.HasSuffix(fieldKey, "last_io_seconds_ago") {
+			fieldKey = "master_last_io_seconds_ago"
+		}
```
oliver006, line 318: nit: this should be an else (or use `switch` for the whole block`)

exporter/info.go
```diff
@@ -240,7 +263,8 @@ func parseDBKeyspaceString(inputKey string, inputVal string) (keysTotal float64,
 */
 func parseConnectedSlaveString(slaveName string, keyValues string) (offset float64, ip string, port string, state string, lag float64, ok bool) {
 	ok = false
-	if matched, _ := regexp.MatchString(`^slave\d+`, slaveName); !matched {
+	if !reSlave.MatchString(slaveName) {
+		//if matched, _ := regexp.MatchString(`^slave\d+`, slaveName); !matched {
```
oliver006, line 267: nit: can you remove this line plz?

oliver006 (commented):
Looks good, thanks for the changes. Two minor issues and then this should be good to go!

PR #751, opened 2023-01-23
Add securityContext to kubernetes example for

sergeyshevch:
I suggest adding security context to the Kubernetes example. It will allow users to easier find the required user/group and more securely and deploy this exporter into their clusters.

I tested this securityContext in my deployment. All works fine


oliver006:
Thanks for the PR!

PR #770, opened 2023-02-24
Addition of latency_usec histogram metric

### LINKED ISSUE #765: Support for the LATENCY HISTOGRAM Redis metrics
Hi!

What do you think about exporting the LATENCY HISTOGRAM metrics added in Redis 7.0.0 (https://redis.io/commands/latency-histogram/)? This would map nicely to a Prometheus histogram metric, perhaps `redis_latency_usec_bucket` or similar, and is something we could use! :)

LarssonOliver:
Hi!

I had a go at implementing an exporter for the `LATENCY HISTOGRAM` metrics added in Redis 7.

Feel free to roast, I'm fairly inexperienced with go.

One thing to note: The histogram metric wants a `sum` of all the latencies observed. This is not provided by `LATENCY HISTOGRAM` but rather by `INFO COMMANDSTATS`. Optimally, this should be passed on after it has been gathered from `INFO ALL`, or if not deemed too costly and to improve readability, queried here.

Fixes #765


oliver006:
Thanks for the PR!
I'll try to have a look and leave some comments in the coming days.

exporter/metrics.go
```diff
@@ -80,27 +80,44 @@ func (e *Exporter) registerConstMetricGauge(ch chan<- prometheus.Metric, metric
...
 
-	if m, err := prometheus.NewConstMetric(descr, valType, val, labelValues...); err == nil {
+	if m, err := prometheus.NewConstMetric(description, valType, val, labelValues...); err == nil {
 		ch <- m
 	}
 }
 
 func (e *Exporter) registerConstSummary(ch chan<- prometheus.Metric, metric string, labelValues []string, count uint64, sum float64, latencyMap map[float64]float64, cmd string) {
-	descr := e.metricDescriptions[metric]
-	if descr == nil {
```
oliver006, line 95: same

exporter/metrics.go
```diff
@@ -80,27 +80,44 @@ func (e *Exporter) registerConstMetricGauge(ch chan<- prometheus.Metric, metric
...
+	histogram := prometheus.MustNewConstHistogram(
+		description,
+		count, sum,
+		buckets,
+		cmd,
+	)
+	ch <- histogram
+}
+
+func (e *Exporter) findOrCreateMetricDescription(metricName string, labels []string) *prometheus.Desc {
```
oliver006, line 115: 👍

exporter/latency.go
```diff
@@ -30,9 +35,77 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
-				e.registerConstMetricGauge(ch, "latency_spike_last", float64(spikeLast), eventName)
-				e.registerConstMetricGauge(ch, "latency_spike_duration_seconds", spikeDurationSeconds, eventName)
+				exporter.registerConstMetricGauge(outChan, "latency_spike_last", float64(spikeLast), eventName)
+				exporter.registerConstMetricGauge(outChan, "latency_spike_duration_seconds", spikeDurationSeconds, eventName)
 			}
 		}
 	}
 }
+
+func extractLatencyHistogramMetrics(exporter *Exporter, outChan chan<- prometheus.Metric, redisConn redis.Conn) {
```
oliver006, line 45: Change this and the other one to make them member functions of `Exporter`, like the other ones.
`func (e *Exporter) extractLatency...`

exporter/latency.go
```diff
@@ -30,9 +35,77 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+
+		details[i].cmd = cmd
+		details[i].details = det
+	}
+
+	for _, detail := range details {
+		var totalCalls uint64
+		var bucketInfo []uint64
+
+		buckets := make(map[float64]uint64)
```
oliver006, line 75: nit: just go with `map[float64]uint64{}` for initializing empty maps.

exporter/latency.go
```diff
@@ -30,9 +35,77 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+		return
+	}
+
+	type cmdDetails struct {
+		cmd     string
+		details []interface{}
+	}
+
+	cmdCount := len(reply) / 2
+	details := make([]cmdDetails, cmdCount)
```
oliver006, line 61: nit: better name for the variable would be `cmdDetails` IMO.

You might not even need the struct, just declare the details array as 

```
        cmdDetails := []struct {
		cmd     string
		details []interface{}
	}{}

Comment (oliver006) [reply exporter/latency.go:61]: But bigger question: why do this in two steps and not just parse and submit the constHistogram instead of storing the intermediate result first?

File: exporter/latency.go [hunk 6]
```diff
@@ -30,9 +35,77 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+		sumEstimated := estimateSum(buckets)
+
+		labelValues := []string{"cmd"}
+		exporter.registerConstHistogram(outChan, "latency_usec", labelValues, totalCalls, sumEstimated, buckets, detail.cmd)
+	}
+}
+
+func estimateSum(buckets map[float64]uint64) float64 {
+	// This should ideally be the total time as reported by info commandstats.
+	// Estimated here instead of querying redis again.
```
oliver006, line 96: Don't think you need to call again.
`extractInfoMetrics()` (which extracts the commandstats) is called right before the latencyMetrics() so you could pass the command stats.
Or leave as is if you think it's good enough.

exporter/latency.go
```diff
@@ -30,9 +35,77 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+
+	sum := .0
+	counted := uint64(0)
+
+	for usec, count := range buckets {
+		bucketCount := count - counted
```
oliver006, line 102: I'm not sure I get this logic.

my understanding is, you're trying to go through the buckets and estimage the avg for each bucket (`bucketAverage`) and then multiply by the times of events in that bucket to get the total latency for that bucket and then sum all the buckets, right?

I'm not sure I understand bucketCount.
But there's a golang idiosyncrasy and that is looping over maps is always in random order of the keys so you don't have any guarantee in which order you'll look at the buckets so I think your "counted=count" logic won't work.

But look at the docs: https://redis.io/commands/latency-histogram/
the buckets increase by power of 2 (1, 2, 4, 8, ...) so you could use that information to look through the buckets (and have better estimates.

Or you could toss this all and use the cmd stats as retrieved earlier (which i think makes more sense)

exporter/metrics.go
LarssonOliver: By "same", do you refer to the rebase comment, or that this should be reverted?

exporter/latency.go
LarssonOliver: My original reasoning was to make it more readable, I now realize after having changed it that it didn't really make a difference on the front :)

exporter/latency.go
LarssonOliver: I no longer do the estimation.

LarssonOliver:
I've gone through and fixed the issues.

To parse the `info all` output, I used a regex that might or might not be a terrible idea :) 

Hope this is more in-line with what you are after!

exporter/latency.go
```diff
@@ -30,9 +39,72 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+	if err != nil {
+		logHistogramErrOnce.Do(func() {
+			log.Errorf("WARNING, LOGGED ONCE ONLY: cmd LATENCY HISTOGRAM, err: %s", err)
+		})
+		log.Debugf("cmd LATENCY HISTOGRAM, err: %s", err)
+		return
+	}
+
+	for i := 0; i < len(reply); i += 2 {
+
```
oliver006, line 60: nit: remove blank line

exporter/latency.go
```diff
@@ -30,9 +39,72 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+		totalUsecs := extractTotalUsecForCommand(infoAll, cmd)
+
+		labelValues := []string{"cmd"}
+		e.registerConstHistogram(outChan, "latency_usec", labelValues, totalCalls, float64(totalUsecs), buckets, cmd)
+	}
+}
+
+func extractTotalUsecForCommand(infoAll string, cmd string) uint64 {
+	sanitizedCmd := strings.ReplaceAll(cmd, "|", "\\|")
```
oliver006, line 87: This looks like we're escaping the pipe symbol. Why do we need this? Can there ever be a pipe symbol in `cmd`? (I don't think so).
If we want a same cmd, maybe just do something like 
`cmd = regexp.MustCompile(`[^a-zA-Z0-9]+`).ReplaceAllString(cmd, "")` (and make the regex a const outside the function so it compiles exactly once)

exporter/latency.go
```diff
@@ -30,9 +39,72 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+	usecRegexp := regexp.MustCompile(`(?m)^cmdstat_` + sanitizedCmd + `(?:\|.*)?:.*usec=([0-9]+).*$`)
+
+	matches := usecRegexp.FindAllStringSubmatch(infoAll, -1)
+
+	if len(matches) == 0 {
+		log.Errorf("Unable to extract total latency for cmd=%s", cmd)
+		return 0
```
oliver006, line 94: Did you test this? Can this cause trouble with the histogram when trying to set a total value of zero?

exporter/exporter.go
```diff
@@ -348,6 +348,7 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 		"commands_rejected_calls_total":                {txt: `Total number of errors within command execution per command`, lbls: []string{"cmd"}},
 		"commands_total":                               {txt: `Total number of calls per command`, lbls: []string{"cmd"}},
 		"latency_percentiles_usec":                     {txt: `A summary of latency percentile distribution per command`, lbls: []string{"cmd"}},
+		"latency_usec":                                 {txt: `A histogram of latencies per command`, lbls: []string{"cmd"}},
```
oliver006, line 351: Maybe make this `command_latency_usec` to be a bit more descriptive?

exporter/metrics_test.go
```diff
@@ -16,3 +19,54 @@ func TestSanitizeMetricName(t *testing.T) {
...
+func TestFindOrCreateMetricsDescriptionFindExisting(t *testing.T) {
+	exp := getTestExporter()
+	exp.metricDescriptions = make(map[string]*prometheus.Desc)
+
+	metricName := "foo"
+	metricDesc := "bar"
+	labels := []string{"1", "2"}
+
+	description := newMetricDescr("ns", metricName, metricDesc, labels)
+	exp.metricDescriptions[metricName] = description
```
oliver006, line 51: Another way to test this without manipulating the `metricDescriptions` directly would be to call `findOrCreateMetricDescription()` twice and check that you still only have one entry in metricDescriptions (len(metricDescriptions) == 1).

exporter/metrics_test.go
```diff
@@ -16,3 +19,54 @@ func TestSanitizeMetricName(t *testing.T) {
...
+
+	for m := range ch {
+		if strings.Contains(m.Desc().String(), metricName) {
+			return
+		}
+	}
+	t.Errorf("Histogram was not registered")
+}
+
+func TestFindOrCreateMetricsDescriptionFindExisting(t *testing.T) {
```
oliver006, line 42: Thanks for adding tests for these!

exporter/latency_test.go
```diff
@@ -104,3 +105,45 @@ func resetLatency(t *testing.T, addr string) error {
...
+	// Since Redis v7 we should have latency histogram stats
+	e := getTestExporterWithAddr(redisSevenAddr)
+	setupDBKeys(t, redisSevenAddr)
+
+	want := map[string]bool{"latency_usec": false}
+	commandStatsCheck(t, e, want)
+	deleteKeysFromDB(t, redisSevenAddr)
+}
+
+func TestExtractTotalUsecForCommand(t *testing.T) {
```
oliver006, line 121: ❤️

oliver006 (commented):
This is shaping up nicely!
Left a few minor comments but this is looking good.

exporter/exporter.go
LarssonOliver: In that case, would it be better to name it `commands_latencies_usec` instead? This would be more in line with the other metrics prefixed with `commands_`. 

We thus have the alternatives:
1. `redis_command_latency_usec_{bucket,sum,count}`
2. `redis_latency_usec_{bucket,sum,count}`
3. `redis_commands_latencies_{bucket,sum,count}`
4. `redis_commands_latency_{bucket,sum,count}`

Let me know your preference :)

exporter/latency.go
LarssonOliver: Yes, I did test this.

Prometheus will just publish `redis_latency_usec_sum{...} = 0`. This does not seem to affect promql queries using `histogram_quantile()` or similar.

It may be better to log this as a warning rather than an error actually.

exporter/latency.go
LarssonOliver: In the current setup, yes, it is needed. As written, the tests will fail if it's not present.

`cmd` will contain a pipe if it describes a sub-command. An example from my setup is:
* `latency|histogram`
* `latency|latest`

The main reason why it has to be escaped is that something like `config|get` will match with `get` which are two separate entries in the `LATENCY HISTOGRAM` output.

LarssonOliver:
Please have a look at the comments I left above :)

One thing to note if this should be merged:

The `latency_usec_{sum,count}` fields (using prefix as placeholder here) are identical to the already existing `commands_total` and `commands_duration_seconds_total` (/ 10e6) metrics already exported. 

My suggestion would be to keep both the old and new ones, as prometheus exports the sum and count metrics for all histograms (https://prometheus.io/docs/instrumenting/exposition_formats/#histograms-and-summaries)

exporter/latency.go
```diff
@@ -30,9 +43,66 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+
+	for i := 0; i < len(reply); i += 2 {
+		cmd := string(reply[i].([]byte))
+		details, _ := redis.Values(reply[i+1], nil)
+
+		var totalCalls uint64
+		var bucketInfo []uint64
+
+		if _, err := redis.Scan(details, nil, &totalCalls, nil, &bucketInfo); err != nil {
+			break
```
LarssonOliver, line 71: I'm not sure if this break can ever be reached unless there is a breaking change in Redis

exporter/latency.go
oliver006: Oh interesting, didn't know that. In that case this makes sense, leave it as is (could possible add a short comment).

exporter/latency.go
oliver006: Ok, cool, thanks for explaining.

exporter/exporter.go
oliver006: `commands_latencies_usec` - I like that

exporter/latency.go
oliver006: That's fine.

oliver006 (commented):
I think the only thing left to do is to rename the metric and then this is good to go - nice work!

exporter/latency.go
```diff
@@ -30,9 +43,66 @@ func (e *Exporter) extractLatencyMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+	if err != nil {
+		logHistogramErrOnce.Do(func() {
+			log.Errorf("WARNING, LOGGED ONCE ONLY: cmd LATENCY HISTOGRAM, err: %s", err)
+		})
+		log.Debugf("cmd LATENCY HISTOGRAM, err: %s", err)
+		return
+	}
+
+	for i := 0; i < len(reply); i += 2 {
+		cmd := string(reply[i].([]byte))
```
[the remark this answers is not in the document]
LarssonOliver: There sure is, good catch!

LarssonOliver:
I believe everything should be sorted now, thanks for your time and good review comments! :)

PR #778, opened 2023-03-14
Add namespace support to the dashboard

Siebjee:
This change supports the selection of the namespace within the grafana dashboard. This allows to selectively check redis metrics per namespace in case you have multiples running within the same kubernetes cluster.


oliver006:
Thanks!

Issue #727, opened 2022-12-14
Add support for -redis.password-file=sample-pwd-file.json hot-reload 


leesirc:

leesirc:
When sample-pwd-file.json changes, redis_exporter must restart to load the new passwords

oliver006:
Agree, this would be helpful to have. Alas, I'm short of time right now but can help with reviewing and merging a PR.

leesirc:
I'm sorry,  I don't have the ability to solve this problem

oliver006:
Note to self: [this package](https://github.com/fsnotify/fsnotify) could be useful for implementing this.

oliver006:
Ability to reload the config was added in [1.50.0](https://github.com/oliver006/redis_exporter/releases/tag/v1.50.0)

PR #783, opened 2023-04-03
feat: Add redis.password-file hot reload via `/-/reload`

dartagnanli:
1. add redis.password-file hot reload via `/-/reload`#727
2. fix a typo which log `level=error msg="Loaded 3 entries from ./redis-pwd-file.json"`
3. change `redis.addr` default "",explain this,when I use `redis.password-file` most time ,I don not need a `redis.addr`,so this will make continuously outputted logs `level=error msg="Couldn't connect to redis instance (redis://localhost:6379)"` disappear

### DIFF
--- a/exporter/http.go
+++ b/exporter/http.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	log "github.com/sirupsen/logrus"
 )
 
 func (e *Exporter) ServeHTTP(w http.ResponseWriter, r *http.Request) {
@@ -87,4 +88,20 @@ func (e *Exporter) scrapeHandler(w http.ResponseWriter, r *http.Request) {
 	promhttp.HandlerFor(
 		registry, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError},
 	).ServeHTTP(w, r)
+}
+
+func (e *Exporter) ReloadPwdFile(w http.ResponseWriter, r *http.Request) {
+	passwordMap := make(map[string]string, 8)
+	if e.options.RedisPwdFile != "" {
+		var err error
+		passwordMap, err = LoadPwdFile(e.options.RedisPwdFile)
+		log.Debugf("Reload redisPwdFile")
+		if err != nil {
+			log.Errorf("Error reloading redis passwords from file %s, err: %s", e.options.RedisPwdFile, err)
+			http.Error(w, "failed to reload passwords file: "+err.Error(), http.StatusInternalServerError)
+			return
+		}
+	}
+	e.options.PasswordMap = passwordMap
+	_, _ = w.Write([]byte(`ok`))
 }
--- a/exporter/http.go
+++ b/exporter/http.go
@@ -95,8 +95,8 @@ func (e *Exporter) reloadPwdFile(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "There is no pwd file specified", http.StatusBadRequest)
 		return
 	}
-	passwordMap, err := LoadPwdFile(e.options.RedisPwdFile)
 	log.Debugf("Reload redisPwdFile")
+	passwordMap, err := LoadPwdFile(e.options.RedisPwdFile)
 	if err != nil {
 		log.Errorf("Error reloading redis passwords from file %s, err: %s", e.options.RedisPwdFile, err)
 		http.Error(w, "failed to reload passwords file: "+err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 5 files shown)


exporter/pwd_file.go
```diff
@@ -23,7 +23,7 @@ func LoadPwdFile(passwordFile string) (map[string]string, error) {
 		return nil, err
 	}
 
-	log.Errorf("Loaded %d entries from %s", len(res), passwordFile)
+	log.Infof("Loaded %d entries from %s", len(res), passwordFile)
```
oliver006, line 26: 👍

main.go
```diff
@@ -54,7 +54,7 @@ func getEnvInt64(key string, defaultVal int64) int64 {
 
 func main() {
 	var (
-		redisAddr            = flag.String("redis.addr", getEnv("REDIS_ADDR", "redis://localhost:6379"), "Address of the Redis instance to scrape")
```
oliver006, line 57: Please leave this as is, this might break things for people as they didn't have to set this in the path to pick up a local redis instance.

exporter/http.go
```diff
@@ -88,3 +89,19 @@ func (e *Exporter) scrapeHandler(w http.ResponseWriter, r *http.Request) {
 		registry, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError},
 	).ServeHTTP(w, r)
 }
+
+func (e *Exporter) ReloadPwdFile(w http.ResponseWriter, r *http.Request) {
+	passwordMap := make(map[string]string, 8)
+	if e.options.RedisPwdFile != "" {
```
oliver006, line 95: Should we return an error if pwd file is not set? E.g. a 400?

exporter/http.go
```diff
@@ -88,3 +89,19 @@ func (e *Exporter) scrapeHandler(w http.ResponseWriter, r *http.Request) {
...
+		var err error
+		passwordMap, err = LoadPwdFile(e.options.RedisPwdFile)
+		log.Debugf("Reload redisPwdFile")
+		if err != nil {
+			log.Errorf("Error reloading redis passwords from file %s, err: %s", e.options.RedisPwdFile, err)
+			http.Error(w, "failed to reload passwords file: "+err.Error(), http.StatusInternalServerError)
+			return
+		}
+	}
+	e.options.PasswordMap = passwordMap
```
oliver006, line 105: Move this into the if statement. And do we need a mutex around this now?

oliver006 (commented):
Thanks for the PR!

Changes look pretty good, I left a few small comments.

oliver006:
Also, this needs some tests to check that the reload endpoint works.

dartagnanli:
I have optimized my code,please check it

exporter/http_test.go
```diff
@@ -288,6 +288,56 @@ func TestHttpHandlers(t *testing.T) {
 	}
 }
 
+func TestReloadHandlers(t *testing.T) {
+	// if os.Getenv("TEST_PWD_REDIS_URI") == "" {
```
oliver006, line 292: Can you uncomment this plz?

exporter/http.go
```diff
@@ -88,3 +89,21 @@ func (e *Exporter) scrapeHandler(w http.ResponseWriter, r *http.Request) {
...
 	).ServeHTTP(w, r)
 }
+
+func (e *Exporter) reloadPwdFile(w http.ResponseWriter, r *http.Request) {
+	if e.options.RedisPwdFile == "" {
+		http.Error(w, "There is no pwd file specified", http.StatusBadRequest)
+		return
+	}
+	passwordMap, err := LoadPwdFile(e.options.RedisPwdFile)
+	log.Debugf("Reload redisPwdFile")
```
oliver006, line 99: can you please move this up one line, before we call LoadPwdFile() ?

exporter/http_test.go
```diff
@@ -288,6 +288,56 @@ func TestHttpHandlers(t *testing.T) {
...
+	} {
+		t.Run(fmt.Sprintf("path: %s", tst.path), func(t *testing.T) {
+			body := downloadURL(t, ts.URL+tst.path)
+			if !strings.Contains(body, tst.want) {
+				t.Fatalf(`error, expected string "%s" in body, got body: \n\n%s`, tst.want, body)
+			}
+		})
+	}
+
+	eWithnoPwdfile, _ := NewRedisExporter(os.Getenv("TEST_PWD_REDIS_URI"), Options{Namespace: "test", Registry: prometheus.NewRegistry()})
```
oliver006, line 318: You can cover the third case by using [sample-pwd-file.json-malformed](https://github.com/oliver006/redis_exporter/blob/master/contrib/sample-pwd-file.json-malformed) as the password file and then check that you get the expected error message when it fails to load passwords from that file.

oliver006 (approved):

Issue #793, opened 2023-05-01
Automatic CI workflows for release


arthurzam:
I would like to contribute multiple CI workflows for the repo. Some of them are needed for easier packaging for Gentoo distribution, and some I think would be nice for you. Before I start to work on it, I wanted to ask if there are things you don't need or are against. Just to clarify, I'll work on it :)

- adding sdist to release which includes vendored modules (so equivalent of git checkout + go mod vendor + tar). This one would be very useful for Gentoo and other distros since it enables to just compile without fetching all the dependencies using net.
- compiling the binaries for multiple platforms on release and attaching them
- on release, build the updated docker and upload to docker hub (would require you to add the secrets in your repo, I won't have access to them, and this workflow would require you to "approve" it before it starts).

All of them would trigger automatically upon you tagging a release in git.

oliver006:
Thanks @arthurzam - All of the above would be great and I'm intersted to move away from Drone if possible. Their free hosted solution has problems every so often and it blocks releases.
A while back I started #737 for Github actions but didn't have time to finish it.

It would take a bit of work to get the tests to run cleanly on master and PR branch pushes.
The packaging should be pretty straight-forward but in my experience, the devil is always in the details and it takes a good bit of fiddling to get it to work.
IMO, the most important thing is that nothing changes in terms of output so that we don't break existing automation. Filenames, file structure of zips, etc all would need to stay exactly the same.

arthurzam:
I got it. I'll do my best to be exactly backward compatible in the release artifacts. I guess I'll split it into different PRs when I work on it, so you will have easier time verifying it 😄 

I can't give exact timeline, but I think in the current week you will be able to see first PRs.

oliver006:
Thank you very very much, your help is very much appreciated!
And let em know if you need my help, e.g. for creating secrets.

oliver006:
Closing, migrating the test, build and release flow to GH Actions in #737 (and a follow-up PR)

PR #797, opened 2023-05-09
ci: add sdist with vendor for release workflow

arthurzam:
OK, this is the first part for https://github.com/oliver006/redis_exporter/issues/793, the simplest one. This job will create a useful for distributions source tarball, which contains the code & `go mod vendor` run inside.

### Required setup in configuration:

1. Go to repository settings.
2. In left sidebar choose "Environments"
3. Click on add new environment
4. name it `release` (this is the name I used here, you can choose another one and I'll modify it here)
5. I recommend to setup `Required reviewers` and add yourself in there. When you create a new tag, you'll receive an email where you need to approve the run. It helps with accidental tag push before everything is pushed :)
6. Click on "Save protection rules"

### Workflow explanation:

1. Triggered automatically (after approval) when you push tags that start with `v`, for example `v1.51.0`.
2. checkouts and setups go1.19
3. runs `go mod vendor`
4. create a `.tar.gz` file containing a directory named `redis_exporter-v${version}`, under which all content is found including `vendor/` dir.
5. Create a new release on github, adds as asset the sdist, marks the release as draft. This is done so you can once again verify it, add release notes (for example changelog) and then manually publish.

I've tested this workflow on my fork, it generated this file [redis_exporter-0.0.9999.tar.gz](https://github.com/oliver006/redis_exporter/files/11433662/redis_exporter-0.0.9999.tar.gz) (yes, I used quite weird version)

### My next step

Add another job to this workflow, which would build the binaries, and add them as assets to the same draft release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,47 @@
+name: Release
+
+on:
+  push:
+    tags: [tst-v*, v*]
+
+jobs:
+  sdist:
+    runs-on: ubuntu-latest
+    environment: release
+
+    permissions:
+      contents: write # Used to authenticate github release publish
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: '1.22'
+
+    - name: Setup Dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libarchive-tools # for bsdtar
+
+    - name: Generate sdist
+      run: |
+        VERSION=${GITHUB_REF_NAME#v}
+        go mod vendor
+        tar -czf ../redis_exporter-${VERSION}.tar.gz --transform "s,^,redis_exporter-${VERSION}/," *
+        tar -cJf ../redis_exporter-${VERSION}.tar.xz --transform "s,^,redis_exporter-${VERSION}/," *
+        bsdtar -c -a -f ../redis_exporter-${VERSION}.zip -s ",^,redis_exporter-${VERSION}/," *
+
+    - name: Create GitHub release
+      uses: softprops/action-gh-release@v2
+      with:
+        files: |
+          ../redis_exporter-*.tar.gz
+          ../redis_exporter-*.tar.xz
+          ../redis_exporter-*.zip
+        fail_on_unmatched_files: true
+        draft: true


oliver006:
Thank you for the PR! ❤️ 
I'm really slammed this week but will hopefully get to reviewing it, probably by the end of this week or next, sorry about the delay.

arthurzam:
ping?

oliver006:
> ping?

I'm so sorry, I totally messed up and didn't get to reviewing this PR.
I'll try to find some time over the next 1-2 weeks to look at this and review it.
Again, my apologies.

arthurzam:
ping?

oliver006:
My sincere apologies, I never got to reviewing this and should have updated this PR sooner.
I'm still super swamped with work and other stuff but I'll get this reviewed as I think it'd be great to remove the reliance on the old drone system (no matter how well it works right now, it'll break eventually and won't be fixable).

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.19'
```
oliver006, line 24: Bump this to 1.22

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Generate sdist
+      run: |
+        VERSION=${GITHUB_REF_NAME#v}
+        go mod vendor
```
oliver006, line 29: Let's not include vendored dependencies

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: Release
+
+on:
+  push:
+    tags: [v*]
```
oliver006, line 5: Can you make this `[tst-v*]` so a) we can play around with it and test it and b) it won't interfere with the current release process?

oliver006 (commented):
Looks good!
Few minor changes and then step 1 should be ready to go.

Thanks for your patience with this!

oliver006:
I created the `release` env per your instructions, thanks for adding that much detail, that helped.

I got a question re: `Triggered automatically ... when you push tags that start with v...`  and what the new release workflow looks like.

Right now I create a new release by clicking "Draft a new release" on the releases page, pick a tag and title, and then fill out the details, done.
This new flow now requires me to tag via git CLI and push the tag or is there a way to create just a tag (without a release) in the Github UI? I'm totally fine with using the CLI, it's more about knowing what the future release flow looks like.

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.19'
+
+    - name: Generate sdist
+      run: |
+        VERSION=${GITHUB_REF_NAME#v}
+        go mod vendor
+        tar czf redis_exporter-${VERSION}.tar.gz --transform "s,^,redis_exporter-${VERSION}/," *
```
oliver006, line 30: Can you also add the equivalent for creating a zip file and attach it to the release? 
In the end this needs to be identical to the current process and that includes a tar.gz and a zip file in the release.

image

.github/workflows/release.yml
arthurzam: I disagree here. We do want to add dependencies as vendored, since it is part of what downstream (like Gentoo) need to perform offline source build. This was my main goal with this PR, an automated process to add vendored dependencies to the release tarball.

.github/workflows/release.yml
arthurzam: done

.github/workflows/release.yml
arthurzam: done

.github/workflows/release.yml
arthurzam: done, and I've decided to also add `tar.xz`, which uses much better compression, for specific downstreams which wants smallest source tarball.

arthurzam:
> I created the `release` env per your instructions, thanks for adding that much detail, that helped.

Great, thanks.

> I got a question re: `Triggered automatically ... when you push tags that start with v...` and what the new release workflow looks like.
> 
> Right now I create a new release by clicking "Draft a new release" on the releases page, pick a tag and title, and then fill out the details, done. This new flow now requires me to tag via git CLI and push the tag or is there a way to create just a tag (without a release) in the Github UI? I'm totally fine with using the CLI, it's more about knowing what the future release flow looks like.

Yes, it would be better to use a git CLI tool to push those tags. You would be also be able to GPG sign the release, which is a great way to give confidence for consumers in the release was indeed by you.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
+name: Release
+
+on:
+  push:
+    tags: [tst-v*, v*]
```
oliver006, line 5: Please only `tst-v*` for now, I don't want this flow to interfere (or break) the real release flow. We'll cut this over when we have confirmed that this is working and ready for prime time.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+        sudo apt-get update
+        sudo apt-get install -y libarchive-tools # for bsdtar
+
+    - name: Generate sdist
+      run: |
+        VERSION=${GITHUB_REF_NAME#v}
+        go mod vendor
+        tar -czf ../redis_exporter-${VERSION}.tar.gz --transform "s,^,redis_exporter-${VERSION}/," *
+        tar -cJf ../redis_exporter-${VERSION}.tar.xz --transform "s,^,redis_exporter-${VERSION}/," *
+        bsdtar -c -a -f ../redis_exporter-${VERSION}.zip -s ",^,redis_exporter-${VERSION}/," *
```
oliver006, line 37: 👍

.github/workflows/release.yml
oliver006: Ok, I see.
I like the idea of including the dependencies as it gives a full snapshot of the release at tjhe time it was cut.
On the other hand "nobody else is doing this" and I think it's cause nobody else has the "offline build" requirement that Gentoo has.

Let's include the vendored deps for now and if it causes problems we can revisit.

Issue #67, opened 2017-05-14
redis slow log metrics


Colstuwjx:
As redis has [slow log metrics](https://redis.io/commands/slowlog
) to show the slowest CMD, we should export these to prom. Thus, grafana could show the topK slow CMD, and makes redis ops feel happy.

oliver006:
@Colstuwjx - what do you think would be useful to export for the slow log?

I could see the number of slow log entries being useful (e.g. export `SLOWLOG LEN`) but the slow log contents itself seems not really suitable for exporting.

Thoughts?

Colstuwjx:
Yes, I agree that. Instead collecting the detail slow log CMD as "metrics" into prometheus, I'd like to scrape them into Elastic.

Thanks a lot for your reply.

mazzy89:
i'd need a slow queries view in Grafana maybe showing the last 10 slow query. My idea would be to have a metric that would look like this:

```
redis_slowlog_info{command='', id='', time='', exeution=''}
```

where the values are the result of the `SLOWLOG` query properly parsed.

wouldn't be hard to add this to the exporter. What do you think @oliver006 ?

oliver006:
I think that's something for log collection, not for metrics and therefor out of scope for the exporter.

siwyd:
FYI, you can use Filebeat to export slow logs: https://www.elastic.co/blog/brewing-in-beats-read-redis-slow-logs-with-filebeat

madhur:
I think slow queries logs are really needed to be exported for prometheus. Atleast the top 10

PR #807, opened 2023-06-10
Added metrics to display the last 10 slow log commands name , command duration and the execution time

alagusivakumar:


oliver006:
Thanks for the PR, I'll try to get to reviewing it soon!

exporter/slowlog.go
```diff
@@ -30,3 +32,36 @@ func (e *Exporter) extractSlowLogMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+			}
+			commandname, err := redis.Values(commandinfo, err)
+			if err != nil {
+				return
+			}
+			//fmt.Println(string(commandname[0].([]uint8)))
+			command := string(commandname[0].([]uint8))
+
+			e.registerConstMetricGauge(ch, "slowlog_history_last_ten", commandDurationSeconds, commandExecutedTimeStamp, command)
+
```
oliver006, line 63: remove blank line

exporter/slowlog.go
```diff
@@ -30,3 +32,36 @@ func (e *Exporter) extractSlowLogMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+		}
+
```
oliver006, line 65: same

exporter/slowlog.go
```diff
@@ -30,3 +32,36 @@ func (e *Exporter) extractSlowLogMetrics(ch chan<- prometheus.Metric, c redis.Co
...
+	valuesArr, err := redis.Values(doRedisCmd(c, "SLOWLOG", "GET", "10"))
+	var commandDurationSeconds float64
+
+	if err != nil {
+		return
+	}
+
+	for i := 0; i < len(valuesArr); i++ {
+		if values, err := redis.Values(valuesArr[i], err); err == nil && len(values) > 2 {
```
oliver006, line 46: This is not safe. You only check `len(values) > 2` but in line 50 you access the fourth element.

oliver006 (commented):
Thanks for the PR!
This can be a useful feature but should be behind a command-line option that's off by default. Putting the commands in a label can 1) drastically blow up the cardinality of metrics (why the timestamp? I don't think that's a good idea) and 2) can be undesired from a privacy/security perspective.

Also, this needs some tests (see slowlog_test.go for inspiration)

exporter/slowlog.go
alagusivakumar: Hi @oliver006  , I have removed the blank lines and i have updated the condition on the line 46.

alagusivakumar:
> Thanks for the PR! This can be a useful feature but should be behind a command-line option that's off by default. Putting the commands in a label can 1) drastically blow up the cardinality of metrics (why the timestamp? I don't think that's a good idea) and 2) can be undesired from a privacy/security perspective.
> 
> Also, this needs some tests (see slowlog_test.go for inspiration)

Please find the response to your quries
 1) Putting the commands in a label can 1) drastically blow up the cardinality of metrics (why the timestamp? I don't think that's a good idea)
    Commands in the label will help the user to identify which commands caused the slowlogs. In the existing code itself , commands are in the labels for the metrics like redis_commands_total , redis_commands_duration_seconds_total. It is similar to that

    In the current exporter code , you are exposing the metrics of the last slowlog. But we donot know when the last slow log has been recorded. For example , consider a scenario where  we have set the retention policy for metrics to 2 days (we can able to see the metrics only for the past 2days) .In that case , If the last slow log  occurred before 2 days ,   we cannot able to pinpoint the time when the last slow has been recorded . In other words , we cannot the able to find the time of the last slow log execution , if the time period between (lastslowlog time and the current time)   is greater than the metrics retention time.

 Also , i am aware that slow log contents itself seems not really suitable for exporting. but some clients are having usecase for that . Example: https://github.com/oliver006/redis_exporter/issues/67 . It will be very helpful in the case where we want to observe the redis cpu spike and the slowlog details in the same place.

oliver006:
> Commands in the label will help the user to identify which commands caused the slowlogs. In the existing code itself , commands are in the labels for the metrics like ....

That's different though. Those commands only record the general command like `GET` but not the parameters that were used so there's a big difference in potential cardinality.

The tests are failing on go formatting - can you run `go fmt ./...` please?

oliver006:
I thought about this a bit more and I came to the conclusion this doesn't belong in the exporter, this belongs into a log collection system (loki or whatever you use for logs).
Prometheus is mainly for exporting and collecting metrics so I'd suggest you use something different to collect logs like the slowlog.

patsevanton:
Bad. This is good feature :(

PR #810, opened 2023-06-13
check redis config values, and skip non string replies.

sergelogvinov:
Hello,

I've got the error - `Redis CONFIG err: redigo: unexpected element type for Strings, got type []interface {}`
It can be noticed only in `-debug` mode.

KeyDB returns `tls-allowlist` array value in CONFIG GET request.
And redis_exporter silently does not set any `redis_config_$key` metrics, which affect the prometheus alerts (they do not work). 

Can we skip non-string keys in config-metrics?

Thank you.

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -481,14 +481,23 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
 	ch <- e.targetScrapeRequestErrors
 }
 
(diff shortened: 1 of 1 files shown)


sergelogvinov:
> This looks good - thanks for the PR.
> Do you think there's value in logging the errors in extractConfigMetrics() ?
> If not then this is good to go.

Good idea, i've added the error log.

oliver006 (approved):

PR #812, opened 2023-06-17
Use go 1.20 in `go.mod`

Ma27:
This seems to work fine considering that it's already used for all build containers. Also, when building 1.51 on NixOS, I get the following build-error:

    vendor/golang.org/x/sys/unix/syscall.go:83:16: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
    vendor/golang.org/x/sys/unix/syscall_linux.go:1018:20: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
    vendor/golang.org/x/sys/unix/syscall_linux.go:2289:9: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
    vendor/golang.org/x/sys/unix/syscall_unix.go:118:7: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
    vendor/golang.org/x/sys/unix/sysvshm_unix.go:33:7: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)

Upgrading the directive and running `go mod tidy` solves the problem.


oliver006:
Can you rebase you PR please?

Also, are there any downsides to upgrading the go version in go.mod? 
I think it shouldn't affect what Go toolchain is used (that should only change in 1.21, I *think*)

Ma27:
> Can you rebase you PR please?

@oliver006 Done!

> Also, are there any downsides to upgrading the go version in go.mod?

None that I'm aware of. We can probably try out 1.19 as well, but any older version is not supported by Go anyways, so it shouldn't be a big deal: https://endoflife.date/go

>  I think it shouldn't affect what Go toolchain is used (that should only change in 1.21, I think)

The version is only enforced at 1.21, but not before. However it appears to imply some behavioral changes: https://go.dev/doc/modules/gomod-ref#go-notes

Also, the value is passed to the `-lang` option of the compiler (as implied by the error I posted above) and that causes the problem when `unsafe.Slice` is attempted to be used (as done by `golang.org/x/sys` here).

oliver006:
Thanks @Ma27  - make sense.
Let's do it.

Issue #835, opened 2023-09-21
Using password file with isCluster enabled trigger: ERRO[10984] Cluster refresh failed: redisc: all nodes failed


chetratep:
**Describe the problem**
When using password file with isCluster enabled, cluster client is created without password when receiving incoming scrape request.
When using password file with url scheme `redis://` is required. However in `connectToRedisCluster` https://github.com/oliver006/redis_exporter/blob/master/exporter/redis.go#L69-L81  removed the URL scheme before creating RedisDialOptions. The removal of the scheme in the url make searching password from the file hit nothing (https://github.com/oliver006/redis_exporter/blob/master/exporter/redis.go#L35). Because in the password file the URL(Keys) has scheme included.

Error is in below:
```
DEBU[10984] Creating cluster object
DEBU[10984] Running refresh on cluster object
ERRO[10984] Cluster refresh failed: redisc: all nodes failed
NOAUTH Authentication required.
```
 

**What version of redis_exporter are you running?**
Please run `redis_exporter --version` if you're not sure what version you're running.

[ ] 1.54.0

**Running the exporter**
./redis_exporter -redis.password-file=pwd-file.json --is-cluster  -redis.addr redis://xxxxxx:6379

**Expected behavior**
When creating the RedisDialOptions, the URI should include the scheme as well. 

**Screenshots**

**Additional context**
Note: The workaround is to specify the -redis.password=xxxx .

oliver006:
Can't you just change the entry in the password file to omit the `redis://` scheme?

PR #836, opened 2023-09-21
Fix cluster refresh failed  when using pasword file with isCluster mode

chetratep:
This PR is to fix the bug when using password file with `isCluster` mode.

This issue #835

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -74,7 +74,7 @@ services:
     pull: if-not-exists
     ports: [ 7000, 7001, 7002, 7003, 7004, 7005 ]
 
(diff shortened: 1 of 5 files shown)


oliver006:
Thanks for the PR, I'll try to review it in the next few days!

oliver006:
I think this makes sense.

Can you please add a test to confirm the expected behavior is working ?

chetratep:
Sure, I'll find sometime later to do it.

oliver006:
Quick bump @chetratep

oliver006:
Thanks for adding the tests @chetratep !

Issue #784, opened 2023-04-05
New Histogram Collection Creates Errors on Each Scrape With Older Redis Versions


Calebjh:
**Describe the problem**
When running redis 6.x, the new histogram metrics cause Redis error counter to increase on every scrape, even though the error message logged is limited to one.
```
time="2023-04-05T18:32:08Z" level=info msg="Redis Metrics Exporter v1.49.0    build date:     sha1: b8555085b299a28fd67642302072c2cf77afbce8    Go: go1.20.2    GOOS: linux    GOARCH: amd64"
time="2023-04-05T18:32:08Z" level=info msg="Providing metrics at :9121/metrics"
time="2023-04-05T18:32:32Z" level=error msg="WARNING, LOGGED ONCE ONLY: cmd LATENCY HISTOGRAM, err: ERR Unknown subcommand or wrong number of arguments for 'HISTOGRAM'. Try LATENCY HELP."
```
It would be convenient to have a way to disable the new histogram via ENV or automatically if the command is failing.

**What version of redis_exporter are you running?**
Please run `redis_exporter --version` if you're not sure what version you're running.
[ ] 0.3x.x
[x] 1.48+

**Running the exporter**
What's the full command you're using to run the exporter? (please remove passwords and other sensitive data)

**Expected behavior**
Testing an upgrade from an older version to the newest versions. Expected things to continue working, with new metrics available.

**Screenshots**
Periods where error counter is increasing are running v1.49, and when it is stable are v1.47
<img width="1250" alt="Screen Shot 2023-04-05 at 11 53 20 AM" src="https://user-images.githubusercontent.com/10749034/230177624-654ff36d-b570-48a6-85ee-d7e9c18e4f3e.png">

**Additional context**

oliver006:
Thank for opening this issue.
This indeed looks wrong. I think there are multiple ways to handle this
1. disable checking for the latency histogram after thr first error. tricky cause now we have to keep state in the exporter across scrapes and the way the `/scrape` endpoint works won't work any longer
2. start making the exporter version-aware of the redis instance it's scraping and only take certain actions if we know they're supported. rather clean but adds complexity and possibly a lot of "if version > x.y" code
3. add a flag to disable accessing the latency histogram - easiest to implement but user needs to take action to configure the exporter right

I'm leaning towards 3. - what do you think?

Calebjh:
[...]

PR #847, opened 2023-11-03
chore: only enable histograms for Redis 7+

grafanalf:
relates-to: https://github.com/oliver006/redis_exporter/issues/784


oliver006:
Thanks for the PR!

I think I mentioned in #784 that I prefer to not make the exporter version-aware, esp if it's not needed to enable certain functionality.
The log-line should show up only once - is it that big of an issue?

PR #857, opened 2023-12-12
feat: add redis sentinel master config

opan:
Expose couple new metrics to show current redis sentinel master config, such as `quorum`, `parallel-syncs`, `failover-timeou`, and `down-after-milliseconds`. As of now, I have only expose these 4 settings. We can add more if needed. 

Thank you


oliver006 (approved):
👍 

Thanks for the PR!

Is it worth adding a few tests for this to check the metrics show up?

opan:
Sure, I can add some test for this.

PR #858, opened 2023-12-12
Add support for loading TLS system certificate store

stanhu:
Previously a server CA had to be specified for TLS work, even if the system store trusted the cert already.


oliver006 (approved):
Thanks for the PR!

PR #860, opened 2023-12-14
fix scrapping for redis < 4.0

jcaspes:
This fix avoid an exception to bee thrown so scrapping fail for redis version < 4.0
Metrics will so only contains keys_group_count and not memory_usage 
Tested on docker redis 3.2.4: (docker pull r900/redis:3.2.4)

### DIFF
--- a/exporter/key_groups.go
+++ b/exporter/key_groups.go
@@ -173,7 +173,10 @@ for i=3,#ARGV do
   end
 end
 for i,key in ipairs(batch[2]) do
-  usage = redis.call("MEMORY", "USAGE", key)
+  local reply = redis.pcall("MEMORY", "USAGE", key)  
+  if reply['err'] == nil then
+    usage = reply;
+  end
   group = nil
   for i=3,#ARGV do
     key_match_result = {string.find(key, ARGV[i])}
--- a/exporter/key_groups.go
+++ b/exporter/key_groups.go
@@ -174,7 +174,7 @@ for i=3,#ARGV do
 end
 for i,key in ipairs(batch[2]) do
   local reply = redis.pcall("MEMORY", "USAGE", key)  
-  if reply['err'] == nil then
+  if type(reply) == "number" then
     usage = reply;
   end
   group = nil


oliver006 (approved):

exporter/key_groups.go
```diff
@@ -173,7 +173,10 @@ for i=3,#ARGV do
   end
 end
 for i,key in ipairs(batch[2]) do
-  usage = redis.call("MEMORY", "USAGE", key)
+  local reply = redis.pcall("MEMORY", "USAGE", key)
```
oliver006, line 176: This doesn't seem to work, see the tests:

```
=== RUN   TestKeyGroupMetrics/synchronous_with_unclassified_keys

time="2023-12-14T11:01:58Z" level=error msg="ERR Error running script (call to f_2111fe4ef69857c0b23529937bb692b02c475a9b): @user_script:20: user_script:20: attempt to index local 'reply' (a number value)"

    key_groups_test.go:159: Key group count metrics are not expected:

         Expected: map[string]int{"key_exp":5, "key_paul":1, "key_ringo":1, "unclassified":6}

        Actual: map[string]int{}

    key_groups_test.go:167: Key group memory usage metrics are not expected:

         Expected: map[string]bool{"key_exp":true, "key_paul":true, "key_ringo":true, "unclassified":true}

        Actual: map[string]bool{}

    key_groups_test.go:171: Unexpected number of distinct key groups, expected: 4, actual: 0

=== RUN   TestKeyGroupMetrics/synchronous_with_overflow_keys

time="2023-12-14T11:01:58Z" level=error msg="ERR Error running script (call to f_2111fe4ef69857c0b23529937bb692b02c475a9b): @user_script:20: user_script:20: attempt to index local 'reply' (a number value)"

    key_groups_test.go:159: Key group count metrics are not expected:

         Expected: map[string]int{"overflow":12, "test-stream":1}

        Actual: map[string]int{}

    key_groups_test.go:167: Key group memory usage metrics are not expected:

         Expected: map[string]bool{"overflow":true, "test-stream":true}

        Actual: map[string]bool{}

    key_groups_test.go:171: Unexpected number of distinct key groups, expected: 13, actual: 0

--- FAIL: TestKeyGroupMetrics (0.11s)
```

jcaspes: oups, my bads i haven't tested on a version that was supporting MEMORY comand.
It should be better now by checking the returned type
1bc1b4584b122297aadb8ceb270376537f2fbe58

oliver006: Looks good now, thanks!

exporter/key_groups.go
```diff
@@ -173,7 +173,10 @@ for i=3,#ARGV do
   end
 end
 for i,key in ipairs(batch[2]) do
-  usage = redis.call("MEMORY", "USAGE", key)
+  local reply = redis.pcall("MEMORY", "USAGE", key)  
+  if type(reply) == "number" then
+    usage = reply;
+  end
```
oliver006, line 179: Question, should you exit if you don't get a value for usage? 
How will that get handled down in line 193/195 ?

jcaspes: Hello,

local usage variable is initialized to 0 so no problem on line 193/195, is value will stay to 0 if MEMORY do not exists

The idea is to obtain at less key_groups count on redis version prior to 4.0, groups will always contain 0 for memory usage but the correct count.

A possibility to make better should be to not generate memory usage metrics when not available... but it's more complex and need more code modifications (but it is possible to do it ;-) ).

The current workaround make possible to use a limited key_groups scrapping that i think is better than the today error that generate nothing for key_group.

oliver006: That makes sense, thanks for the explanation!

Issue #863, opened 2023-12-22
--export-client-list option doesn't work when IPv6 addresses are returned


edward-milkey:
**Describe the problem**
First, thanks for the great work here!

When `--export-client-list` option is enabled, `redis_connected_clients_details` metrics are not provided by the exporter as expected when Redis `CLIENT INFO` returns IPv6 addresses. This is when using v6.0.14 of Redis.

I encounter IPv6 addresses returned from `CLIENT INFO` when using Azure Redis Cache service with private endpoints.

**What version of redis_exporter are you running?**
v1.55.0

**Running the exporter**
What's the full command you're using to run the exporter? (please remove passwords and other sensitive data)
`--redis.addr rediss://MY_REDIS:6380 --redis.password "REDACTED" --debug --export-client-list --export-client-port`

**Expected behavior**
What metrics are missing? What metrics are wrong? Is something missing that was present in an earlier version? 

The exporter returns other metrics, but the client list metrics (`redis_connected_clients_details`) are not present.

**Screenshots**

**Additional context**

oliver006:
Released: https://github.com/oliver006/redis_exporter/releases/tag/v1.56.0

PR #864, opened 2023-12-22
update client info parsing to accept IPv6 addr

edward-milkey:
This PR addresses issue #863. `parseClientListString` is updated to parse IPv6 addresses.

### DIFF
--- a/exporter/clients.go
+++ b/exporter/clients.go
@@ -32,7 +32,7 @@ type ClientInfo struct {
 	id=40253233 addr=fd40:1481:21:dbe0:7021:300:a03:1a06:44426 fd=19 name= age=782 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=26 qbuf-free=32742 argv-mem=10 obl=0 oll=0 omem=0 tot-mem=61466 ow=0 owmem=0 events=r cmd=client user=default lib-name=redis-py lib-ver=5.0.1 numops=9
 */
 func parseClientListString(clientInfo string) (*ClientInfo, bool) {
-	if matched, _ := regexp.MatchString(`^id=\d+ addr=\w+`, clientInfo); !matched {
+	if matched, _ := regexp.MatchString(`^id=\d+ addr=\S+`, clientInfo); !matched {
 		return nil, false
 	}
 	connectedClient := ClientInfo{}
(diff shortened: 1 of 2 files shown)


exporter/clients.go
```diff
@@ -71,11 +72,12 @@ func parseClientListString(clientInfo string) (*ClientInfo, bool) {
...
 		case "addr":
 			hostPortString := strings.Split(vPart[1], ":")
-			if len(hostPortString) != 2 {
+			if len(hostPortString) < 2 {
+				log.Debug("Invalid value for 'addr' found in client info")
 				return nil, false
 			}
-			connectedClient.Host = hostPortString[0]
-			connectedClient.Port = hostPortString[1]
+			connectedClient.Host = strings.Join(hostPortString[:len(hostPortString)-1], ":")
```
oliver006, line 79: How's that working with IPv4? In that case we just have one item in the "Join()" and no ":" will be added?

exporter/clients.go
```diff
@@ -29,9 +29,10 @@ type ClientInfo struct {
 	Valid Examples
 	id=11 addr=127.0.0.1:63508 fd=8 name= age=6321 idle=6320 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0 events=r cmd=setex user=default resp=2
 	id=14 addr=127.0.0.1:64958 fd=9 name= age=5 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=26 qbuf-free=32742 obl=0 oll=0 omem=0 events=r cmd=client user=default resp=3
+	id=40253233 addr=fd40:1481:21:dbe0:7021:300:a03:1a06:44426 fd=19 name= age=782 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=26 qbuf-free=32742 argv-mem=10 obl=0 oll=0 omem=0 tot-mem=61466 ow=0 owmem=0 events=r cmd=client user=default lib-name=redis-py lib-ver=5.0.1 numops=9
 */
 func parseClientListString(clientInfo string) (*ClientInfo, bool) {
-	if matched, _ := regexp.MatchString(`^id=\d+ addr=\d+`, clientInfo); !matched {
+	if matched, _ := regexp.MatchString(`^id=\d+ addr=\w+`, clientInfo); !matched {
```
oliver006, line 35: nit: is it worth changing this from \d or \w to "not a (white)space character"?

oliver006 (approved):
Thanks for the PR!
Two minor comments/questions but this looks good.

exporter/clients.go
edward-milkey: Yes, I like your suggestion better. Updated.

exporter/clients.go
edward-milkey: Hi @oliver006 - thanks for the quick review!

Yes, that's correct. When there's 1 item in `Join` the join character is not added.

exporter/clients.go
oliver006: Looks good, thanks for updating!

PR #866, opened 2024-01-02
Update metrics that missed labels

opan:
Update for PR #857


opan:
This update is required. Without the labels, the metrics won't show up in `/metrics`. I have tested this locally. I will try to add some tests for this.

oliver006:
Thanks.
PR looks good but this time let's add tests to make sure the metrics really get recorded.

opan:
I have try to add the test in  [here](https://github.com/oliver006/redis_exporter/pull/866/commits/863ca08c5a90941ff726ead9843fc83713df8bf8).

If this does not suffice, I will try to add more test regarding this.

oliver006:
Ok, thanks, let me have another look.

contrib/docker-compose-for-tests.yml
```diff
@@ -32,24 +32,28 @@ services:
       - "16390:6390"
 
   redis-2-8:
+    platform: linux/amd64
```
oliver006, line 35: What are these all about?

oliver006 (approved):
PR looks good - just one Q about the docker-compose changes.

contrib/docker-compose-for-tests.yml
oliver006: Can we revert this? let's keep the docker-compose file platform-agnostic.

contrib/docker-compose-for-tests.yml
```diff
@@ -32,24 +32,28 @@ services:
...
     image: redis:2.8
     command: "redis-server --port 6381 --dbfilename dump2-8.rdb"
     ports:
       - "16381:6381"
 
   keydb-01:
+    platform: linux/amd64
```
oliver006, line 42: Revert these plz

oliver006 (changes requested):

contrib/docker-compose-for-tests.yml
opan: Ah yes, I am running the test on M1 Apple silicone machine. There are some warnings shown and that's why I put it there. Sure, I will remove it

oliver006: Thanks!

PR #867, opened 2024-01-05
Add graceful shutdown

wanghaowei0107:
#868

### DIFF
--- a/main.go
+++ b/main.go
@@ -235,7 +235,7 @@ func main() {
 		}
 	}()
 
-	// Set up signal handling for graceful shutdown
+	// graceful shutdown
 	quit := make(chan os.Signal, 1)
 	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
 	_quit := <-quit
--- a/main.go
+++ b/main.go
@@ -229,11 +229,7 @@ func main() {
 
 	s := make(chan os.Signal, 1)
 	signal.Notify(s, syscall.SIGINT, syscall.SIGTERM)
-	go func() {
-		select {
-		case _s := <-s:
-			log.Infof("Receive exit signal, bye!    signal: %s", _s.String())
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -224,4 +226,10 @@ func main() {
 	} else {
 		log.Fatal(http.ListenAndServe(*listenAddress, exp))
 	}
+
+	s := make(chan os.Signal, 1)
+	signal.Notify(s, syscall.SIGINT, syscall.SIGTERM)
+	_sig := <-s
```
oliver006, line 232: Thanks for the PR!

How does this fix the issue? 
The listen call in line 225/227 should block so I don't think we ever get here.

oliver006:
Thanks for looking into this!
How does this fix the issue? 
The listen call in line 225/227 should block so I don't think we ever down to where you catch the signal.

wanghaowei0107:
@oliver006 Hello! Thanks for your reply! I rewrote the code, but the workflow is pending for a long time. Can you help me find out what's going on?

wanghaowei0107:
test

oliver006:
Thanks for updating this PR, this looks a lot better. 
Did you have a chance to test it? Does it work?

wanghaowei0107:
Yes,I did test in local.It's work.
By the way, Can you tell me why drone/pr check is always pending on first step? I don't know how to pass this check. Thank you.
微信图片_20240115101136

oliver006 (approved):
Sorry for the delay.
This looks good - please one last change for the message that is printed and then we should be good to go.

oliver006 (approved):

wanghaowei0107:
So when will this feature be released?

oliver006:
Released as https://github.com/oliver006/redis_exporter/releases/tag/v1.57.0

PR #883, opened 2024-02-17
update sample contrib dashboard with non-Angular widgets

oliver006:


stefanandres:
Oh, that is great. I was actually waiting for that and I realize nobody updated the mixin yet (that we use for alerts and the dashboard): https://github.com/oliver006/redis_exporter/tree/master/contrib/redis-mixin

PR #890, opened 2024-03-12
Add password key option

taoso:
The Redis does not support auth by username and password until version 6.0. So in legacy version instances, administrator can only set one password.

The current version of redis_exporter support read password from the path specified by the password-file option according to the URI of redis connection to collect metrics from multiple targets.

But this design requires the administrator list all target URIs into the password file. Every time new instance created, the password file has to be updated. Besides, having many different instances dose not necessarily require all of them use different passwords. They may be separated into several groups, and instances of the same group may use identical password.

In this scenario, it will be more convenient to introduce the concept of password key. When the collecting job fetch the probe API, it can transport one additional key for password:

```
http://10.0.0.1:9121/scrape?target=foo:6379&passkey=dev
http://10.0.0.1:9121/scrape?target=bar:6379&passkey=dev
```

Both the instances of foo:6379 and bar:6379 could use the same password, and we can only add one line into the password file:

```
dev:dev-pass
```

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -78,6 +78,7 @@ type Options struct {
 	RedisMetricsOnly          bool
 	PingOnConnect             bool
 	RedisPwdFile              string
+	RedisPwdKey               string
 	Registry                  *prometheus.Registry
 	BuildInfo                 BuildInfo
 }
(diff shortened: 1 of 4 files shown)


oliver006:
Thanks for the PR!
I'm slammed right now but hopefully will get to this in the next few days and weeks.
Please ping me if you don't hear back from me.

oliver006:
I finally had time to take a look.

I'm not sure I fully understand what pain point you're trying to address and what scenario would be improved by adding this.

Can you elaborate more?

taoso:
@oliver006 thanks for reply

> I'm not sure I fully understand what pain point you're trying to address and what scenario would be improved by adding this.

The redis_exporter support using different password for different instance by the REDIS_PASSWORD_FILE option.
However, you need to hard code all instance URI into the password file, statically.

In my scenario, I need to use one redis_exporter to export metrics many instances dynamically. There may be several projects and different may use different instances. I can not let all redis use the same passwords. And if I use the current password file, I have to update them every time there is new instance created.

So I propose to introduce the password key, which is something like password group. By this means, all the redis instances of the same password key (group) share the same password. When the collector query their metrics, it need set both the target and password key. So there is no need to update the password file every time there is new instance created.

oliver006:
> I can not let all redis use the same passwords

That makes sense.

However, the situation you're describing can be solved by automation that handles the deployment of Redis instances and the updating of the password file.

I understand that "password groups" offer a convenient half-way path towards that but I prefer to not add any unnecessary tooling to the exporter that can be solved with other tooling.

sfdb:
For what it's worth I was looking forward to this PR being merged as I was running into the same issue. You're right that it could be solved by automating it, but it's something that most people with multiple Redis clusters will run into, so it would be good to tackle the painpoint once in a central place.

Any reconsideration would be appreciated, thanks regardless for your work on this repo.

PR #899, opened 2024-04-14
ValKey changes

oliver006:


exporter/streams_test.go
```diff
@@ -397,7 +397,7 @@ func TestStreamsScanStreamGroupsConsumers(t *testing.T) {
 	if os.Getenv("TEST_REDIS_URI") == "" {
 		t.Skipf("TEST_REDIS_URI not set - skipping")
 	}
-	addr := os.Getenv("TEST_REDIS7_URI")
+	addr := os.Getenv("TEST_VALKEY7_URIl")
```
bjosv, line 400: Drive by; spotted the reason for the CI failure, but you might have already seen it :)
```suggestion
	addr := os.Getenv("TEST_VALKEY7_URI")
```

PR #900, opened 2024-04-16
Support Docker Compose V2 in Makefile

bjosv:
Since Docker Compose is installed as a plugin in v2 the binary `docker-compose` is not always available.
The makefile will now use `docker compose` in that case.

The [recommendation](https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose) is to start using the command `docker compose` instead of `docker-compose`, so an alternative is to replace, but this gives somewhat smoother migration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .DEFAULT_GOAL := build
 
+DOCKER_COMPOSE := $(if $(shell which docker-compose),docker-compose,docker compose)
+
 .PHONY: build
 build:
 	go build .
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,7 @@
 .DEFAULT_GOAL := build
 
+DOCKER_COMPOSE := $(if $(shell which docker-compose),docker-compose,docker compose)
```
oliver006, line 3: 👍

oliver006 (approved):

oliver006:
> The [recommendation](https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose) is to start using the command `docker compose` instead of `docker-compose`

This is neat, I didn't know that!

PR #906, opened 2024-05-18
Feature: Add option to exclude consumer from Stream metrics

nantiferov:
Tested it manually, looks like working properly, PTAL.

Probably need to add more tests, can check it later.

Amount of metrics on test redis:
* without streams metrics collect - 260
* with `REDIS_EXPORTER_CHECK_STREAMS=*` - 680
* with this feature enabled in addition to previous - 430

Fixes: #902

PR #911, opened 2024-05-27
Feature: Add option to skip collecting latency histogram metrics

Nklya:
I will try to add tests later, PTAL in terms of naming and style.

It fixes: #784 via 3rd option you mentioned here https://github.com/oliver006/redis_exporter/issues/784#issuecomment-1500739568


oliver006:
Approach looks good.
Naming nit: in this PR #906 we used `exclude-...-metrics`, here it's `skip-...-metrics` - let's use `exclude` for both?

PR #913, opened 2024-05-30
Feature: Add option to skip collecting latency histogram metric

Nklya:
It fixes: #784 via 3rd option you mentioned here https://github.com/oliver006/redis_exporter/issues/784#issuecomment-1500739568

Accidentally closed previous PR https://github.com/oliver006/redis_exporter/pull/911 due to mess-up with force pushes.
Tested it locally, works as expected.
Can check adding tests later on weekend if needed.


oliver006 (approved):
Thanks for the PR, looks good. You can add the test in a follow-up PR if you like, not urgent. Let me merge this and release a new version

Nklya:
Thank you 🙂

PR #916, opened 2024-06-17
Update Dockerfile to test image with sentinel

Akansha-Nigam:


docker/Dockerfile
```diff
@@ -11,6 +11,7 @@ ARG SHA1="[no-sha]"
 ARG TAG="[no-tag]"
 ARG GOARCH
 
+RUN apk add --no-cache bash
```
oliver006, line 14: What's this about?

oliver006 (commented):
My apologies - I totally missed this PR, getting to it only now.

Can you explain why this change is needed?

docker/Dockerfile
Akansha-Nigam: While enabling sentinel for redis, it expects the exporter to have bash with it.
So, to make it compatible with redis bitnami sentinel, raised this.

PR #917, opened 2024-07-01
fix: compatible with stream entry id metric for redis version 6.x/5.x.

gitsang:
The entry id in different versions of redis is not always in the 17th and 19th positions of the `xinfo stream`.

In 6.x and 5.x, `first-entry` is 11th, `last-entry` is 13th

### DIFF
--- a/exporter/streams.go
+++ b/exporter/streams.go
@@ -52,8 +52,18 @@ func getStreamInfo(c redis.Conn, key string) (*streamInfo, error) {
 	}
 
 	// Extract first and last id from slice
-	stream.FirstEntryId = getStreamEntryId(values, 17)
-	stream.LastEntryId = getStreamEntryId(values, 19)
+	for idx, v := range values {
+		vbytes, ok := v.([]byte)
+		if !ok {
+			continue
+		}
+		if strings.EqualFold(string(vbytes), "first-entry") {
+			stream.FirstEntryId = getStreamEntryId(values, idx+1)
+		}
+		if strings.EqualFold(string(vbytes), "last-entry") {
+			stream.LastEntryId = getStreamEntryId(values, idx+1)
+		}
+	}
 
 	stream.StreamGroupsInfo, err = scanStreamGroups(c, key)
 	if err != nil {
--- a/exporter/streams.go
+++ b/exporter/streams.go
@@ -57,10 +57,10 @@ func getStreamInfo(c redis.Conn, key string) (*streamInfo, error) {
 		if !ok {
 			continue
 		}
-		if strings.EqualFold(string(vbytes), "first-entry") {
+		if string(vbytes) == "first-entry" {
 			stream.FirstEntryId = getStreamEntryId(values, idx+1)
 		}
-		if strings.EqualFold(string(vbytes), "last-entry") {
+		if string(vbytes) == "last-entry" {
 			stream.LastEntryId = getStreamEntryId(values, idx+1)
 		}
 	}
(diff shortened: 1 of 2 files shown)


exporter/streams_test.go
```diff
@@ -88,11 +88,11 @@ func TestStreamsGetStreamInfo(t *testing.T) {
 			if isNotTestTimestamp(info.LastGeneratedId) {
 				t.Errorf("Stream LastGeneratedId mismatch.\nActual: %#v;\nExpected any of: %#v", info.LastGeneratedId, TestStreamTimestamps)
 			}
-			if info.FirstEntryId != "" {
-				t.Errorf("Stream FirstEntryId mismatch.\nActual: %#v; - Expected empty", info.FirstEntryId)
+			if isNotTestTimestamp(info.FirstEntryId) {
```
oliver006, line 91: Why does this need to change?

exporter/streams.go
```diff
@@ -65,7 +75,9 @@ func getStreamInfo(c redis.Conn, key string) (*streamInfo, error) {
 }
 
 func getStreamEntryId(redisValue []interface{}, index int) string {
-	if len(redisValue) < index || redisValue[index] == nil || len(redisValue[index].([]interface{})) < 2 {
+	if _, ok := redisValue[index].([]interface{}); !ok ||
```
oliver006, line 78: This is getting a little crowded. 

I *think* this one is a direct prerequisite so that the last condition `len(redisValue[index].([]interface{}))` will not fail, right? Can we group them together?

oliver006 (commented):
Thanks for the PR, change looks good, just two minor questions.

exporter/streams.go
gitsang: You means `if _, ok := redisValue[index].([]interface{}); !ok || len(redisValue) < index || redisValue[index] == nil || len(redisValue[index].([]interface{})) < 2` ?

exporter/streams_test.go
gitsang: `exporter/exporter_test.go +110` Add `TestStreamTimestamps` to the stream
Since I have changed the method of getting the first and last entry id from a slice, when redis > v5.0, `getStreamInfo` should be able to get non-empty entry ids. 

https://redis.io/docs/latest/commands/xinfo-stream/

exporter/streams.go
gitsang: Sorry, I mistakenly used `strings.Contains` to determine `entry-id`, so it matched `max-deleted-entry-id`, which value cannot be asserted as `[]interface{}`.

gitsang:
Already rebase the code and commit:
- Change `strings.Contains` to `strings.EqualFold`
- `TestStreamsGetStreamInfo`: Change judgment for `XxxEntryId` from `isNotTestTimestamp` to the same accurate judgment as `TestStreamsGetStreamInfoUsingValKey7`.

exporter/streams.go
oliver006: I like the second grouping, let's do that. And agreed, we should try to prevent the panic either way.

exporter/streams.go
```diff
@@ -52,8 +52,18 @@ func getStreamInfo(c redis.Conn, key string) (*streamInfo, error) {
...
 
 	// Extract first and last id from slice
-	stream.FirstEntryId = getStreamEntryId(values, 17)
-	stream.LastEntryId = getStreamEntryId(values, 19)
+	for idx, v := range values {
+		vbytes, ok := v.([]byte)
+		if !ok {
+			continue
+		}
+		if strings.EqualFold(string(vbytes), "first-entry") {
```
oliver006, line 60: why not simple `=` equality?

gitsang: I habitually used EqualFold because sometimes there are issues with case conversion. But indeed, you are right, here we can directly use `=`.

Has been modified.

exporter/streams.go
gitsang: Has been modified.

oliver006 (approved):
Thanks for the PR!

PR #925, opened 2024-07-21
fix: stop return metrics when parse float value err

wilkice:
if parse value err, means the value is not float. So there may be something bad happened. If we dont't return, the metric value is 0, which may has a different meaning with the real value. From the code above, `0` also means `fail`, `err` and `false`.


oliver006 (approved):
Makes sense - thanks for the PR!

PR #933, opened 2024-08-06
feat: export client-output-buffer-limit

### LINKED ISSUE #932: Export `client-output-buffer-limits` as a metric
**Describe the problem**

In redis-exporter [v1.60.0](https://github.com/oliver006/redis_exporter/releases/tag/v1.60.0), a new metric is exported for tracking the total replication buffer size: `mem_total_replication_buffers`

Given that this metric increases up to the defined replication buffer limit in `client-output-buffer-limit`, I think it makes sense to export that into a metric too so that we can do some alerting on when `mem_total_replication_buffers` reaches a threshold of `client-output-buffer-limit`.

I tried out using `--include-config-metrics`, but that exports the config in an unusable format:

```
redis_config_key_value{key="client-output-buffer-limit",value="normal 0 0 0 slave 1610612736 1610612736 0 pubsub 33554432 8388608 60"} 1
```

This could be in some new metrics:

```
redis_config_client_output_buffer_limit{class="normal", limit="hard"} 0
redis_config_client_output_buffer_limit{class="slave", limit="hard"} 1610612736
redis_config_client_output_buffer_limit{class="pubsub", limit="hard"} 33554432
```

I'm not sure on the best way of presenting the soft limits, and the soft seconds. Keen to hear other opinions before raising a PR.

Config file docs: https://redis.io/docs/latest/operate/oss_and_stack/management/config-file/

After above is implemented, we could have an alert on below:

```
(mem_total_replication_buffers / redis_config_client_output_buffer_limit{class="slave", limit="hard"}) > 0.7
```

[x] 1.6.2

**Running the exporter**
N/A

**Screenshots**
N/A

**Additional context**
N/A

jdheyburn:
Fixes #932

Exports configured `client-output-buffer-limit` as a metric. Combined with `mem_total_replication_buffers` it will allow you to alert on when replication buffers are reaching the limit.

```
(mem_total_replication_buffers / config_client_output_buffer_limit_bytes{class="slave", limit="hard"}) > 0.7
```

Happy for suggestions on alternative metric name and labelling tactic.

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -353,6 +353,8 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 		"commands_total":                                  {txt: `Total number of calls per command`, lbls: []string{"cmd"}},
 		"commands_latencies_usec":                         {txt: `A histogram of latencies per command`, lbls: []string{"cmd"}},
 		"latency_percentiles_usec":                        {txt: `A summary of latency percentile distribution per command`, lbls: []string{"cmd"}},
+		"config_client_output_buffer_limit_bytes":         {txt: `The configured buffer limits per class`, lbls: []string{"class", "limit"}},
+		"config_client_output_buffer_limit_overcome_seconds": {txt: `How long for buffer limits per class to be exceeded before replicas are dropped`, lbls: []string{"class", "limit"}},
 		"config_key_value":                                {txt: `Config key and value`, lbls: []string{"key", "value"}},
 		"config_value":                                    {txt: `Config key and value as metric`, lbls: []string{"key"}},
 		"connected_slave_lag_seconds":                     {txt: "Lag of connected slave", lbls: []string{"slave_ip", "slave_port", "slave_state"}},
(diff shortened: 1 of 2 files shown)


oliver006 (approved):
Thanks for the PR - looks good!

nantiferov:
@jdheyburn thank you for PR, I added `mem_total_replication_buffers` in 1.60, but haven't thought about collecting buffer limits.

P.S. Tested it quickly now, I think query for alert should look like this because of labels mismatch on left and right of `/`:

```promql
redis_mem_total_replication_buffers_bytes / sum(redis_config_client_output_buffer_limit_bytes{class="slave", limit="hard"}) without (class, limit)
```

nantiferov:
@oliver006 I had similar case with sysctl values and node_exporter and they have similar functionality to config metrics, but it's possible to put numeric values into metric value and also generate metric name based on value labels.
image

https://github.com/prometheus/node_exporter/tree/master?tab=readme-ov-file#sysctl-collector

I think you might consider this as an option if someone will have time to improve config collector.

PR #940, opened 2024-08-31
Support docker buildx for multiple platforms building.

xiaocj:
e.g.

```
docker buildx build \
  --push \
  --platform linux/amd64,linux/arm64 \
  -f ./docker/Dockerfile \
  -t oliver006/redis_exporter:v1.6.2 \
  ./
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,26 +1,27 @@
-ARG GOARCH
 #
 # build container
 #
-FROM --platform=linux/amd64 golang:1.22-alpine as builder
+FROM --platform=$BUILDPLATFORM golang:1.22-alpine AS builder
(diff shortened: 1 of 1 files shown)


oliver006:
Can you explain what this is about?

xiaocj:
I think it's a more official way to build multi-platform combinations by docker. https://docs.docker.com/build/building/multi-platform/ 

It also support the traditional single platform build command:

```
docker build -f ./docker/Dockerfile -t oliver006/redis_exporter:v1.6.2 ./
```

oliver006:
Ok, thanks for explaining.

The current setup is working and I don't really see a benefit in moving to the official way other than potentially breaking the existing, working setup (esp as the test docker build step is broken right now).

Let's leave this as is as it works.

PR #944, opened 2024-09-06
Support exporting `configured_hz` from info server

justinmir:
When the `dynamic-hz` configuration option is in-use (Redis 6+), Redis will include `configured_hz` in the output of `info server`. The `hz` that is currently configured can impact Redis CPU utilization.

This introduces exporting the server's current `configured_hz` value.


oliver006:
Looks good.
`configured_hz` - question: what unit is that in? Can we add it to the metric name? Also, should you expose `hz` as well or is that already there?

justinmir:
> configured_hz - question: what unit is that in? Can we add it to the metric name? Also, should you expose hz as well or is that already there?

The unit in this case is actually just `hz` which is the unit of frequency: `1 hz == once per second`.

oliver006:
PR looks good!

Can you please rebase and pull in the new changes that run the tests in Github Actions?

PR #950, opened 2024-09-09
Re-implement connected_client to have deterministic cardinality

mindw:
The current connected client suffers from some drawbacks rendering much less useful than it should:
1. unbound cardinality - the labels "idle_since",  "omem" and "cmd" values are unpredictable and there for unbound.
2. Having values as labels made using them in queries all but impossible, getting the idlest client for example.
3. The `id` field which uniquely identifies the client is missing from the labels.

This PR attempts to ratify all that:
* renames `connected_clients_details` to `connected_client_info` (as per prometheus naming conventions) - dropping all high cardinality labels.
* adds `connected_client_output_buffer_memory_usage_bytes`, `connected_client_total_memory_consumed`, `connected_client_created_at_timestamp`, `connected_client_idle_since_timestamp`, `connected_client_channel_subscriptions_count`, `connected_client_pattern_matching_subscriptions_count`, `connected_client_shard_channel_subscriptions_count` and `connected_client_shard_channel_watched_keys` metrics per client.
* add `id` to all metrics.

~I'm still running some tests on the output metrics, so please don't merge it yet :)~

Input, missing bits (docs?) are welcome!


hiroshi-ishii:
Quick comments (warning: I do not know any go):

1. Did you intend to remove the parsing of `Cmd` entirely (i.e. not just removing it as a label)?
2. Can you also parse/output the following: `qbuf`, `qbuf-free`, `obl`, `oll`
3. `port` could be an `int64`; it may be useful to do numerical comparisons on it
5. Typo on several lines (e.g. 63): `cloud not parse ...`; not yours but might as well fix it
6. line 102: error message for `watch` case instead prints `... 'sub' field ...`
7. Can we replace `connected_client_info` with `flags` as a label (there are many flags that can change moment-to-moment), with `connected_client_flags` as `flags` as the gauge value? Not sure about this one.
8. Don't forget to remove the `log.Debug()` in the `ExportClientsInclPort` case
9. Should metric description's descriptions start with caps? i.e. `A connected client's ...`
10. I think `name` should not be part of `clientBaseLabels` since `name` can be changed with `CLIENT SETNAME`. The `name` should be mapped to the `id` by other means.
11. Could we refactor `metricDescriptions` /  `registerConstMetricGauge` calls with an array & loop -- make it shorter and easier to read?

mindw:
Thanks @hiroshi-ishii for the review - see below:
> Quick comments (warning: I do not know any go):
> 
> 1. Did you intend to remove the parsing of `Cmd` entirely (i.e. not just removing it as a label)?
It was - are there leftovers?
> 2. Can you also parse/output the following: `qbuf`, `qbuf-free`, `obl`, `oll`
Yes, I'll add it later.
> 3. `port` could be an `int64`; it may be useful to do numerical comparisons on it
Port is arbitrary, and isn't unique, I don't recall ever it being exposed as a value.
> 4. Typo on several lines (e.g. 63): `cloud not parse ...`; not yours but might as well fix it
Fixed.
> 5. line 102: error message for `watch` case instead prints `... 'sub' field ...`
Fixed.
> 6. Can we replace `connected_client_info` with `flags` as a label (there are many flags that can change moment-to-moment), with `connected_client_flags` as `flags` as the gauge value? Not sure about this one.
Possible - need to look into this.
> 7. Don't forget to remove the `log.Debug()` in the `ExportClientsInclPort` case
Fixed, thanks!
> 8. Should metric description's descriptions start with caps? i.e. `A connected client's ...`
Yes, fixed.
> 9. I think `name` should not be part of `clientBaseLabels` since `name` can be changed with `CLIENT SETNAME`. The `name` should be mapped to the `id` by other means.
That's actually the reason for it to be on the info metric. While ID is unique and name is not, the name is far more human friendly IMO.
> 10. Could we refactor `metricDescriptions` /  `registerConstMetricGauge` calls with an array & loop -- make it shorter and easier to read?
Maybe, from my experience these kind of generics hurt readability and maintainability. In case of exporters it is quite common to avoid them unless there is something smarter behind them (like generating metrics from struct members dynamically to support future changes).

oliver006:
Hey there and thanks for opening the PR!
I haven't looked at the code yet but I like the idea a lot to break out the details into separate metrics that can be queried and graphed. So go ahead with that. 

One request though: can we keep the name of the current client detail metric the same so you don't introduce backwards incompatible changes?

Let me know when you're ready for code review and want comments and then I'll have a look. 

Oh, and one more request: can you rebase the PR please? I moved the tests over to Github Actions yesterday and I don't see that here yet.

mindw:
> Hey there and thanks for opening the PR! I haven't looked at the code yet but I like the idea a lot to break out the details into separate metrics that can be queried and graphed. So go ahead with that.

Tx!

> One request though: can we keep the name of the current client detail metric the same so you don't introduce backwards incompatible changes?
It is an optional feature and I'm not sure how much it was used given it's current state also it wont have the same labels. Then again the old name doesn't really follow the Prometheus naming convention. I would suggest to duplicate it the name instead.
 
> Let me know when you're ready for code review and want comments and then I'll have a look.
It's ready :)
 
> Oh, and one more request: can you rebase the PR please? I moved the tests over to Github Actions yesterday and I don't see that here yet.
I've rebased on the latest and greatest main/master.

Thanks!

hiroshi-ishii:
@mindw
On (2) Thanks!
On (3) I didn't understand your response.
On (6) maybe we could replace `flags` in the labels with a subset of mostly permanent flags (e.g. client is replica, client is master)
On (9), on second thought, you're right, and client name doesn't change that often so we're OK.
On (10) I understand what you're saying but I still think it's worthwhile in this case. But it's your rodeo...

exporter/clients.go
```diff
@@ -103,30 +181,185 @@ func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c
...
+					"a connected client's number of keys it's currently watching",
+					clientBaseLabels,
+				)
+				e.registerConstMetricGauge(
+					ch, "connected_client_shard_channel_watched_keys", float64(info.Ssub),
+					clientBaseLabelsValues...,
+				)
+			}
+
+			e.metricDescriptions["connected_client_query_buffer_length"] = newMetricDescr(
```
hiroshi-ishii, line 321: `qbuf` and `qbuf-free` are both bytes ([Redis source](https://github.com/redis/redis/blob/7.4.0/src/networking.c#L2878)); `sds` is a string buffer with a variable size length field.

exporter/clients.go
```diff
@@ -103,30 +181,185 @@ func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c
...
+				"connected_client_query_buffer_free_space_bytes",
+				"A connected client's free space of the query buffer in bytes (0 means the buffer is full)",
+				clientBaseLabels,
+			)
+			e.registerConstMetricGauge(
+				ch, "connected_client_query_buffer_free_space_bytes", float64(info.QbufFree),
+				clientBaseLabelsValues...,
+			)
+
+			e.metricDescriptions["connected_client_output_buffer_length"] = newMetricDescr(
```
hiroshi-ishii, line 343: `obl` is bytes; `oll` is not

exporter/clients_test.go
```diff
@@ -122,15 +117,15 @@ func TestExportClientList(t *testing.T) {
 
 		found := false
 		for m := range chM {
-			if strings.Contains(m.Desc().String(), "connected_clients_details") {
```
hiroshi-ishii, line 125: I think you may want to revert some of these changes from `connected_clients_details` to `connected_client_info`

exporter/clients_test.go
```diff
@@ -65,26 +60,26 @@ func TestParseClientListString(t *testing.T) {
 		expectedLbls ClientInfo
```
hiroshi-ishii, line 60: Possibly rename `expectedLbls` to `expectedInfo` ?

exporter/clients_test.go
```diff
@@ -102,7 +97,7 @@ func TestParseClientListString(t *testing.T) {
 		}
 
 		if *lbls != tst.expectedLbls {
```
hiroshi-ishii, line 99: Perhaps rename `lbls` to `info` ?

exporter/clients_test.go
```diff
@@ -65,26 +60,26 @@ func TestParseClientListString(t *testing.T) {
 		expectedLbls ClientInfo
 	}{
 		{
```
hiroshi-ishii, line 62: Would be good to add a test case for `oll` and `obl` with non-zero values

hiroshi-ishii (commented):
Looked at the new changes

exporter/clients_test.go
mindw: added tests for both

mindw:
@oliver006 this is ready for your review :)

oliver006:
> @oliver006 this is ready for your review :)

Sorry about the delay - having a look now.

docker-compose.yml
```diff
@@ -20,6 +19,12 @@ services:
     ports:
       - "16379:6379"
 
+  redis74:
+    image: redis:7.4
```
oliver006, line 23: Is there anything that sets redis 7.4 apart, feature-wise, that warrants adding an additional service container for tests *AND* keeping 7.2? 

If not I'd rather bump the existing redis 7.2 to 7.4 and use that.

exporter/clients_test.go
```diff
@@ -120,17 +115,51 @@ func TestExportClientList(t *testing.T) {
 			close(chM)
 		}()
 
-		found := false
+		tsts := []struct {
```
oliver006, line 118: 👍

exporter/clients.go
```diff
@@ -44,32 +55,100 @@ func parseClientListString(clientInfo string) (*ClientInfo, bool) {
...
 			connectedClient.Db = vPart[1]
+		case "sub":
+			Sub, err := strconv.ParseInt(vPart[1], 10, 64)
+			if err != nil {
+				log.Debugf("could not parse 'sub' field(%s): %s", vPart[1], err.Error())
+				return nil, false
+			}
+			connectedClient.Sub = Sub
+		case "psub":
+			Psub, err := strconv.ParseInt(vPart[1], 10, 64)
```
oliver006, line 90: I think we can simplify all these a lot by ignoring the parse error and simply do

`connectedClient.Psub, _ = strconv.ParseInt(vPart[1], 10, 64)`

(and can you maybe use Atoi instead ? is there an explicit need for the uint64?)

Can you just add a little to the tests that the metrics are not only present but have some none-zero value?

exporter/clients.go
```diff
@@ -103,30 +181,185 @@ func (e *Exporter) extractConnectedClientMetrics(ch chan<- prometheus.Metric, c
...
+				"connected_client_info",
+				"Details about a connected client",
+				clientInfoLabels,
+			)
+			e.registerConstMetricGauge(
+				ch, "connected_client_info", 1.0,
+				clientInfoLabelsValues...,
+			)
 
+			// keep the old name for backwards compatability
```
oliver006, line 215: I know I said to keep the old metrics for backwards compatibility but seeing this now makes me reconsider that. It's not going to be fully backwards compatible anyway (labels will have changed) so let's just toss the old metric and make this cleaner.

oliver006 (commented):
Overall approach looks good - just a couple of smaller things

docker-compose.yml
mindw: `watch` was added in Redis 7.4 and in ValKey 8.0. The latter also added `tot-net-in`, `tot-net-out`, `tot-cmds`. It seems ValKey != Redis (at least in versions).
But for `watch` we can use Valkey 8.0.

mindw: Yes, there there is at least one field which is Redis 7.4 specific. There are a lot more info fields - but that's another PR.

ValKey 7.2.5 is the latest ValKey 7.x release and is compatible with Redis 7.2.4 . In fact he just released ValKey 8.0 seems to include some Redis 7.4 changes but its officially compatible with Redis 7.2.4. It seems that Redis & ValKey can no longer be treated the same.

exporter/clients.go
mindw: think we can simplify all these a lot by ignoring the parse error and simply do

connectedClient.Psub, _ = strconv.ParseInt(vPart[1], 10, 64)

(and can you maybe use Atoi instead ? is there an explicit need for the uint64?)
Atoi returns a native int - whose size varies between platforms (32 bit in x86 & arm). Usage of it seems like an avoidable risk. The above was lifted from the previous code.

Can you just add a little to the tests that the metrics are not only present but have some none-zero value?
Yep, added!

oliver006: > Atoi returns a native int - whose size varies between platforms (32 bit in x86 & arm). Usage of it seems like an avoidable risk. The above was lifted from the previous code.

what's the risk? 

I don't see 'ParseInt()` used other than in `durationFieldToTimestamp()` (where it's not really needed IMO) so let's keep it simple and use Atoi

hiroshi-ishii: If I may interject...

The risk is integer overflow. At least for some of the fields (e.g. the memory & buffer size fields), we should use an integer type that guarantees 64-bits, since that's what Redis code uses: https://github.com/redis/redis/blob/7.4.0/src/networking.c#L2878

My understanding is that Atoi casts to an int which only guarantees 32 bits.

oliver006: Fair point @hiroshi-ishii - let's keep the 64 bit ints

docker-compose.yml
oliver006: > There are a lot more info fields - but that's another PR.

Great - looking forward to it!

> It seems that Redis & ValKey can no longer be treated the same.

Yeah, unfortunately.

Do you think it makes sense to bump the dependencies to the latest valkey and Redis versions?

exporter/clients.go
mindw: applied simplified parsing.

docker-compose.yml
mindw: Made a simple test of using valkey 8..0 - test run failed with many errors. It seems extended compatibility mode is needed https://github.com/valkey-io/valkey/pull/306 .
It seems like beyond the scope of this PR?

mindw:
@oliver006 IMO all comments were addressed, anything else?

oliver006:
> @oliver006 IMO all comments were addressed, anything else?

Apologies for the delay, I didn't have much time last week, will get back to this again in the next few days.

docker-compose.yml
oliver006: Agreed. Looks like we need some dedicated effort to make this work with valkey

oliver006 (approved):
Nice work - thanks for getting this done!

Issue #942, opened 2024-09-03
RediSearch module metrics collection


nantiferov:
**Describe the problem**
Hi,

I recently started testing [RediSearch](https://github.com/RediSearch/RediSearch) module for redis and noticed that there is no support for its metrics collection.

So I found these info commands which shows useful information about module stats:

```log
$ redis-cli -p 36379 info modules
# Modules
module:name=RedisCompat,ver=1,api=1,filters=0,usedby=[],using=[],options=[]
module:name=redisgears_2,ver=20020,api=1,filters=0,usedby=[],using=[],options=[]
module:name=search,ver=21005,api=1,filters=0,usedby=[],using=[ReJSON],options=[handle-io-errors]
module:name=ReJSON,ver=20803,api=1,filters=0,usedby=[search],using=[],options=[handle-io-errors]
module:name=timeseries,ver=11202,api=1,filters=0,usedby=[],using=[],options=[]
module:name=bf,ver=20802,api=1,filters=0,usedby=[],using=[],options=[]

# redisgears_2_UninitialisedBackends
redisgears_2_backend_name:js

# search_version
search_version:2.10.5
search_redis_version:7.4.0 - oss

# search_index
search_number_of_indexes:2

# search_fields_statistics
search_fields_text:Text=5,Sortable=1
search_fields_numeric:Numeric=2,Sortable=1
search_fields_tag:Tag=1

# search_memory
search_used_memory_indexes:7052
search_used_memory_indexes_human:0.006725311279296875
search_total_indexing_time:5.9800000190734863

# search_cursors
search_global_idle:0
search_global_total:0

# search_gc
search_bytes_collected:0
search_total_cycles:0
search_total_ms_run:0

# search_dialect_statistics
search_dialect_1:1
search_dialect_2:0
search_dialect_3:0
search_dialect_4:0

# search_runtime_configurations
search_enableGC:ON
search_minimal_term_prefix:2
search_minimal_stem_length:4
search_maximal_prefix_expansions:200
search_query_timeout_ms:500
search_timeout_policy:return
search_cursor_read_size:1000
search_cursor_max_idle_time:300000
search_max_doc_table_size:1000000
search_max_search_results:10000
search_max_aggregate_results:10000
search_gc_scan_size:100
search_min_phonetic_term_length:3
```

I'm thinking would be good to add these metrics collection:
* search_version
* search_number_of_indexes
* search_used_memory_indexes
* search_total_indexing_time
* search_global_idle
[...]

PR #953, opened 2024-09-10
Feature: Add redis search module metrics

nantiferov:
fixes: #942

Found that redis returns all related information to modules with `info modules` and added collection of all metrics that looks useful so far.

Parsing of `module:name=RedisCompat,ver=1,api=1,filters=0,usedby=[],using=[],options=[]` is not ideal, but works fine with current Redis 7.4 /  Redis search module 2.10


exporter/exporter.go
```diff
@@ -347,77 +358,77 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 		txt  string
 		lbls []string
 	}{
-		"commands_duration_seconds_total":                 {txt: `Total amount of time in seconds spent per command`, lbls: []string{"cmd"}},
-		"commands_failed_calls_total":                     {txt: `Total number of errors prior command execution per command`, lbls: []string{"cmd"}},
-		"commands_rejected_calls_total":                   {txt: `Total number of errors within command execution per command`, lbls: []string{"cmd"}},
-		"commands_total":                                  {txt: `Total number of calls per command`, lbls: []string{"cmd"}},
-		"commands_latencies_usec":                         {txt: `A histogram of latencies per command`, lbls: []string{"cmd"}},
-		"latency_percentiles_usec":                        {txt: `A summary of latency percentile distribution per command`, lbls: []string{"cmd"}},
-		"config_client_output_buffer_limit_bytes":         {txt: `The configured buffer limits per class`, lbls: []string{"class", "limit"}},
```
nantiferov, line 356: I haven't changed this part, but looks like `go fmt` wasn't run in for `config_client_output_buffer_limit_overcome_seconds` in https://github.com/oliver006/redis_exporter/pull/933

I think it's worth to add linter check for formatting in CI.

oliver006:
Sorry about the delay - looking at the PR now.

exporter/exporter.go
```diff
@@ -267,6 +268,21 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 			"server_threads":        "server_threads_total",
 			"long_lock_waits":       "long_lock_waits_total",
 			"current_client_thread": "current_client_thread",
+
+			// Redis Modules metrics
+			// RediSearch module
+			"search_number_of_indexes":   "search_number_of_indexes",
+			"search_used_memory_indexes": "search_used_memory_indexes",
```
oliver006, line 275: `search_used_memory_indexes` 

I think this could use the unit this is in. Add `_bytes` ?

exporter/exporter.go
```diff
@@ -350,77 +366,79 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
...
+		"stream_max_deleted_entry_id":                        {txt: `The epoch timestamp (ms) of last message was deleted from the stream`, lbls: []string{"db", "stream"}},
+		"stream_first_entry_id":                              {txt: `The epoch timestamp (ms) of the first message in the stream`, lbls: []string{"db", "stream"}},
+		"stream_last_entry_id":                               {txt: `The epoch timestamp (ms) of the last message in the stream`, lbls: []string{"db", "stream"}},
+		"stream_radix_tree_keys":                             {txt: `Radix tree keys count"`, lbls: []string{"db", "stream"}},
+		"stream_radix_tree_nodes":                            {txt: `Radix tree nodes count`, lbls: []string{"db", "stream"}},
+		"up":                                                 {txt: "Information about the Redis instance"},
+		"module_info":                                        {txt: "Information about loaded Redis module", lbls: []string{"name", "ver", "api", "filters", "usedby", "using"}},
+		"search_version":                                     {txt: "Information about the RediSearch module", lbls: []string{"version"}},
```
oliver006, line 441: can we bake this into the `module_info` metric? Add a `version` label? seems tightly related to the module but I don't know if they keep different versions for the two.

oliver006 (commented):
General approach looks good - just a couple of minor questions

exporter/exporter.go
nantiferov: It's already exists there, just in different format. Here it's `search_version:2.10.5`, but there it's `module:name=search,ver=21005,api=1,filters=0,usedby=[],using=[ReJSON],options=[handle-io-errors]`, i.e. `ver=21005`

In theory we can drop this metric and keep only module_info, but it looks nicer with dots as separators here.

nantiferov: I added also now continue after registering metrics to avoid unneeded parsing of metrics after `registerConstMetricGauge`

exporter/exporter.go
nantiferov: Thanks, fixed

docker-compose.yml
```diff
@@ -83,3 +83,8 @@ services:
     image: tile38/tile38:latest
     ports:
       - "19851:9851"
+
+  redis-stack:
+    image: redis/redis-stack-server:7.4.0-v0
```
nantiferov, line 88: Just in case, `7.4.0-v0` is current version of `redis-stack-server` and it includes redis 7.4 + modules with recent versions.

exporter/exporter.go
oliver006: Yeah, good catch, that wasn't working as expected.
I just pushed https://github.com/oliver006/redis_exporter/commit/827e6b5bfb6bbbee1fac3ae3d3b70da0489b9d1b to master - that should fix it.

exporter/exporter.go
```diff
@@ -267,6 +268,21 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 			"server_threads":        "server_threads_total",
 			"long_lock_waits":       "long_lock_waits_total",
 			"current_client_thread": "current_client_thread",
+
+			// Redis Modules metrics
+			// RediSearch module
+			"search_number_of_indexes":   "search_number_of_indexes",
+			"search_used_memory_indexes": "search_used_memory_indexes_bytes",
```
oliver006, line 275: Thanks for changing this one, what about the others, do they all adhere to prom best practices re: the units? 

For instance: `search_total_ms_run` - if this is in milliseconds then the metric name should be in seconds and end in `_seconds`

exporter/exporter.go
oliver006: `In theory we can drop this metric and keep only module_info, but it looks nicer with dots as separators here.` - in that case, if the versions match, then I'd prefer to just capture one metric - the formatting is just eye candy.
But if the module version could differ from the search engine version then it'd make sense to capture both.

exporter/modules.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		log.Errorf("extractSearchMetrics() err: %s", err)
+		return
+	}
+
+	lines := strings.Split(info, "\r\n")
+	for _, line := range lines {
+		log.Debugf("info: %s", line)
+
+		split := strings.Split(line, ":")
```
oliver006, line 22: can we add a quick check here that `len(split) == 2`?  I'm not sure the switch statement would catch that despite the `case` in line 39

exporter/modules.go
```diff
@@ -0,0 +1,51 @@
...
+		switch {
+		case split[0] == "module":
+			module := strings.Split(split[1], ",")
```
oliver006, line 25: same, can we wrap this with a len() check?

`if module := split(); len(module) > ...` should do the trick

exporter/exporter.go
nantiferov: > But if the module version could differ from the search engine version then it'd make sense to capture both.

Nope, that afaik shouldn't happen, removed search_version and let's keep version in module_info as is.

exporter/modules.go
nantiferov: Added. Though I thing case would catch that.
With this approach switch/case is not needed, so I replaced it with if.

exporter/modules.go
nantiferov: Done, added.

exporter/exporter.go
nantiferov: That's good question.
I don't know much details about metrics Search module provides and how to handle them.

On example of search_total_ms_run, how I can modify it in exporter from ms to sec?

exporter/exporter.go
oliver006: 👍

exporter/exporter.go
oliver006: Ah, good point, that would require extracting the metric manually and then divide by 1000 to get to seconds.
Let's keep it simple and not do that.

exporter/modules.go
oliver006: 👍

exporter/modules.go
oliver006: 👍

oliver006 (approved):
Nice work, looks good!

Will merge later today or tomorrow and cut a new release.

exporter/exporter.go
nantiferov: Actually, just re-checked adding metrics and I think it might make sense to add units to these 3 more, so I commit changes. I cannot convert ms to sec, but I think it's worth to move `ms` to the end.

```
"search_total_indexing_time": "search_total_indexing_time_ms",
"search_bytes_collected":     "search_collected_bytes",
"search_total_ms_run":        "search_total_run_ms",
```

exporter/exporter.go
```diff
@@ -267,6 +268,21 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
...
+			"search_total_indexing_time": "search_total_indexing_time_ms",
+			"search_global_idle":         "search_global_idle",
+			"search_global_total":        "search_global_total",
+			"search_bytes_collected":     "search_collected_bytes",
+			"search_total_cycles":        "search_total_cycles",
+			"search_total_ms_run":        "search_total_run_ms",
+			"search_dialect_1":           "search_dialect_1",
+			"search_dialect_2":           "search_dialect_2",
+			"search_dialect_3":           "search_dialect_3",
+			"search_dialect_4":           "search_dialect_4",
```
nantiferov, line 285: One more thing - I haven't added metrics collection configuration section because it's kinda static. It could be changed via module settings, but they're not changing by itself.

I'll revisit this later when have more statistics on this module usage.

```
# search_runtime_configurations
search_enableGC:ON
search_minimal_term_prefix:2
search_minimal_stem_length:4
search_maximal_prefix_expansions:200
search_query_timeout_ms:500
search_timeout_policy:return
search_cursor_read_size:1000
search_cursor_max_idle_time:300000
search_max_doc_table_size:1000000
search_max_search_results:1000000
search_max_aggregate_results:-1
search_gc_scan_size:100
search_min_phonetic_term_length:3
```

oliver006:
Apologies for the delay - I was busy this past week. I'll get back to this in the next few days.

exporter/exporter.go
oliver006: No, I think this is fine.

exporter/exporter.go
oliver006: Yeah, we can safe that for later.

exporter/modules.go
```diff
@@ -0,0 +1,52 @@
...
+
+		split := strings.Split(line, ":")
+		if len(split) != 2 {
+			continue
+		}
+
+		if split[0] == "module" {
+			// module format: 'module:name=search,ver=21005,api=1,filters=0,usedby=[],using=[ReJSON],options=[handle-io-errors]'
+			module := strings.Split(split[1], ",")
+			if len(module) != 7 {
```
oliver006, line 30: I've been thinking about this section a bit. 
I was wondering if we could make this more generic (e.g. make it work for other modules as well), if it's worth it and, if yes, how to do it.

The way this is written now, it's very much working for just the search module (the hardcoded labels, the len() == 7) but the metric is just "module"info" (I know, we talked about that) and that suggests infos for mor than just one module (and that's what could be the case in the future)

I can think of two ways forward.

1. Either make this for just search but then let's call this "module_search_info" or so, check here not only len() ==7 but that the name is "search" 

2. Make this truly generic. Generate labels on the fly by parsing, use metric names for individual modules, etc etc.

What do you think?

nantiferov: Morning,
It's actually already generic and collects all modules (not only search) information that available from `info modules` output.

oliver006: I see. So there are always 7 items in the module info string? (that wasn't clear to me, I kind of expected that each module could have a different number of pieces returned)

In that case we're good here.

nantiferov: No problem.

I'm not sure if it's 100% guaranteed by Redis, but so far looks like this:
```log
# Modules
module:name=RedisCompat,ver=1,api=1,filters=0,usedby=[],using=[],options=[]
module:name=redisgears_2,ver=20020,api=1,filters=0,usedby=[],using=[],options=[]
module:name=search,ver=21005,api=1,filters=0,usedby=[],using=[ReJSON],options=[handle-io-errors]
module:name=ReJSON,ver=20803,api=1,filters=0,usedby=[search],using=[],options=[handle-io-errors]
module:name=timeseries,ver=11202,api=1,filters=0,usedby=[],using=[],options=[]
module:name=bf,ver=20802,api=1,filters=0,usedby=[],using=[],options=[]
```

nantiferov: I clarified comment in the code in [last commit](https://github.com/oliver006/redis_exporter/pull/953/commits/9e7822e716a6dcacf5210039fc9092a85a69d8c9)

oliver006: Thanks!

oliver006 (approved):
Nice work - thanks for your patience and for pushing this all the way to completion!

nantiferov:
Thank you for review.
Would be great to release new version this week if you have time.

oliver006:
Yeah, let me cut a release and get this out, sorry for the delay.

exporter/exporter.go
```diff
@@ -267,6 +268,21 @@ func NewRedisExporter(redisURI string, opts Options) (*Exporter, error) {
 			"server_threads":        "server_threads_total",
 			"long_lock_waits":       "long_lock_waits_total",
 			"current_client_thread": "current_client_thread",
+
+			// Redis Modules metrics
+			// RediSearch module
+			"search_number_of_indexes":   "search_number_of_indexes",
+			"search_used_memory_indexes": "search_used_memory_indexes_bytes",
+			"search_total_indexing_time": "search_total_indexing_time_ms",
```
nantiferov, line 276: `search_total_indexing_time` is apparently a counter, will prepare PR with renaming it to `search_indexing_time_ms_total`

PR #958, opened 2024-10-08
Hotfix: Fix search_total_indexing_time metric type

nantiferov:
According to my tests with real workload, this metric type is counter, I'm renaming it according to Prometheus recommendations and moving to `metricMapCounters` map.


nantiferov:
Not urgent, so far we just have warnings from Grafana that metric name doesn't end with `_total`

image

oliver006 (approved):

oliver006:
Thanks for the quick fix!

oliver006:
https://github.com/oliver006/redis_exporter/releases/tag/v1.64.1

PR #962, opened 2024-10-29
Hotfix: Change type of search_total_cycles/search_total_ms_run metrics

nantiferov:
Initially added in https://github.com/oliver006/redis_exporter/pull/953 as gauges, but usage shows that they're counters.
I'm not sure if there are many users of this feature, but technically metrics rename might break some dashboards/alerts.

Related: #942


nantiferov:
Dashboard with rate:
image


And without:
image

Current state doesn't block anything, but I think it would be correct to define these metrics as counters and fix naming:
image

oliver006 (approved):

oliver006:
> I'm not sure if there are many users of this feature, but technically metrics rename might break some dashboards/alerts.

Yeah, I think this is still early so we can just change it without too much ado. Appreciate the concern though.

oliver006:
https://github.com/oliver006/redis_exporter/releases/tag/v1.66.0

nantiferov:
Thank you for review and release 👍

PR #968, opened 2024-12-25
maxclients extracted info all

liuqian1990:
GCP Redis does not support the config command, making it impossible to retrieve maxclients. It needs to be extracted from info all


oliver006 (approved):
Thanks for the PR!

PR #978, opened 2025-02-12
add aof incr file size

### LINKED ISSUE #977: monitor aof incr file size
**Describe the problem**
Due to the way AOF persistence works, sometimes the `AOFREWRITE` command must be sent to the Redis server manually. Even then, the AOF incr file can grow much before the rewrite is triggered.

It'd be interesting to add a new map metric for the incr file size, which can later trigger alerts when surpassing a threshold. 

The `INFO` command exports the following AOF-related metrics:
- aof_last_cow_size
- aof_current_size
- aof_base_size
- aof_pending_rewrite
- aof_buffer_length
- aof_rewrite_buffer_length

However, none of them exports directly the size for the incr files, only the total sum of the base RDB + all incr files.

The proposal is to add a new metric `aof_incr_file_size` with the label `filename` where it will constantly monitor all AOF incr files from the path retrieved from the redis config:
Base Dir: `CONFIG GET dir`
Appendonly Dir: `CONFIG GET appenddirname`

The limitation is that it will only work if `redis_exporter` has access to the same directories as the redis server.

- [x] v1.67.0

dudizimber:
fix https://github.com/oliver006/redis_exporter/issues/977

Add a new opt-in metric to monitor AOF incr file sizes.


oliver006:
I think this is fundamentally broken as it assumes that the exporter has access to the `appenddirname` directory which is not necessarily true, putting it in the exporter doesn't seem right.
There are probably other ways for you to expose this metric (shell script and node exporter?).

dudizimber:
The permission issue is a valid point, so the metric requires the flag `-include-aof-file-size` to be set for the exporter to run.

Although it's possible to expose the metric with `node_exporter`, bundling it with the rest of Redis's metrics makes more sense. Moreover, exposing it through `node_exporter` is a cumbersome process and requires installation and scraping, which is not always guaranteed. `redis_exporter` is the standard tool for exporting Redis's metrics, so it's safer to assume that it will be used to handle Redis-related metrics.

oliver006:
The exporter currently works exclusively via the network connection that the exporter opens to the valkey/redis instance, this change would break that paradigm. 
As I pointed out above, you can add this independently from the exporter to e.g. the node_exporter that you might be running on your valkey/redis node and gather the data like that.

dudizimber:
@oliver006 can you clarify why https://github.com/oliver006/redis_exporter/pull/1095 was merged when "The exporter currently works exclusively via the network connection that the exporter opens to the valkey/redis instance, this change would break that paradigm."?

Issue #980, opened 2025-02-28
check-keys option is extremely affected by latency


matiasba:
**Describe the problem**
I'm currently running a single exporter in us-east-1 and using it to monitor a dozen AWS managed Redis in a few regions around de world. It works great but I recently added a check-keys config to pull about 50 keys per server and the collection times skyrocketed in all regions except us-east-1. After some testing it seems solely related to latency. Even a relative low increase latency seems to greatly affect scrape time. But it only happens when I add the check-keys option.

For example scraping from us-east-1 to eu-central-1 usually takes less than a second but with 
check-keys: ["db0=mt:*:pending,db0=mt:*:avg_proc_time,db0=mt:*:avg_wait_time"]
It takes about 1.3 minutes and it only needs to add about 24 extra keys 

**What version of redis_exporter are you running?**
v1.67.0

**Running the exporter**
Oficial docker image

**Expected behavior**
Latency should not impact the scrape time so much

oliver006:
Oh interesting, that seems odd that check-keys increases the scrape time so drastically.
Do you have `--debug` output that can shed a bit more light on what's going on?

oliver006:
Ok, I think I was able to reproduce the problem. The exporter does 5(!) network calls for each check-key key, this can easily take 1-2 secs per check-key entry, depending on your latency, which checks out with what you're observing.

`testkey2` is a simple string key, set to `test` like this:

`redis-cli > set testkey2 test`



Example `--debug` output: 
```
DEBU[0277] c.Do() - running command: SELECT [0]
DEBU[0278] c.Do() - done
DEBU[0278] c.Do() - running command: TYPE [testkey2]
DEBU[0278] c.Do() - done
DEBU[0278] c.Do() - running command: PFCOUNT [testkey2]
DEBU[0278] c.Do() - err: WRONGTYPE Key is not a valid HyperLogLog string value.
DEBU[0278] c.Do() - done
DEBU[0278] c.Do() - running command: STRLEN [testkey2]
DEBU[0278] c.Do() - done
DEBU[0278] c.Do() - running command: GET [testkey2]
DEBU[0278] c.Do() - done
```

I think there are a bunch of optimizations that can help here.
1. we don't need `SELECT` if it's the same DB as for the last call. This should be an easy fix.
[...]

PR #981, opened 2025-03-01
improve latency by running "SELECT" only if the db is different

oliver006:
related to #980


oliver006:
Some rudimentary benchmarking with about 10 keys showed a reduction in latency for a full scrape from about 5s to 3s.
There's still room for improvement but that's a lot more work and I don't have bandwidth for that right now.

PR #988, opened 2025-03-25
add ignoreEmptyDatabases for ignore empty database redis_db_keys* metric

ktlcove:
I use redis for dev, single redis instance has 100000 databases, but 99% empty
so i add a IgnoreEmptyDatabases for ignore all empty databases metrics (db_keys, db_keys_expiring)

I am not a developer, i just edit code do what i want and put code here. maybe someone need this

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -84,6 +84,7 @@ type Options struct {
 	BuildInfo                      BuildInfo
 	BasicAuthUsername              string
 	BasicAuthPassword              string
+	IgnoreEmptyDatabases           bool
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR!

oliver006 (commented):
Two things
1. Code change looks alright - can you add a test for this please? 
2. I think we could use a bit better name for the flag. Something along the lines of "don't set key metrics for empty databases". Thoughts?

oliver006:
Closing in favor of #1001

PR #997, opened 2025-04-24
Add all Go runtime metrics

dongjiang1989:
expose all Go runtime metrics like GC stats, memory stats etc.


oliver006:
Thanks for the PR!

Looking at the docs (and at the metrics that the exporter currently exports), I think thee two are already included in the default registry:

From:
https://github.com/prometheus/client_golang/blob/96a5ad6ef56e0d41d3bf06ba0478b34a0c718a92/prometheus/doc.go#L185-L188

```
// Also note that the DefaultRegisterer comes registered with a Collector for Go
// runtime metrics (via NewGoCollector) and a Collector for process metrics (via
// NewProcessCollector). With a custom registry, you are in control and decide
// yourself about the Collectors to register.
```

dongjiang1989:
> Looking at the docs (and at the metrics that the exporter currently exports), I think thee two are already included in the default registry:

Thanks @oliver006 review

Differences:
```
MustRegister(NewGoCollector())
```
change to
```
collectors.NewGoCollector(collectors.WithGoCollectorRuntimeMetrics(collectors.MetricsAll)),
```

ref: https://github.com/prometheus/client_golang/blob/main/prometheus/registry.go#L60-L63

oliver006:
> `MustRegister(collectors.NewGoCollector(collectors.WithGoCollectorRuntimeMetrics(collectors.MetricsAll)))`

Thanks for highlighting the diff.

What does that effectively mean? What metrics are we getting with that ^^^ that we're currently missing out on.

On the one hand I like being explicit about initializing the Registry, on the other hand I like using the DefaultRegistry because it means I'm going with the defaults and if something new is added to the defaults then I'll automagically pull it in when upgrading the client_golang dependency.

Thoughts?

oliver006:
I had another look at this and I think this is fine.

Thanks for the PR and sorry for the delay!

nantiferov:
> So why not go the other way around and add a flag "extended go metrics" (or similar) which defaults to false and that would enable the full set of 209 metrics.

Yep, I think that would the best option.

nantiferov:
Created issue to track it and PR with fix:

- https://github.com/oliver006/redis_exporter/issues/1034
- https://github.com/oliver006/redis_exporter/pull/1035

Issue #1043, opened 2025-10-03
Feature: Add Redis search module indexes metrics collection


nantiferov:
**Describe the problem**

Redis modules generic metrics collection was added in https://github.com/oliver006/redis_exporter/issues/942, now I think it's worth to add per-index metrics collection with optional pattern to match indexes, similar to streams metrics collection now.

Example of index info output (no difference in fields between v7.4 and v8.0):

```log
> ft.info idx:bicycle
 1) index_name
 2) idx:bicycle
 3) index_options
 4) (empty array)
 5) index_definition
 6) 1) key_type
    2) JSON
    3) prefixes
    4) 1) bicycle:
    5) default_score
    6) "1"
    7) indexes_all
    8) false
 7) attributes
 8) 1) 1) identifier
       2) $.brand
       3) attribute
       4) brand
       5) type
       6) TEXT
       7) WEIGHT
       8) "1"
    2) 1) identifier
       2) $.model
       3) attribute
       4) model
       5) type
       6) TEXT
       7) WEIGHT
       8) "1"
    3) 1) identifier
       2) $.description
       3) attribute
       4) description
       5) type
       6) TEXT
       7) WEIGHT
       8) "1"
    4) 1) identifier
       2) $.price
       3) attribute
       4) price
       5) type
       6) NUMERIC
    5) 1) identifier
       2) $.condition
       3) attribute
       4) condition
       5) type
       6) TAG
       7) SEPARATOR
       8) ,
 9) num_docs
10) (integer) 3
11) max_doc_id
12) (integer) 3
13) num_terms
14) (integer) 33
15) num_records
16) (integer) 105
17) inverted_sz_mb
18) "0.003833770751953125"
19) vector_index_sz_mb
20) "0"
21) total_inverted_index_blocks
22) (integer) 35
23) offset_vectors_sz_mb
24) "1.0013580322265625e-4"
25) doc_table_size_mb
26) "0.015542030334472656"
27) sortable_values_size_mb
28) "0"
29) key_table_size_mb
30) "1.4495849609375e-4"
31) tag_overhead_sz_mb
32) "4.57763671875e-5"
33) text_overhead_sz_mb
34) "0.0011014938354492188"
35) total_index_memory_sz_mb
36) "0.021017074584960938"
37) geoshapes_sz_mb
38) "0"
39) records_per_doc_avg
40) "35"
41) bytes_per_record_avg
42) "38.28571319580078"
43) offsets_per_term_avg
44) "1"
[...]

PR #1046, opened 2025-10-07
Feature: Add Redis search module indexes metrics collection

nantiferov:
Fix: #1043


nantiferov:
So far in draft mode as unfinished.

I believe I hit limitations of `ScanStruct()` https://github.com/gomodule/redigo/issues/691

exporter/exporter.go
```diff
@@ -480,6 +482,30 @@ func NewRedisExporter(uri string, opts Options) (*Exporter, error) {
 		"number_of_distinct_key_groups":                      {txt: `Number of distinct key groups`, lbls: []string{"db"}},
 		"script_result":                                      {txt: "Result of the collect script evaluation", lbls: []string{"filename"}},
 		"script_values":                                      {txt: "Values returned by the collect script", lbls: []string{"key", "filename"}},
+		"search_index_num_docs":                              {txt: "Number of documents in search index", lbls: []string{"index_name"}},
```
nantiferov, line 485: It seems this map should be sorted alphabetically, but according to my check, it wasn't fully sorted before I added new metrics. Should I do this?

Second, I added my new metrics in order according to output of `FT.INFO` command. Same as in [exporter/search_indexes.go](https://github.com/oliver006/redis_exporter/pull/1046/files#diff-31b5f69586dd37c686ba61440cee6217e2987092a0ee59d3ce627d6238ecb6f7R15). Should I sort them alphabetically instead?

exporter/exporter.go
```diff
@@ -480,6 +482,30 @@ func NewRedisExporter(uri string, opts Options) (*Exporter, error) {
...
+		"search_index_sortable_values_size_bytes":            {txt: "Memory used by sortable values, used for sorting purposes", lbls: []string{"index_name"}},
+		"search_index_key_table_size_bytes":                  {txt: "Memory used by the key table, stores mapping between document IDs and keys", lbls: []string{"index_name"}},
+		"search_index_tag_overhead_size_bytes":               {txt: "Memory overhead of the TAG TrieMaps", lbls: []string{"index_name"}},
+		"search_index_text_overhead_size_bytes":              {txt: "Memory overhead of the TEXT Tries", lbls: []string{"index_name"}},
+		"search_index_total_index_memory_size_bytes":         {txt: "Total memory used by search index", lbls: []string{"index_name"}},
+		"search_index_geoshapes_size_bytes":                  {txt: "Memory used by GEO-related fields", lbls: []string{"index_name"}},
+		"search_index_records_per_doc_avg":                   {txt: "Average number of records (including deletions) per document", lbls: []string{"index_name"}},
+		"search_index_bytes_per_record_avg":                  {txt: "Average size of each record in bytes", lbls: []string{"index_name"}},
```
nantiferov: I converted other metrics from megabytes, `_mb` to `_bytes` and keep `_bytes` in the end according to Prometheus metrics naming schema.

However, for this metric I think it's worth to keep naming as is, since it's pre-aggregated value from redis and there're 4 metrics in total with similar naming, like `search_index_records_per_doc_avg`

exporter/exporter.go
```diff
@@ -480,6 +482,30 @@ func NewRedisExporter(uri string, opts Options) (*Exporter, error) {
...
+		"search_index_offsets_per_term_avg":                  {txt: "Average number of offsets (position information) per term", lbls: []string{"index_name"}},
+		"search_index_offset_bits_per_record_avg":            {txt: "Average number of bits used for offsets per record", lbls: []string{"index_name"}},
```
nantiferov: I think it's worth to not convert this metric to bytes, since it's pre-aggregated metrics from redis, like `search_index_bytes_per_record_avg`, it shows "Average number of bits used for offsets per record" and it seems it expected to be in bits.

exporter/search_indexes.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"regexp"
+
+	"github.com/gomodule/redigo/redis"
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+)
+
+// All fields of the streamInfo struct must be exported
```
nantiferov: I added this header similar to `exporter/streams.go` file, I guess limitation is the same, since I'm using `redis.ScanStruct` here too.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.23.0
 toolchain go1.24.2
 
 require (
-	github.com/gomodule/redigo v1.9.2
+	github.com/gomodule/redigo v1.9.3
```
nantiferov, line 8: I'm bumping redigo to latest version due to fix in https://github.com/gomodule/redigo/pull/692

exporter/search_indexes_test.go
```diff
@@ -0,0 +1,83 @@
...
+func TestExtractSearchIndexesMetrics(t *testing.T) {
+	if os.Getenv("TEST_REDIS8_URI") == "" {
+		t.Skipf("TEST_REDIS8_URI not set - skipping")
+	}
+
+	tsts := []struct {
+		addr                     string
+		inclSearchIndexesMetrics bool
+		wantSearchIndexesMetrics bool
+	}{
```
nantiferov, line 46: Initially I also added 2 more scenarios for `TEST_REDIS_URI`, like in `exporter/modules_test.go`. But it fails on tests because `FT._LIST` command is not available in that Redis version (since it doesn't have search module), [logs](https://github.com/oliver006/redis_exporter/actions/runs/18356936001/job/52291324604#step:10:22550)

```golang
{addr: os.Getenv("TEST_REDIS_URI"), inclSearchIndexesMetrics: true, wantSearchIndexesMetrics: false},
{addr: os.Getenv("TEST_REDIS_URI"), inclSearchIndexesMetrics: false, wantSearchIndexesMetrics: false},
```

Do you think code should be somehow changed to enable these scenarios?

exporter/search_indexes_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/gomodule/redigo/redis"
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+)
+
+func setupSearchIndex(t *testing.T, addr string) error {
```
nantiferov, line 13: I added helper function, similar to `setupKeys()` in `exporter/exporter_test.go` to create index in empty redis.

Otherwise there's nothing and this test scenario fails because no search_index metrics collected: `{addr: os.Getenv("TEST_REDIS8_URI"), inclSearchIndexesMetrics: true, wantSearchIndexesMetrics: true}`

exporter/search_indexes.go
```diff
@@ -0,0 +1,96 @@
...
+		e.registerConstMetricGauge(ch, "search_index_geoshapes_size_bytes", indexInfo.GeoshapesSizeMb*1048576, indexInfo.IndexName)
+		e.registerConstMetricGauge(ch, "search_index_records_per_doc_avg", indexInfo.RecordsPerDocAvg, indexInfo.IndexName)
+		e.registerConstMetricGauge(ch, "search_index_bytes_per_record_avg", indexInfo.BytesPerRecordAvg, indexInfo.IndexName)
+		e.registerConstMetricGauge(ch, "search_index_offsets_per_term_avg", indexInfo.OffsetsPerTermAvg, indexInfo.IndexName)
+		e.registerConstMetricGauge(ch, "search_index_offset_bits_per_record_avg", indexInfo.OffsetBitsPerRecordAvg, indexInfo.IndexName)
+		e.registerConstMetricGauge(ch, "search_index_indexing", float64(indexInfo.Indexing), indexInfo.IndexName)
+		e.registerConstMetricGauge(ch, "search_index_percent_indexed", indexInfo.PercentIndexed, indexInfo.IndexName)
+		e.registerConstMetricGauge(ch, "search_index_hash_indexing_failures", float64(indexInfo.HashIndexingFailures), indexInfo.IndexName)
+		e.registerConstMetric(ch, "search_index_number_of_uses_total", float64(indexInfo.NumberOfUses), prometheus.CounterValue, indexInfo.IndexName)
```
nantiferov, line 93: According to my observations, this metrics is a counter and always growing. So I added `_total` to the end of name according to Prometheus recommendations.

Apparently there is `e.registerConstMetricGauge` for Gauges, but from my understanding, `e.registerConstMetric(` with `prometheus.CounterValue` should be used for Counters. Please correct me if I'm wrong.

exporter/exporter.go
oliver006: I could go either way about the pre-aggregated ones. in theory they can be calculated on the prometheus side (plus avg is only so useful) - if you want to just drop them then I
'm fine too

one thing re naming is, unit should go to the end so sohuld be something like `search_index_avg_per_record_bytes`

exporter/exporter.go
oliver006: this is fine for now.

exporter/exporter.go
oliver006: makes sense

exporter/search_indexes.go
oliver006: yup, sounds about right (see e.g. info:548)

exporter/search_indexes_test.go
oliver006: 👍

exporter/search_indexes_test.go
oliver006: Does valkey have the search module? You can change `TEST_REDIS8_URI` to a version that has the search module?

go.mod
oliver006: I just had a PR from dependabot come in, let me merge that and we can separate that change from your larger PR.

oliver006: (so you gotta rebase)

exporter/search_indexes_test.go
oliver006: One other thing you could do is this:

you refasctor all the actual parsing out line 65 onwards from your new file), and only pass in the string that you get from the call in line 58.
You can then write tests where you make up the contents of the string to your liking and see if it generates/parses the right metrics.

Something like what `TestExtractInfoMetrics()` does.

exporter/exporter.go
nantiferov: I think it's worth to keep these avg metrics just in case as I'm not sure atm about their usefulness. But it's only 4 metrics per index.

Regarding renaming. This will break naming schema for these 4 metric - now they all end with `_avg`. Though doesn't look like a big deal to me.

nantiferov: Actually I think it's worth to rename all 4 of them a little bit, like

* search_index_records_per_doc_avg  -> search_index_avg_per_doc_records
* search_index_bytes_per_record_avg -> search_index_avg_per_record_bytes
* search_index_offsets_per_term_avg -> search_index_avg_per_term_offsets
* search_index_offset_bits_per_record_avg -> search_index_avg_per_record_offset_bits

exporter/search_indexes_test.go
nantiferov: > Does valkey have the search module? You can change TEST_REDIS8_URI to a version that has the search module?

https://hub.docker.com/r/valkey/valkey-bundle should have modules built-in. I think it's worth to add it to docker-compose and test for both redis 8 + valkey.

exporter/exporter.go
nantiferov: Done

exporter/search_indexes_test.go
nantiferov: Added valkey-bundle image to tests and created issue in backlog to add modules info tests for it too in another PR.

exporter/exporter.go
oliver006: Thanks!

exporter/search_indexes_test.go
oliver006: Works, thanks!

oliver006 (approved):
Looks god, thank you!

exporter/search_indexes.go
oliver006: super nit: this should be `searchIndexInfo` and not `streamInfo`

exporter/modules_test.go
```diff
@@ -75,7 +75,7 @@ func TestModulesv74(t *testing.T) {
 
 func TestModulesv80(t *testing.T) {
 	if os.Getenv("TEST_REDIS8_URI") == "" || os.Getenv("TEST_REDIS_URI") == "" {
-		t.Skipf("TEST_REDIS8_URI / TEST_REDIS_URI aren't set - skipping")
+		t.Skipf("TEST_REDIS8_URI or TEST_REDIS_URI aren't set - skipping")
```
oliver006, line 78: 👍

exporter/search_indexes.go
nantiferov: Thank you for noticing, fixed now.

oliver006:
RFM?

Issue #841, opened 2023-10-09
Config metrics for Sentinel instances


zygisa:
Hi, 

A bit of a general question: Is it possible to get config metrics for Sentinel instances? Enabling `include-config-metrics` doesn't seem to work.

Thanks!

oliver006:
Interesting. Looks like sentinel configs can be retrieved via `sentinel config get *`

Should be an easy change, happy to review and merge a PR if you're interested in taking this on.

destinyoooo:
> Interesting. Looks like sentinel configs can be retrieved via `sentinel config get *`
> 
> Should be an easy change, happy to review and merge a PR if you're interested in taking this on.

I added sentinel config metrics in this PR https://github.com/oliver006/redis_exporter/pull/1058
Do you have time to take a look？ @oliver006

oliver006:
Fixed by https://github.com/oliver006/redis_exporter/releases/tag/v1.80.0

PR #1058, opened 2025-10-30
feat: Add config metrics for Sentinel instances

destinyoooo:
Add support for sentinel config metrics via `sentinel config get *`
Fix issue https://github.com/oliver006/redis_exporter/issues/841

### DIFF
--- a/exporter/sentinels_test.go
+++ b/exporter/sentinels_test.go
@@ -2,6 +2,7 @@ package exporter
 
 import (
 	"fmt"
+	"net/http/httptest"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


exporter/sentinels_test.go
```diff
@@ -356,3 +356,51 @@ func TestSentinelScrapeRedisHostSentinelPath(t *testing.T) {
...
+	c, err := redis.DialURL(addr)
+	if err != nil {
+		t.Fatalf("Couldn't connect to %#v: %#v", addr, err)
+	}
+	defer c.Close()
+
+	infoAll, err := redis.String(doRedisCmd(c, "INFO", "ALL"))
+	if err != nil {
+		t.Logf("Redis INFO ALL err: %s", err)
+		infoAll, err = redis.String(doRedisCmd(c, "INFO"))
```
oliver006: Is this really needed? We're connecting to a sentinel instance so should always be succeed in line :388

exporter/sentinels_test.go
```diff
@@ -356,3 +356,51 @@ func TestSentinelScrapeRedisHostSentinelPath(t *testing.T) {
...
+		go func() {
+			e.extractSentinelConfig(chM, c)
+			close(chM)
+		}()
+	} else {
+		t.Fatalf("Couldn't find sentinel section in Redis INFO: %s", infoAll)
+	}
+	allFound := true
+	for m := range chM {
+		if !strings.Contains(m.Desc().String(), "sentinel_config") {
```
oliver006: Looking for "sentinel_config" is a good start, can we look for the metric names like "sentinel_config_key_value" ?

Could also go the "doanloadMetrics()" route and check the the full metric with value is exported

exporter/sentinels_test.go
destinyoooo: OK，I have remove it

exporter/sentinels_test.go
destinyoooo: Good suggestion. I have already checked it through downloadMetrics

oliver006 (approved):

Issue #976, opened 2025-02-11
About password encryption


Pengfei-Lu:
**Describe the problem**
For now, the redis password and basic auth secret is not encrypted. Is there any plan to provide further password encryption?

[reference (node_exporter basic auth encryption)](https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md#about-bcrypt)

**What version of redis_exporter are you running?**
v1.67.0

oliver006:
> Is there any plan to provide further password encryption?

Not as of right now.

PR #1065, opened 2025-11-24
feat: add basic-auth-hash-password flag for encrypted basic auth

weidongkl:
- Add new `-basic-auth-hash-password` command line flag accepting bcrypt hashes
- Preserve legacy `-basic-auth-password` parameter for backward compatibility

fix: https://github.com/oliver006/redis_exporter/issues/976

https://prometheus.io/docs/guides/basic-auth/#hashing-a-password


weidongkl:
plainpassword
image
error hash password
image
correct hash password

image

set both hash-password and password

image

PR #1066, opened 2025-11-25
feat: add basic-auth-hash-password flag for encrypted basic auth

weidongkl:
- Add new `-basic-auth-hash-password` command line flag accepting bcrypt hashes
- Preserve legacy `-basic-auth-password` parameter for backward compatibility

fix: https://github.com/oliver006/redis_exporter/issues/976

https://prometheus.io/docs/guides/basic-auth/#hashing-a-password

plainpassword
image
error hash password
image
correct hash password

image

set both hash-password and password

image


exporter/http.go
```diff
@@ -170,23 +171,44 @@ func (e *Exporter) reloadPwdFile(w http.ResponseWriter, r *http.Request) {
...
 func (e *Exporter) verifyBasicAuth(user, password string, authHeaderSet bool) error {
-
 	if !e.isBasicAuthConfigured() {
 		return nil
 	}
 
 	if !authHeaderSet {
 		return errors.New("Unauthorized")
 	}
+	if err := e.verifyBasicUser(user); err != nil {
```
oliver006: nit: `verifyBasicUser` --> `verifyBasicAuthUser` ?

exporter/http.go
```diff
@@ -170,23 +171,44 @@ func (e *Exporter) reloadPwdFile(w http.ResponseWriter, r *http.Request) {
...
+
+func (e *Exporter) validateBasicAuthHashPassword(hashedPassword string) error {
+	return bcrypt.CompareHashAndPassword([]byte(e.options.BasicAuthHashPassword), []byte(hashedPassword))
+}
+
+func (e *Exporter) validateBasicAuthPassword(password string) error {
 	passCorrect := subtle.ConstantTimeCompare([]byte(password), []byte(e.options.BasicAuthPassword))
 
-	if userCorrect == 0 || passCorrect == 0 {
+	if passCorrect == 0 {
```
oliver006: I think we should keep the same pattern of always comparing both username and password to avoid timing attacks.

I'd rewrite `verifyBasicAuth()` and check username first (store result in "userCorrect" and then the password (depending on hash or basic) and store the result in "passCorrect" - then check and return)

main.go
```diff
@@ -80,6 +81,22 @@ func validateTLSClientConfig(certFile, keyFile string) error {
 	return nil
 }
 
+// validateHashPassword checks if the the hashed password is a valid bcrypt hash
```

main.go
```diff
@@ -287,12 +306,22 @@ func main() {
...
 	if err != nil {
 		log.Fatal(err)
 	}
+	// Validate auth parameters
+	if err := validateAuthParams(*basicAuthPassword, *basicAuthHashPassword); err != nil {
+		log.Fatal(err)
+	}
+
+	// Validate basic auth hashed password
+	if err := validateHashPassword(*basicAuthHashPassword); err != nil {
```
oliver006: Let's do this inside `validateAuthParams()` where we validate all the auth params already?

weidongkl: Okay, I will modify the code according to the review comments.

exporter/http.go
```diff
@@ -183,12 +183,24 @@ func (e *Exporter) verifyBasicAuth(user, password string, authHeaderSet bool) er
 		return errors.New("Unauthorized")
 	}
 
-	userCorrect := subtle.ConstantTimeCompare([]byte(user), []byte(e.options.BasicAuthUsername))
-	passCorrect := subtle.ConstantTimeCompare([]byte(password), []byte(e.options.BasicAuthPassword))
-
-	if userCorrect == 0 || passCorrect == 0 {
+	var (
```
oliver006: nit: can we please keep the `userCorrect :=` style?

exporter/http.go
```diff
@@ -183,12 +183,24 @@ func (e *Exporter) verifyBasicAuth(user, password string, authHeaderSet bool) er
...
+		userCorrect bool
+		passCorrect bool
+	)
+	userCorrect = subtle.ConstantTimeCompare([]byte(user), []byte(e.options.BasicAuthUsername)) > 0
+	if e.options.BasicAuthHashPassword != "" {
+		passCorrect = e.validateBasicAuthHashPassword(password)
+	} else {
+		passCorrect = subtle.ConstantTimeCompare([]byte(password), []byte(e.options.BasicAuthPassword)) > 0
+	}
+	if !userCorrect || !passCorrect {
```
oliver006, line 189: 👍

main.go
```diff
@@ -80,6 +81,22 @@ func validateTLSClientConfig(certFile, keyFile string) error {
 	return nil
 }
 
+// validateHashPassword checks if the the hashed password is a valid bcrypt hash
+func validateHashPassword(hashPassword string) error {
```
oliver006: nit: can we just inline this in `validateAuthParams()` ? it's not used anywhere.

oliver006 (commented):
Thanks, looks good. Two minor nits and then we're good to go.

exporter/http.go
weidongkl: Thank you for the review. You're right, and I've refactored the code to use the := style for immediate assignment where possible.

The key change is with the userCorrect variable, which is now directly assigned with a boolean expression using :=.

For the passCorrect variable, the var declaration is retained because its assignment depends on a conditional branch (if/else). Using var allows us to declare the variable first and then assign its value within the respective code blocks. If we used := inside both branches, they would create new, block-scoped variables, leaving the outer passCorrect variable unused.

main.go
```diff
@@ -192,8 +204,10 @@ func main() {
...
 		basicAuthUsername              = flag.String("basic-auth-username", getEnv("REDIS_EXPORTER_BASIC_AUTH_USERNAME", ""), "Username for basic authentication")
-		basicAuthPassword              = flag.String("basic-auth-password", getEnv("REDIS_EXPORTER_BASIC_AUTH_PASSWORD", ""), "Password for basic authentication")
-		inclMetricsForEmptyDatabases   = flag.Bool("include-metrics-for-empty-databases", getEnvBool("REDIS_EXPORTER_INCL_METRICS_FOR_EMPTY_DATABASES", true), "Whether to emit db metrics (like db_keys) for empty databases")
+		basicAuthPassword              = flag.String("basic-auth-password", getEnv("REDIS_EXPORTER_BASIC_AUTH_PASSWORD", ""), "Password for basic authentication, conflicts with --basic-auth-hash-password")
+		basicAuthHashPassword          = flag.String("basic-auth-hash-password", getEnv("REDIS_EXPORTER_BASIC_AUTH_HASH_PASSWORD", ""), "Hashed password for basic authentication, bcrypt format, conflicts with --basic-auth-password")
```
oliver006, line 208: Can you add this to the README as well? There's a table with all the cmd line options.

main_test.go
```diff
@@ -497,3 +497,52 @@ func TestMainFunctionsIntegration(t *testing.T) {
 		t.Error("Registry creation failed")
 	}
 }
+
+func TestValidateAuthParams(t *testing.T) {
```
oliver006, line 501: 👍

exporter/http.go
```diff
@@ -183,12 +183,23 @@ func (e *Exporter) verifyBasicAuth(user, password string, authHeaderSet bool) er
 		return errors.New("Unauthorized")
 	}
 
-	userCorrect := subtle.ConstantTimeCompare([]byte(user), []byte(e.options.BasicAuthUsername))
-	passCorrect := subtle.ConstantTimeCompare([]byte(password), []byte(e.options.BasicAuthPassword))
+	userCorrect := subtle.ConstantTimeCompare([]byte(user), []byte(e.options.BasicAuthUsername)) == 1
+
+	var passCorrect bool
+	if e.options.BasicAuthHashPassword != "" {
+		passCorrect = e.validateBasicAuthHashPassword(password)
```
oliver006: can we add a test that makes sure we hit this line in this `if` block? 

https://app.codecov.io/gh/oliver006/redis_exporter/pull/1066 suggests that it's not covered by tests.

maybe modify the new tests in http_test.go to use "verifyBasicAuth()" as the entry point instead of "validateBasicAuthHashPassword()"

main.go
weidongkl: Done. Added to the README.

exporter/http.go
weidongkl: Done! I've updated the tests in http_test.go
image

exporter/http.go
```diff
@@ -183,12 +183,23 @@ func (e *Exporter) verifyBasicAuth(user, password string, authHeaderSet bool) er
...
 
-	if userCorrect == 0 || passCorrect == 0 {
+	if !userCorrect || !passCorrect {
 		return errors.New("Unauthorized")
 	}
 
 	return nil
 }
+
+func (e *Exporter) validateBasicAuthHashPassword(hashedPassword string) bool {
```
oliver006: one more thing: we got to rename `hashedPassword` to `password`, it's not the hashed password, the hashed password is the `e.options.BasicAuthHashPassword` param

weidongkl: You're absolutely right—what's being passed in are passwords, not hashed passwords. As for changing it to validateBasicPassword, I'd suggest putting the password validation logic directly inside this method.

oliver006: e.Options... is a hashed password though, no?

weidongkl: > e.Options... is a hashed password though, no?

Yes, it's a hashed password based on the bcrypt algorithm, consistent with the Prometheus design.

PR #1067, opened 2025-11-28
feat: Improve cluster node discovery with TLS and auth

waddles:
This commit enhances the service discovery for Redis/Valkey clusters, particularly for those using TLS and authentication.

Key changes:
1. The exporter now caches authentication credentials for dynamically discovered cluster nodes, allowing it to scrape metrics from clusters that require passwords.
2. The `/discover-cluster-nodes` endpoint now correctly preserves the `rediss://` or `valkeys://` scheme for TLS-enabled clusters.
3. The discovery handler can now connect to an arbitrary cluster specified by the `target` parameter, making it more flexible.
4. TLS configuration now includes the server name (SNI), improving security and compatibility.

A comprehensive set of tests has been added to validate these new capabilities, including scenarios for:
- Discovering and scraping nodes in password-protected clusters.
- Discovering and scraping nodes in TLS-enabled clusters.
- Handling different URI schemes (`redis://`, `rediss://`, `valkey://`, `valkeys://`).

Note: Depends on https://github.com/oliver006/docker-valkey-cluster/pull/1 to update the `valkey-cluster.tmpl` file used by envsubst when building the test cluster.

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"flag"
 	"net/http"
-	"net/url"
 	"os"
 	"os/signal"
 	"runtime"
@@ -329,11 +328,7 @@ func main() {
 	if err := validateTLSClientConfig(*tlsClientCertFile, *tlsClientKeyFile); err != nil {
 		log.Fatal(err)
 	}
-	u, err := url.Parse(*redisAddr)
-	if err != nil {
-		log.Fatalf("Couldn't parse redis addr, err: %s", err)
-	}
-	_, err = exp.CreateClientTLSConfig(u.Hostname())
+	_, err = exp.CreateClientTLSConfig()
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"flag"
 	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
 	"runtime"
@@ -328,7 +329,11 @@ func main() {
 	if err := validateTLSClientConfig(*tlsClientCertFile, *tlsClientKeyFile); err != nil {
 		log.Fatal(err)
 	}
-	_, err = exp.CreateClientTLSConfig()
+	u, err := url.Parse(*redisAddr)
+	if err != nil {
+		log.Fatalf("Couldn't parse redis addr, err: %s", err)
+	}
+	_, err = exp.CreateClientTLSConfig(u.Hostname())
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,8 @@ test:
 	TEST_REDIS88_CLUSTER_MASTER_URI="redis://localhost:47000" \
 	TEST_REDIS_CLUSTER_SLAVE_URI="redis://localhost:17005" \
 	TEST_VALKEY_CLUSTER_PASSWORD_URI="redis://localhost:17006" \
+	TEST_VALKEY_CLUSTER_PASSWORD_TLS_URI="redis://localhost:17012" \
+	TEST_VALKEY_CLUSTER_PASSWORD="redis-password" \
 	TEST_TILE38_URI="redis://localhost:19851" \
 	TEST_VALKEY_SENTINEL_URI="redis://localhost:26379" \
 	go test -v -covermode=atomic -cover -race -coverprofile=coverage.txt -p 1 ./...
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ test:
 	TEST_REDIS88_CLUSTER_MASTER_URI="redis://localhost:47000" \
 	TEST_REDIS_CLUSTER_SLAVE_URI="redis://localhost:17005" \
 	TEST_VALKEY_CLUSTER_PASSWORD_URI="redis://localhost:17006" \
-	TEST_VALKEY_CLUSTER_PASSWORD_TLS_URI="redis://localhost:17012" \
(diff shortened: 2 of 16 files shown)


oliver006:
Thanks for the PR - I'll try to get to reviewing it in the next couple of days.

> Note: Depends on https://github.com/oliver006/docker-valkey-cluster/pull/1 to update the valkey-cluster.tmpl file used by envsubst when building the test cluster.

I just merged the `docker-valkey-cluster` PR and released tags 8.1.4 and 9.0.0 so you should be able to pull those in.

oliver006:
> I released 8.1.5 and 9.0.1 with your changes - and I bumped the version here to use oliver006/valkey-cluster:8.1.5

@waddles - a whole bunch of the tests related to Discovery failed, can you have a look?

main.go
```diff
@@ -317,7 +318,11 @@ func main() {
 	if err := validateTLSClientConfig(*tlsClientCertFile, *tlsClientKeyFile); err != nil {
 		log.Fatal(err)
 	}
-	_, err = exp.CreateClientTLSConfig()
+	u, err := url.Parse(*redisAddr)
+	if err != nil {
+		log.Fatalf("Couldn't parse redis addr, err: %s", err)
+	}
+	_, err = exp.CreateClientTLSConfig(u.Hostname())
```
oliver006: Is `u.Hostname()` a good default choice here? Should this be confgurable and fall back to `u.Hostname()` ? 
Also, all other TLS config is done via the exporter struct itself [down here](https://github.com/oliver006/redis_exporter/pull/1067/changes#diff-5b6103100c7f483153b677e9adf258109fa1a24fd0b22adf18bbf4d53d3e9e50R14-R16) - it would make sense to me to do the same for the hostname

waddles: `u.Hostname()` is used to extract just the hostname portion of the redis://hostname:port url which would be useful when connecting to a cluster that uses certificates with multiple SANs so you can specify the SNI. It was added because at one point I thought I had an SNI issue when connecting to an AWS Elasticache cluster but in the end I'm not sure it was required. There is probably a better way to handle it if you want to make it configurable but I couldn't see any down side to leaving it in as is.

waddles: ok so the tests all succeed but when I run the redis_exporter binary after running `make test-certs docker-env-up` then curl the exporter for cluster discovery I get
```
ERRO[81828] Cluster refresh failed: redisc: all nodes failed
tls: failed to verify certificate: x509: certificate relies on legacy Common Name field, use SANs instead
ERRO[81881] Cluster refresh failed: redisc: all nodes failed
tls: failed to verify certificate: x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs
```

The fix for that is to generate the test cert with SANs using a named pipe
```diff
diff --git a/contrib/tls/gen-test-certs.sh b/contrib/tls/gen-test-certs.sh
index 4a2810a..dafbcc8 100755
--- a/contrib/tls/gen-test-certs.sh
+++ b/contrib/tls/gen-test-certs.sh
@@ -29,4 +29,5 @@ openssl req \
         -CAserial ${dir}/ca.txt \
         -CAcreateserial \
         -days 3650 \
+        -extfile <(printf "subjectAltName=DNS:localhost,IP:127.0.0.1") \
         -out ${dir}/redis.crt

```
but I would be happy to revert main.go for my PR and let you play with SANs separately.

oliver006:
@waddles - wanna see if we can break this down into smaller piece and get this moving forward one by one? 

Maybe let's start with (4) - the SNI part (it came up in #1125) - do you think you can break that part out as a separate PR and we'll get that merged and then we can look at the rest of the changes? 

(2) also seems useful and could be a next step, not sure yet about (1) and (3) - need to think about those more.

waddles:
I have a refactor of the TLS part that's almost ready to go, I just needed to finish testing it and have been flat out for months. I will try and clean it up against master and finish testing this weekend.

I really need this PR merged soon, I just finished a heap of elasticache upgrades and even though they added a bunch of nice new cloudwatch metrics, having independent prometheus metrics will allow us to make smarter dashboards and alerts, plus it give us something to compare CW against.

oliver006:
> I have a refactor of the TLS part that's almost ready to go, I just needed to finish testing it and have been flat out for months. I will try and clean it up against master and finish testing this weekend.

Awesome, keep me posted.

waddles:
Rebased my changes onto current master.

Tested on m4 mac (arm64) where possible (some older images are x86 only). Also tested on x86 linux - unfortunately valkey-bundle:8 (valkey 8.1.8) is compiled for x86-64-v3 / AVX2 but my 2013 cpu only has AVX so it dies with a SIGILL. I also notice the `docker-test` target fails but did not commit a fix for that.

Also fixed some minor bugs and dropped the `stretchr/testify` module to simplify go mod as per commit message.

oliver006:
This is still a massive PR with many different changes lumped together - can you please break this down into a PR per feature? See my earlier comment, maybe start with item (4) - the TLS server name and then take it from there.

oliver006:
Closing this.
I added tsl server name support in #1135 
Please add a new PR if you want to add a fix for (2)

Regarding (1) - why cache passwords in the exporter? Doesn't this expose us to the risk of stale passwords? What does get me? 

And regarding (3) - I think this could be helpful but if we add this then it needs safeguards and a flag cause otherwise it can be used to probe nodes (see the flag that was added to /scrape).

If you intend to add any of these features then please separate them by PR to make review easier.

waddles:
I'm not sure it's any clearer but I split it out into #1136, #1137 and #1138 (depends on 1137). Maybe doing that made the motivation for the change clearer in the PR body but I don't think there is much point to 1137 without 1138 as you would be able to discover the nodes properly but wouldn't be able to authenticate to them.

However, it did allow me find a couple of race conditions that were previously not considered.

PR #1070, opened 2025-12-12
Introduce weekly security scanning workflow (Grype)

rjaros87:
### 🚀 Feature: Automated Container Security Scanning

This Pull Request introduces automated vulnerability scanning for Docker images using **Anchore/Grype**, significantly improving the project's security posture.

Given that `redis_exporter` images frequently surface CVEs inherited from base layers, this change ensures proactive and continuous security management.

All discovered vulnerabilities will be automatically reported and visible under the repository's **Security tab** (Code Scanning section), providing centralized and filterable results for the maintainers.


rjaros87:
Thanks @oliver006. 
My goal was simply to automate security so we don't have to worry about manual checks. Automated scans are standard practice to catch critical CVEs early, rather than reacting when it’s too late.

Issue #1031, opened 2025-08-13
Feature: Optionally add redis role label to all redis related metrics


nantiferov:
**Describe the problem**
This is feature request.

I think it would be useful to implement optional feature to add redis role label to all redis related metrics. Atm I think there's a problem how to distinguish master from replica stats on dashboard. We have selector in our dashboards to choose proper instance/pod based on role, but then when failover happens, it becomes hard to combine metrics from different instances/pods on one graph. It's possible to achieve this by using join with `redis_instance_info`, but promQL query is pretty huge after that.

I can implement PR adding flag to redis_exporter to propagate redis role to all redis metrics.

P.S. I considered adding relabeling on Prometheus side to achieve this, but apparently kubernetes_sd is having bugs detecting update in labels, so it's pretty unreliable https://github.com/prometheus/prometheus/issues/13322#issuecomment-3045694269
For ec2 it requires custom hacks to propagate role to ec2 tags.

And for both of these solutions, propagation of changed role will take time, while exporter can do it immediately.

**What version of redis_exporter are you running?**
Please run `redis_exporter --version` if you're not sure what version you're running.
[ ] 0.3x.x
[x] 1.x.x

**Running the exporter**
NA

**Screenshots**
NA

**Additional context**
NA

oliver006:
Sorry about the delay here. And thanks for writing this up, that's helpful!

>  It's possible to achieve this by using join with redis_instance_info, but promQL query is pretty huge after that.

tbh, I don;'t think that's a good reason, this can be solved in multiple ways (templating, recording derived metrics).

In a cluster scenario, I think it'd be helpful to add the run_id to a bunch of metrics (see PR #973)  so we can distinguish which process answered the metrics request. And that should also tell you what process was master / slave at a given time and from there you can go (via the query you describe) to filter or query by role.

I just don't want to pollute all the metrics with a label that can be derived (even if it's not trivial to construct the query).

Let me know if I misunderstand or miss something.

nantiferov:
Thank you for details, I'll try to check recording rules to achieve this.

oliver006:
Ok, sounds good. 
Reading your last response I think I see what the problem is with grafana and switching roles. 

Maybe you can proof out what you're trying to do with a small, hacky PR and we'll take it from there.

nantiferov:
> Maybe you can proof out what you're trying to do with a small, hacky PR and we'll take it from there.

[...]

PR #1075, opened 2025-12-24
Feature: Optionally add instance_role label to all redis related metrics

nantiferov:
If enabled, only these 12 metrics will miss `instance_role` label:

```log
$ curl -s localhost:9121/metrics|grep -v ^#|grep -v instance_role
process_cpu_seconds_total 0.009418
process_max_fds 61440
process_open_fds 13
process_resident_memory_bytes 1.5368192e+07
process_start_time_seconds 1.766599331e+09
process_virtual_memory_bytes 4.47086985216e+11
process_virtual_memory_max_bytes 9.223372036854776e+18
redis_exporter_build_info{build_date="<<< filled in by build >>>",commit_sha="<<< filled in by build >>>",golang_version="go1.25.0",version="<<< filled in by build >>>"} 1
redis_exporter_scrape_duration_seconds_sum 0.018436792
redis_exporter_scrape_duration_seconds_count 2
redis_exporter_scrapes_total 2
redis_target_scrape_request_errors_total 0
```

Metrics example with enabled option:

```log
redis_db_keys{db="db0",instance_role="master"} 10
redis_db_keys_cached{db="db0",instance_role="master"} 0
redis_db_keys_expiring{db="db0",instance_role="master"} 0
```

Fixes: #1031

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -813,10 +813,11 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 			return err
 		}
 	}
+	log.Debugf("Redis INFO ALL result: [%#v]", infoAll)
+
 	if e.options.AppendInstanceRoleLabel {
 		e.instanceRole = getInstanceRoleFromInfo(infoAll)
 	}
-	log.Debugf("Redis INFO ALL result: [%#v]", infoAll)
 
 	dbCount := 0
 	if e.options.ConfigCommandName == "-" {


exporter/exporter.go
```diff
@@ -29,7 +29,8 @@ type BuildInfo struct {
 type Exporter struct {
 	sync.Mutex
 
-	redisAddr string
+	redisAddr    string
+	instanceRole string
```
nantiferov, line 33: Not sure this is ideal place for it, but works for PoC now.

exporter/exporter.go
```diff
@@ -594,6 +596,7 @@ func NewRedisExporter(uri string, opts Options) (*Exporter, error) {
 		"stream_radix_tree_nodes":                            {txt: `Radix tree nodes count`, lbls: []string{"db", "stream"}},
 		"up":                                                 {txt: "Information about the Redis instance"},
 	} {
+		desc.lbls = append(desc.lbls, "instance_role")
```
nantiferov: We can add this label only if option is enabled, but afaik if value is not provided, label doesn't exist in metrics. 

Please let me know if it's better to add `if` here.

main.go
```diff
@@ -206,8 +206,8 @@ func main() {
 		basicAuthUsername              = flag.String("basic-auth-username", getEnv("REDIS_EXPORTER_BASIC_AUTH_USERNAME", ""), "Username for basic authentication")
 		basicAuthPassword              = flag.String("basic-auth-password", getEnv("REDIS_EXPORTER_BASIC_AUTH_PASSWORD", ""), "Password for basic authentication, conflicts with --basic-auth-hash-password")
 		basicAuthHashPassword          = flag.String("basic-auth-hash-password", getEnv("REDIS_EXPORTER_BASIC_AUTH_HASH_PASSWORD", ""), "Hashed password for basic authentication, bcrypt format, conflicts with --basic-auth-password")
-
```
nantiferov, line 210: In some recent PRs empty line was added between `basicAuthHashPassword` and `inclMetricsForEmptyDatabases`, which doesn't match previous options, so I removed it here.

README.md
```diff
@@ -212,10 +212,11 @@ scrape_configs:
...
-| basic-auth-username                 | REDIS_EXPORTER_BASIC_AUTH_USERNAME               | Username for Basic Authentication with the redis exporter needs to be set together with basic-auth-password to be effective
-| basic-auth-password                 | REDIS_EXPORTER_BASIC_AUTH_PASSWORD               | Password for Basic Authentication with the redis exporter needs to be set together with basic-auth-username to be effective, conflicts with `basic-auth-hash-password`.
-| basic-auth-hash-password            | REDIS_EXPORTER_BASIC_AUTH_HASH_PASSWORD          | Bcrypt-hashed password for Basic Authentication with the redis exporter needs to be set together with basic-auth-username to be effective,  conflicts with `basic-auth-password`. 
-| include-metrics-for-empty-databases | REDIS_EXPORTER_INCL_METRICS_FOR_EMPTY_DATABASES  | Whether to emit db metrics (like db_keys) for empty databases
```
nantiferov, line 242: All these 4 metrics were missing closing `|` for table rows, so I added them here.

README.md
```diff
@@ -263,7 +264,7 @@ Docker images are also published to the [Github Container Registry (ghcr.io)](ht
 ```sh
 docker run -d --name redis_exporter -p 9121:9121 quay.io/oliver006/redis_exporter
 
-# or
```
nantiferov, line 290: This is trailing space, automatically cleaned up.

main.go
```diff
@@ -206,8 +206,8 @@ func main() {
...
-		inclMetricsForEmptyDatabases = flag.Bool("include-metrics-for-empty-databases", getEnvBool("REDIS_EXPORTER_INCL_METRICS_FOR_EMPTY_DATABASES", true), "Whether to emit db metrics (like db_keys) for empty databases")
+		inclMetricsForEmptyDatabases   = flag.Bool("include-metrics-for-empty-databases", getEnvBool("REDIS_EXPORTER_INCL_METRICS_FOR_EMPTY_DATABASES", true), "Whether to emit db metrics (like db_keys) for empty databases")
+		appendInstanceRoleLabel        = flag.Bool("append-instance-role-label", getEnvBool("REDIS_APPEND_INSTANCE_ROLE_LABEL", false), "Whether to append 'instance_role' to all exporter metrics")
```
nantiferov: I was thinking to name label `redis_role`, but since exporter supports more variations and existing variable named `instanceRole`, I think `instance_role` sounds better.

exporter/exporter.go
nantiferov: Never-mind, tests in CI proved that label shouldn't be added if not used.

It was causing panics in local test, like:

```log
panic: inconsistent label cardinality: expected 2 label values but got 1 in []string{"set"}

goroutine 39 [running]:
github.com/prometheus/client_golang/prometheus.MustNewConstSummary(...)
```

exporter/metrics.go
```diff
@@ -88,7 +88,9 @@ func (e *Exporter) registerConstMetric(ch chan<- prometheus.Metric, metric strin
 	} else {
 		desc = e.mustFindMetricDescription(metric)
 	}
-
+	if e.options.AppendInstanceRoleLabel { // append instance_role label to all metrics
```
nantiferov: If idea look ok in general, I'll try to add tests to cover it. 
Right now coverage decreased by 0.2%

main.go
oliver006: maybe just `role` ?

oliver006:
Thanks for the PR - I'll try to get to it in the coming days but not sure yet how much time I'll have.

main.go
nantiferov: My concern initially was that `redis_instance_info` metric has exactly same label with same value.

But now after quick check I found that my option breaks existing metrics and there's >100 metrics are missing:

```log
# default exporter run
$ curl -s localhost:9121/metrics|grep -v ^#|wc -l                   
     355

# with `-append-instance-role-label` enabled
$ curl -s localhost:9121/metrics|grep -v ^#|wc -l
     205
```

I'll mark PR as draft and will check which and why metrics are missing.

nantiferov: Currently missing metrics:
```diff
--- metrics.disabled.2	2025-12-26 20:47:35
+++ metrics.enabled.2	2025-12-26 20:47:27
@@ -5,36 +5,6 @@
 process_start_time_seconds
 process_virtual_memory_bytes
 process_virtual_memory_max_bytes
-redis_acl_access_denied_auth_total
-redis_acl_access_denied_channel_total
-redis_acl_access_denied_cmd_total
-redis_acl_access_denied_key_total
-redis_active_defrag_running
-redis_allocator_active_bytes
-redis_allocator_allocated_bytes
-redis_allocator_frag_bytes
-redis_allocator_frag_ratio
-redis_allocator_muzzy_bytes
-redis_allocator_resident_bytes
-redis_allocator_rss_bytes
-redis_allocator_rss_ratio
-redis_aof_current_rewrite_duration_sec
-redis_aof_enabled
-redis_aof_last_bgrewrite_status
-redis_aof_last_cow_size_bytes
-redis_aof_last_rewrite_duration_sec
-redis_aof_last_write_status
-redis_aof_rewrite_in_progress
-redis_aof_rewrite_scheduled
-redis_async_loading
-redis_blocked_clients
-redis_client_output_buffer_limit_disconnections_total
-redis_client_query_buffer_limit_disconnections_total
-redis_client_recent_max_input_buffer_bytes
-redis_client_recent_max_output_buffer_bytes
-redis_clients_in_timeout_table
-redis_cluster_connections
-redis_cluster_enabled
 redis_commands_duration_seconds_total
 redis_commands_duration_seconds_total
 redis_commands_duration_seconds_total
@@ -123,7 +93,6 @@
 redis_commands_latencies_usec_sum
 redis_commands_latencies_usec_sum
 redis_commands_latencies_usec_sum
-redis_commands_processed_total
 redis_commands_rejected_calls_total
 redis_commands_rejected_calls_total
 redis_commands_rejected_calls_total
@@ -149,20 +118,6 @@
 redis_config_client_output_buffer_limit_overcome_seconds
 redis_config_client_output_buffer_limit_overcome_seconds
 redis_config_client_output_buffer_limit_overcome_seconds
-redis_config_io_threads
-redis_config_maxclients
-redis_config_maxmemory
-redis_configured_hz
-redis_connected_clients
-redis_connected_slaves
-redis_connections_received_total
-redis_cpu_sys_children_seconds_total
-redis_cpu_sys_main_thread_seconds_total
-redis_cpu_sys_seconds_total
-redis_cpu_user_children_seconds_total
-redis_cpu_user_main_thread_seconds_total
-redis_cpu_user_seconds_total
-redis_current_eviction_exceeded_time_ms
 redis_db_avg_ttl_seconds
 redis_db_keys_cached
 redis_db_keys_expiring
@@ -197,39 +152,11 @@
 redis_db_keys
 redis_db_keys
 redis_db_keys
-redis_defrag_hits
-redis_defrag_key_hits
-redis_defrag_key_misses
-redis_defrag_misses
-redis_dump_payload_sanitizations
-redis_eventloop_cycles_total
-redis_eventloop_duration_cmd_sum_usec_total
-redis_eventloop_duration_sum_usec_total
-redis_evicted_clients_total
-redis_evicted_keys_total
-redis_evicted_scripts_total
-redis_eviction_exceeded_time_ms_total
-redis_expire_cycle_cpu_time_ms_total
-redis_expired_keys_total
-redis_expired_stale_percentage
-redis_expired_subkeys_total
-redis_expired_time_cap_reached_total
 redis_exporter_build_info
-redis_exporter_last_scrape_connect_time_seconds
-redis_exporter_last_scrape_duration_seconds
 redis_exporter_last_scrape_error
 redis_exporter_scrape_duration_seconds_count
 redis_exporter_scrape_duration_seconds_sum
 redis_exporter_scrapes_total
-redis_hz
-redis_instance_info
-redis_instantaneous_eventloop_cycles_per_sec
-redis_instantaneous_eventloop_duration_usec
-redis_io_threaded_reads_processed
-redis_io_threaded_writes_processed
-redis_io_threads_active
-redis_keyspace_hits_total
-redis_keyspace_misses_total
 redis_last_key_groups_scrape_duration_milliseconds
 redis_last_slow_execution_duration_seconds
 redis_latency_percentiles_usec_count
@@ -272,88 +199,8 @@
 redis_latency_percentiles_usec
 redis_latency_percentiles_usec
 redis_latency_percentiles_usec
-redis_latest_fork_seconds
-redis_lazyfree_pending_objects
-redis_lazyfreed_objects
-redis_loading_dump_file
-redis_master_repl_offset
-redis_max_clients
-redis_mem_aof_buffer_bytes
-redis_mem_clients_normal
-redis_mem_clients_slaves
-redis_mem_cluster_links_bytes
-redis_mem_fragmentation_bytes
-redis_mem_fragmentation_ratio
-redis_mem_not_counted_for_eviction_bytes
-redis_mem_overhead_db_hashtable_rehashing_bytes
-redis_mem_replication_backlog_bytes
-redis_mem_total_replication_buffers_bytes
-redis_memory_max_bytes
-redis_memory_used_bytes
-redis_memory_used_dataset_bytes
-redis_memory_used_functions_bytes
-redis_memory_used_lua_bytes
-redis_memory_used_overhead_bytes
-redis_memory_used_peak_bytes
-redis_memory_used_rss_bytes
-redis_memory_used_scripts_bytes
-redis_memory_used_scripts_eval_bytes
-redis_memory_used_startup_bytes
-redis_memory_used_vm_eval_bytes
-redis_memory_used_vm_functions_bytes
-redis_memory_used_vm_total
-redis_migrate_cached_sockets_total
-redis_module_fork_in_progress
-redis_module_fork_last_cow_size
-redis_net_input_bytes_total
-redis_net_output_bytes_total
-redis_net_repl_input_bytes_total
-redis_net_repl_output_bytes_total
-redis_number_of_cached_scripts
-redis_number_of_functions
-redis_number_of_libraries
-redis_process_id
-redis_pubsub_channels
-redis_pubsub_clients
-redis_pubsub_patterns
-redis_pubsubshard_channels
-redis_rdb_bgsave_in_progress
-redis_rdb_changes_since_last_save
-redis_rdb_current_bgsave_duration_sec
-redis_rdb_last_bgsave_duration_sec
-redis_rdb_last_bgsave_status
-redis_rdb_last_cow_size_bytes
-redis_rdb_last_load_expired_keys
-redis_rdb_last_load_loaded_keys
-redis_rdb_last_save_timestamp_seconds
-redis_rdb_saves_total
-redis_rejected_connections_total
-redis_repl_backlog_first_byte_offset
-redis_repl_backlog_history_bytes
-redis_repl_backlog_is_active
-redis_replica_partial_resync_accepted
-redis_replica_partial_resync_denied
-redis_replica_resyncs_full
-redis_replication_backlog_bytes
-redis_reply_buffer_expands_total
-redis_reply_buffer_shrinks_total
-redis_second_repl_offset
-redis_slave_expires_tracked_keys
 redis_slowlog_last_id
 redis_slowlog_length
 redis_start_time_seconds
 redis_target_scrape_request_errors_total
-redis_total_blocking_keys
-redis_total_blocking_keys_on_nokey
-redis_total_error_replies
-redis_total_reads_processed
-redis_total_watched_keys
-redis_total_writes_processed
-redis_tracking_clients
-redis_tracking_total_items
-redis_tracking_total_keys
-redis_tracking_total_prefixes
-redis_unexpected_error_replies
 redis_up
-redis_uptime_in_seconds
-redis_watching_clients
```

I think it's because I haven't added label to metricDescriptions, but adding label value later.

nantiferov: Fixed bug, now with both option enabled and disabled, metrics count stays same.

Speaking of using label `role` instead of `instance_role`, I think it will make code even more complicated, as there should be added exception for `redis_instance_info` metric name to not add dynamically label and value.

I might be wrong though, if you think it could be implemented nicely,  I don't mind in general to keep label name unified.

oliver006: Thanks for the deep dive, agree with not making things more complicated due to working around the `redis_instance_info`.

We cna leave as `instance_role`. Or another idea would be `cluster_role`. Maybe just leave as `instance_role`

README.md
```diff
@@ -212,10 +212,11 @@ scrape_configs:
...
+| append-instance-role-label          | REDIS_APPEND_INSTANCE_ROLE_LABEL                 | Whether to append 'instance_role' label to all exporter metrics.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
```
oliver006: Let's add 1-2 sentences here (or in a seperate section of the README) to explain what `instance_role` means and maybe even how to use it/think about it (you had a great explanation in that other thread, explaining that I can't rely on joining instance metrics).

exporter/exporter.go
```diff
@@ -594,6 +596,9 @@ func NewRedisExporter(uri string, opts Options) (*Exporter, error) {
 		"stream_radix_tree_nodes":                            {txt: `Radix tree nodes count`, lbls: []string{"db", "stream"}},
 		"up":                                                 {txt: "Information about the Redis instance"},
 	} {
+		if e.options.AppendInstanceRoleLabel {
```
oliver006, line 610: Slick to to this here.

exporter/info.go
```diff
@@ -170,6 +170,7 @@ func (e *Exporter) extractInfoMetrics(ch chan<- prometheus.Metric, info string,
 	}
 
 	instanceRole := keyValues["role"]
```
oliver006, line 172: I guess we call it `instanceRole` here to the proposed label `instance_role` is fine.

main.go
```diff
@@ -206,8 +206,8 @@ func main() {
...
 		basicAuthPassword              = flag.String("basic-auth-password", getEnv("REDIS_EXPORTER_BASIC_AUTH_PASSWORD", ""), "Password for basic authentication, conflicts with --basic-auth-hash-password")
 		basicAuthHashPassword          = flag.String("basic-auth-hash-password", getEnv("REDIS_EXPORTER_BASIC_AUTH_HASH_PASSWORD", ""), "Hashed password for basic authentication, bcrypt format, conflicts with --basic-auth-password")
-
-		inclMetricsForEmptyDatabases = flag.Bool("include-metrics-for-empty-databases", getEnvBool("REDIS_EXPORTER_INCL_METRICS_FOR_EMPTY_DATABASES", true), "Whether to emit db metrics (like db_keys) for empty databases")
+		inclMetricsForEmptyDatabases   = flag.Bool("include-metrics-for-empty-databases", getEnvBool("REDIS_EXPORTER_INCL_METRICS_FOR_EMPTY_DATABASES", true), "Whether to emit db metrics (like db_keys) for empty databases")
+		appendInstanceRoleLabel        = flag.Bool("append-instance-role-label", getEnvBool("REDIS_APPEND_INSTANCE_ROLE_LABEL", false), "Whether to append 'instance_role' label to all exporter metrics")
```
oliver006: This currently defaults to `false` which makes sense.
I could see it being useful to make it default to `true` if the exporter is in cluster mode but I don't think there's a good way to do this weithout introducing basically a `nil` (aka not set) state and that'll make things ugly.

oliver006: Maybe you can just add it to the cluster part of the README here
https://github.com/oliver006/redis_exporter?tab=readme-ov-file#prometheus-configuration-to-scrape-all-nodes-in-a-redis-cluster

so that someone that sets it up for a cluster is immediately aware of this config flag

main.go
nantiferov: > Or another idea would be cluster_role

I believe this additional label propagation is useful not only for clusters, but also for master/replica setup with sentinel. We plan to use it for both. This is especially useful for dynamic environments, like K8s.

So I think `instance_role` would suit better than `cluster_role`.

main.go
nantiferov: Sure, thank you, will update Readme parts together with https://github.com/oliver006/redis_exporter/pull/1075#discussion_r2655386746

nantiferov:
Thank you for review.

> we should def add a test that checks that instance_role appears as expected

I'll try to add tests and address other parts of review next week.

README.md
nantiferov: Expanded description.

main.go
nantiferov: Done, expanded description of option and added P.S. to cluster section.

exporter/exporter_test.go
```diff
@@ -476,6 +477,26 @@ func TestRedisMetricsOnly(t *testing.T) {
 	}
 }
 
+func TestRedisAppendInstanceRoleLabel(t *testing.T) {
```
nantiferov, line 480: This function is made similar to others in this file. It catches if labels are present or not, depending on option.

I also noticed that big portion of metrics don't set properly value for `instance_role` label, not sure why. On real tests I haven't noticed that.

**UPD**: It reproduces on real tests as well - first scrapes just after exporter start returns `{instance_role=""}`

Example [log](https://github.com/oliver006/redis_exporter/actions/runs/21009513982/job/60400622847?pr=1075#step:10:1070):

```
        # HELP test_uptime_in_seconds uptime_in_seconds metric
        # TYPE test_uptime_in_seconds gauge
        test_uptime_in_seconds{instance_role=""} 71
        # HELP test_watching_clients watching_clients metric
        # TYPE test_watching_clients gauge
        test_watching_clients{instance_role=""} 0
```

Seems only latencies histograms have it:

```
        test_commands_latencies_usec_sum{cmd="set",instance_role="master"} 99
        test_commands_latencies_usec_count{cmd="set",instance_role="master"} 18
        test_commands_latencies_usec_bucket{cmd="setex",instance_role="master",le="2"} 5   
        test_commands_latencies_usec_bucket{cmd="setex",instance_role="master",le="4"} 13
```

nantiferov: They're called right after `e.extractInfoMetrics` at 
https://github.com/oliver006/redis_exporter/blob/master/exporter/exporter.go#L849

But other metrics seems collected after `e.extractLatencyMetrics`

exporter/metrics_test.go
```diff
@@ -21,21 +21,30 @@ func TestSanitizeMetricName(t *testing.T) {
 }
 
 func TestRegisterConstHistogram(t *testing.T) {
```
nantiferov, line 23: PTAL if this approach for test looks good.

Basically I extended existing `TestRegisterConstHistogram` with check for both `AppendInstanceRoleLabel` combinations.

If this looks good, I think similar function should be created for `registerConstMetric` and `registerConstSummary`.

Maybe it's possible to make this function universal and pass go functions as parameter, wdyt?

```
go func() {
			exp.createMetricDescription(metricName, []string{"test"})
			exp.registerConstHistogram(ch, metricName, 12, .24, map[float64]uint64{}, "test")
			close(ch)
		}
```

Basically only `exp.registerConstHistogram()` part differs I think.

main.go
oliver006: > instance_role

👍

exporter/exporter_test.go
oliver006: > UPD: It reproduces on real tests as well - first scrapes just after exporter start returns {instance_role=""}

I think I see the problem and I think it has directly to do with your decision to store `instance_role` in the `Exporter` struct.
(and I think this might be a bigger problem for the entire approach here)

on the first scrape, the instanceRole is still empty. It's not set until the exporter gets to the part that parses the `GET INFO` output and encounters the instance role and sets the field.
That means that every call to `registerConstMetric|Summary|Histrogram` uses the emtpy (or outdated from the last scrape) value of instanceRole until we find the value in `GET INFO`.

My apologies, I said earlier that the approach looks fine but I think this makes the approach is somewhat flawed. 
Two ways to do this:
1. "collect" all metrics (in a temp place) before submitting them, and then, in the end, when you have all the metrics and you have the instance_role, you can set the label and ship off all the metrics
2. the very first thing you do for a scrape is to you call GET INFO and the nyou find the instance_role. Then you continue to extract and submit the remaining metrics as you go (just like right now) using the extracted instance role.

(2) makes more sense to me. Could even call "GET INFO" twice if that makes things significantly easier, the extra call makes for a pretty small overhead.

What do you think? Am I wrong with my assessment that we can't continue with the general approach as is?

nantiferov: No worries, good that we caught this issue at PR stage.

Second approach looks better to me as well. I'll mark PR as draft and try to implement it.

oliver006: Here's another (related) thought: let's try to not store state in the exporter struct for the currently ongoing scrape *OR* put a mutex in there so there can be only one scrape going on at once.

nantiferov: In this commit I moved **InstanceRole** collection to the earliest point in `scrapeRedisHost()` function.

https://github.com/oliver006/redis_exporter/pull/1075/changes/ba53a45ddc36d5f71a6e703df121f956fea16b1a

Looks like it fixes the issue.


> let's try to not store state in the exporter struct for the currently ongoing scrape OR put a mutex in there so there can be only one scrape going on at once.

I'm not 100% sure, but there's a lock at the beginning of metrics collection:

```go
// Collect fetches new metrics from the RedisHost and updates the appropriate metrics.
func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
	e.Lock()
	defer e.Unlock()
```

Are you sure there's a need for one more for instanceRole only?

oliver006: You're right, no need for another.

exporter/metrics_test.go
oliver006: I think that works.
Or you use the approach where you down the /metrics and check if the `instance_role="..."` is present. Both are fine.

README.md
```diff
@@ -236,10 +214,11 @@ is deployed in container environments.
...
+| append-instance-role-label          | REDIS_APPEND_INSTANCE_ROLE_LABEL                 | Whether to append 'instance_role' label to all exporter metrics. It allows easy creation of dashboards/alerts with a selector for instance role (master/replica). NOTE: This increases the cardinality of Redis metrics.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
```
oliver006: here and in main.go

`REDIS_APPEND_INSTANCE_ROLE_LABEL` 
-->
`REDIS_EXPORTER_APPEND_INSTANCE_ROLE_LABEL`

nantiferov: Thank you, fixed now.

exporter/exporter_test.go
```diff
@@ -10,13 +10,14 @@ package exporter
...
 import (
 	"fmt"
-	"github.com/mna/redisc"
 	"net/http/httptest"
 	"os"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/mna/redisc"
```
nantiferov, line 19: This import moved from line 13 after `go fmt`

exporter/exporter_test.go
nantiferov: I included check for `instance_role=""` in tests and found one more corner case with metrics registering before fetching role.

So I moved it up, right after connect, https://github.com/oliver006/redis_exporter/pull/1075/changes#diff-c722ac09a12b0798d4b56729c7f9c7236651f195ea1d3ee985db75543d5ce219R785

exporter/metrics_test.go
nantiferov: Done, added check for TestRegisterConstMetric as well.

nantiferov:
I added tests, checked that they seems working correctly.
I think PR is ready for review now, PTAL when you have time.

exporter/exporter.go
```diff
@@ -751,7 +767,6 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 	startTime := time.Now()
 	c, err := e.connectToRedis()
 	connectTookSeconds := time.Since(startTime).Seconds()
-	e.registerConstMetricGauge(ch, "exporter_last_scrape_connect_time_seconds", connectTookSeconds)
```
oliver006, line 754: Why move this? So you can grab the role first?

exporter/exporter.go
```diff
@@ -767,6 +782,15 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 	}
 	defer c.Close()
 
+	// Get redis role right after connect before metrics register
+	infoTmp, err := redis.String(doRedisCmd(c, "INFO"))
```
oliver006: Let's do this only when we need to (e.g. the `e.options.AppendInstanceRoleLabel` flag is set) and otherwise skip this block

oliver006: In addition

1. could do just "INFO replication" to move less data over the wire 
2. and use redis.Strings instead of splitting this into rows ourselves? 
   (I guess I could/should that too for the other `INFO` call further down but let's save that for another PR)

exporter/exporter.go
oliver006: Main concern is that we no longer record this metric now if the first command fails (e.g. the "INFO" cmd to extract the role)

exporter/exporter.go
oliver006: Thinking about this a little more, why can't we re-use the result of this INFO call? 
I think calling INFO twice is not the end of hte world but if you'd move the block from line 830 up here (gotta do the full "INFO ALL" thing), then you can extract the role first, get everything set up and then later down around line 830 you can just go ahead with the already received INFO data.
I know this is a bit of a refactor late in the PR but I don't think it's a big change.

What do you think?

exporter/exporter.go
nantiferov: Yep, I moved it to have instance_role set properly on first scrape.

Another option I think is to exclude adding instance_role label to `exporter_last_scrape_connect_time_seconds` metric.

exporter/exporter.go
nantiferov: Yep, I agree it makes sense to get info earlier here, let me check this.

exporter/exporter.go
```diff
@@ -788,6 +804,18 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 		}
 	}
 
+	infoAll, err := redis.String(doRedisCmd(c, "INFO", "ALL"))
```
nantiferov, line 807: I moved info collection just a little bit in order to set properly label for extractConfigMetrics.

exporter/exporter.go
nantiferov: I excluded setting this label for couple of metrics that collected before getting infoAll.

Please let me know what do you think, it looks as quite good compromise to me.
https://github.com/oliver006/redis_exporter/pull/1075/changes#diff-76e8689da13c206d6904d14dea719b7808bd7b18c7b1b45935f99090aa835c43R15

exporter/exporter.go
nantiferov: Movement part done, will check `redis.Strings` little bit later

exporter/exporter.go
oliver006: 👍

exporter/metrics.go
```diff
@@ -11,6 +12,10 @@ import (
 )
 
 var metricNameRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
+var skipInstanceRoleLabel = []string{ // These metrics are collected before instanceRole set
```
oliver006: use a `map[string]bool` and it'll make the lookup a lot easier (and possibly faster if we ever add more metrics)

considering the amall amount i'm also fine with making it `s == "m1" || s == "m2"` and be done with it

nantiferov: > considering the amall amount i'm also fine with making it s == "m1" || s == "m2" and be done with it

Yep, I was thinking in the beginning it will be required in all places, so moved it to variable. But now it's indeed only couple of times.

nantiferov: Fixed in https://github.com/oliver006/redis_exporter/pull/1075/commits/cc756bfdc88fba8156e95e0288c0d51a6b5c18d8

exporter/exporter.go
```diff
@@ -788,6 +804,21 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
...
+		log.Debugf("Redis INFO ALL err: %s", err)
+		infoAll, err = redis.String(doRedisCmd(c, "INFO"))
+		if err != nil {
+			log.Errorf("Redis INFO err: %s", err)
+			return err
+		}
+	}
+	log.Debugf("Redis INFO ALL result: [%#v]", infoAll)
+
+	if e.options.AppendInstanceRoleLabel {
```
nantiferov, line 818: I also wrapped in if call to getInstanceRoleFromInfo()

exporter/exporter.go
nantiferov: I checked comment about `redis.Strings` and this method in general and it seems it's irrelevant now since we're using now existing `InfoAll` variable, which is string.

Looks like PR is ready for review now, I fixed all the issues.
PTAL when you have time.

exporter/exporter.go
```diff
@@ -813,9 +813,12 @@ func (e *Exporter) scrapeRedisHost(ch chan<- prometheus.Metric) error {
 			return err
 		}
 	}
-	e.instanceRole = getInstanceRoleFromInfo(infoAll)
 	log.Debugf("Redis INFO ALL result: [%#v]", infoAll)
 
+	if e.options.AppendInstanceRoleLabel {
+		e.instanceRole = getInstanceRoleFromInfo(infoAll)
```
oliver006, line 819: 👍

PR #1083, opened 2026-01-23
chore: update Go version from 1.25.0 to 1.25.5

sfc-gh-jsyryczynski:
[Version 1.25.5](https://github.com/golang/go/issues?q=milestone%3AGo1.25.5+label%3ACherryPickApproved) contains two security fixes to the crypto/x509 package and other fixes.

I would appreciate a release with this change.

Previous PR with this change had to be closed: https://github.com/oliver006/redis_exporter/pull/1079
Organization policy require all OSS commits coming from Snowflake-Labs github org.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG ALPINE_VERSION=3.23
 #
 # build container
 #
-FROM --platform=$BUILDPLATFORM golang:1.25.5-alpine${ALPINE_VERSION} AS builder
+FROM --platform=$BUILDPLATFORM golang:1.25-alpine${ALPINE_VERSION} AS builder
 WORKDIR /go/src/github.com/oliver006/redis_exporter/
 
 ADD . /go/src/github.com/oliver006/redis_exporter/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG ALPINE_VERSION=3.23
 #
 # build container
 #
-FROM --platform=$BUILDPLATFORM golang:1.25-alpine${ALPINE_VERSION} AS builder
```
oliver006, line 7: Please keep this at `1.25` so it'll pull in the latest version during release time - I don't want to constantly updating this whenever a new minor version got released

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/oliver006/redis_exporter
 
-go 1.25.0
+go 1.25.5
```
oliver006: Same, make this `1.25` - no need to impose the minor version on anyone that wants to include this as a module (there are some people that create "super exporters")

sfc-gh-jsyryczynski: Done, removed patch version

Dockerfile
sfc-gh-jsyryczynski: Done, removed patch version

PR #1097, opened 2026-03-06
Add sentinel_peer_info metric for Sentinel peer discovery

tomatopunk:
### Summary
When the exporter is pointed at a Redis Sentinel, it now exposes a new metric `sentinel_peer_info` that lists **other Sentinel peers** discovered via `SENTINEL SENTINELS`. This allows discovering all Sentinel addresses by scraping a single Sentinel instance.

### Motivation
- Operators need to know the full set of Sentinel peers (e.g. for monitoring, failover tooling, or service discovery).
- Scraping one Sentinel is enough: `SENTINEL SENTINELS <master_name>` returns the other Sentinels monitoring that master. Each peer is emitted as one time series, so there is no overlap or overwriting when multiple Sentinels are monitored.

### Changes

**New metric: `sentinel_peer_info`**
- **Type:** Gauge (value 1, info-style).
- **Labels:** `master_name`, `master_address`, `name`, `ip`, `port`, `runid`, `flags`.
- Only stable, low-cardinality labels are used to avoid time-series churn (e.g. no `down_after_milliseconds` or `voted_leader`).

### Example output
```text
# HELP redis_sentinel_peer_info Other Sentinel peers discovered via SENTINEL SENTINELS (one scrape from one Sentinel)
# TYPE redis_sentinel_peer_info gauge
redis_sentinel_peer_info{master_name="mymaster",master_address="127.0.0.1:6379",name="abc...",ip="192.168.1.2",port="26379",runid="abc...",flags="sentinel"} 1
```

### Testing
- Existing Sentinel tests remain valid; the new metric is emitted in addition to `sentinel_master_ok_sentinels`.
- No new dependencies; behavior is backward compatible (new metric only when scraping a Sentinel).


exporter/sentinels.go
```diff
@@ -152,6 +158,28 @@ func (e *Exporter) processSentinelSentinels(ch chan<- prometheus.Metric, sentine
...
+		name := ""
+		if v, ok := sentinelDetailMap["name"]; ok {
+			name = v
+		}
+		ip := ""
+		if v, ok := sentinelDetailMap["ip"]; ok {
+			ip = v
+		}
+		port := ""
+		if v, ok := sentinelDetailMap["port"]; ok {
```
oliver006, line 171: What port is that? 
If it's changing frequently then this can blow up the cardinality of this metric significantly.
Same concern for `runid`

oliver006: bump

oliver006: ah, nvm, I just saw you're optionally including all of this based on the flag. makes sense. Resolving.

exporter/sentinels.go
```diff
@@ -152,6 +158,30 @@ func (e *Exporter) processSentinelSentinels(ch chan<- prometheus.Metric, sentine
...
+			port = v
+		}
+		runid := ""
+		if v, ok := sentinelDetailMap["runid"]; ok {
+			runid = v
+		}
+		flags := ""
```
oliver006, line 178: let's use this down in line 175 and onwards instead of extracting it again down there

exporter/sentinels.go
```diff
@@ -144,6 +144,12 @@ func (e *Exporter) processSentinelSentinels(ch chan<- prometheus.Metric, sentine
 
 	// If we are here then this master is in ok state
 	masterOkSentinels := 1
+	masterName := ""
+	masterAddr := ""
+	if len(labels) >= 2 {
```
oliver006: Do we still want to emit the metric if labels is empty? 
(plus we outright use labels down in 197 for the other metric)
would be good to be consistent here and maybe just return if labels is empty?

exporter/exporter.go
```diff
@@ -614,6 +616,10 @@ func NewRedisExporter(uri string, opts Options) (*Exporter, error) {
 		e.metricDescriptions[k] = newMetricDescr(opts.Namespace, k, desc.txt, desc.lbls)
 	}
 
+	if !opts.InclSentinelPeerInfo {
```
oliver006: I don't think we need to delete the description (and we don't do it for any other flags), just not emitting the metric should be enough?

Issue #1115, opened 2026-05-09
Two more things on /scrape — UNIX-socket reach and an INFO reply that kills the process


0xVijay:
Building on #1104 / #1106 — two extensions to the `/scrape?target=` problem worth flagging on `oliver006/redis_exporter:v1.77.0-alpine`.

**Scheme matrix is wider than `redis://`.** `connectToRedis` falls back to `redis.Dial(scheme, host, ...)` at `exporter/redis.go:79` when `DialURL` rejects the URL, feeding the user-supplied scheme to `net.Dialer` as the network argument. So `unix://`, `tcp://`, `tcp4://`, schemeless `host:port`, and `valkey://` all work — and the exporter writes `AUTH <REDIS_PASSWORD>` as the first bytes on whatever it dials. The UNIX-socket case is the one that bit me: an exporter container with `/var/run/docker.sock` or a sidecar postgres socket bind-mounted gets the password written straight into that socket.

**`INFO` reply with an empty percentile key panics the process.** `extractPercentileVal` at `exporter/info.go:60` does `split[0][1:]` without a length check:

```go
percentile, err = strconv.ParseFloat(split[0][1:], 64)  // panic if split[0] == ""
```

A reply line like `latency_percentiles_usec_get:=1.0,p99=2.0` (leading `=`) splits to `["", "1.0"]` and `""[1:]` panics. Default config, no flag needed. The panic blows past the http handler and the process exits with code 2 — under `restart: always` one curl becomes a restart loop.

### Repro

```bash
docker network create rxlab
docker run -d --name listener --network rxlab alpine sh -c \
  'apk add -q socat && exec socat -v -x TCP-LISTEN:6666,reuseaddr,fork - 2>/tmp/cap.log'
docker run -d --name rx --network rxlab -p 9121:9121 \
  -e REDIS_PASSWORD='LAB_TEST_VALUE' oliver006/redis_exporter:v1.77.0-alpine

curl -sS 'http://localhost:9121/scrape?target=redis://listener:6666/0' >/dev/null
docker exec listener cat /tmp/cap.log
# → *2\r\n$4\r\nAUTH\r\n$14\r\nLAB_TEST_VALUE\r\n
```

For the panic, point a fake-Redis at the exporter that returns the malformed line for any `INFO`; one `/scrape` request and `docker inspect rx` shows `exited exit=2`.

### Suggested fix

```go
// exporter/redis.go connectToRedis — allowlist scheme.
[...]

PR #1118, opened 2026-05-11
fix: guard empty percentile key in extractPercentileVal to prevent panic

0xVijay:
A `latency_percentiles_usec_*` line with an empty key segment (leading `=`, e.g. `=1.0,p99=2.0`) splits to `["", "1.0"]` and `split[0][1:]` panics with index out of range. The panic propagates through promhttp and exits the process, so under `restart: always` one malformed `INFO` reply becomes a restart loop.

Adds a length check before the slice and returns an error so the existing caller (`parseMetricsLatencyStats`) records the parse failure and moves on. Test case added covering the empty-key input.

Refs #1115.

### DIFF
--- a/exporter/info.go
+++ b/exporter/info.go
@@ -57,6 +57,10 @@ func extractPercentileVal(s string) (percentile float64, val float64, err error)
 	if len(split) != 2 {
 		return
 	}
+	if len(split[0]) == 0 {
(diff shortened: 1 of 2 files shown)


0xVijay:
Closing to clean up commit attribution; will reopen with the same fix from a fresh fork.

PR #1119, opened 2026-05-11
fix: guard empty percentile key in extractPercentileVal to prevent panic

0xVijay:
A `latency_percentiles_usec_*` line with an empty key segment (leading `=`, e.g. `=1.0,p99=2.0`) splits to `["", "1.0"]` and `split[0][1:]` panics with index out of range. The panic propagates through promhttp and exits the process, so under `restart: always` one malformed `INFO` reply becomes a restart loop.

Adds a length check before the slice and returns an error so the existing caller (`parseMetricsLatencyStats`) records the parse failure and moves on. Test case added covering the empty-key input.

Refs #1115.

### DIFF
--- a/exporter/info.go
+++ b/exporter/info.go
@@ -57,6 +57,10 @@ func extractPercentileVal(s string) (percentile float64, val float64, err error)
 	if len(split) != 2 {
 		return
 	}
+	if len(split[0]) == 0 {
+		err = fmt.Errorf("empty percentile key")
+		return
+	}
 	percentile, err = strconv.ParseFloat(split[0][1:], 64)
 	if err != nil {
 		return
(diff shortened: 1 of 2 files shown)


exporter/info.go
```diff
@@ -57,6 +57,10 @@ func extractPercentileVal(s string) (percentile float64, val float64, err error)
 	if len(split) != 2 {
 		return
 	}
+	if len(split[0]) == 0 {
```
oliver006, line 60: Shouldn't this be "len() < 2" ?

oliver006: nvm, I think this is right

oliver006:
Thanks for the PR - I'll try to get look at this in more depth in the next few days.

oliver006:
Test failure should be unrelated (and fixed by #1124) - I'll try to review this asap

oliver006 (approved):
👍

PR #1122, opened 2026-05-16
fix(http): correct format verb typo in scrapeHandler error message

jackiesre721:
## Summary

Fix a format verb typo in the `/scrape` endpoint error message that produces garbled output when an invalid target URL is provided.

## Bug

In `exporter/http.go:61`, the format string uses `%ck` which is not a valid Go format verb:

```go
http.Error(w, fmt.Sprintf("Invalid 'target' parameter, parse err: %ck ", err), http.StatusBadRequest)
```

Go's `fmt.Sprintf` interprets `%ck` as an invalid format sequence, producing garbled output like:

```
Invalid target parameter, parse err: %!c(string=parse ://: missing protocol scheme)k
```

## Fix

Change `%ck` to `%v` for correct error message display:

```
Invalid target parameter, parse err: parse ://: missing protocol scheme
```

## Reproduction

```bash
curl "http://localhost:9121/scrape?target=://invalid"
```

Before fix: `Invalid 'target' parameter, parse err: %!c(string=...)k`
After fix: `Invalid 'target' parameter, parse err: parse ://: missing protocol scheme`

## Testing

- `go build ./...` — compiles cleanly
- `go test ./exporter/ -run TestScrapeEndpointDisabled -v` — passes

### DIFF
--- a/exporter/http.go
+++ b/exporter/http.go
@@ -58,7 +58,7 @@ func (e *Exporter) scrapeHandler(w http.ResponseWriter, r *http.Request) {
 
 	u, err := url.Parse(target)
 	if err != nil {
-		http.Error(w, fmt.Sprintf("Invalid 'target' parameter, parse err: %ck ", err), http.StatusBadRequest)
+		http.Error(w, fmt.Sprintf("Invalid 'target' parameter, parse err: %v ", err), http.StatusBadRequest)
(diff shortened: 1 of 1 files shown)


oliver006:
Thanks for the PR!

I think the test failure is unrelated (and should be fixed by #1124)

PR #1128, opened 2026-06-03
Add SBOM generation to docker image release

hxyannay:
Hey :)
First of all wanted to thank you for that project!

I am using it in my environment and our security teams started requiring docker images to use SBOMs to validate CVEs, since it is very easy to change the docker build process to include a SBOM, I would highly appreciate it if you could approve it and then I will be able to keep using this in our environment.

Thank you in advance!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,6 +89,7 @@ jobs:
           target: scratch-release
           platforms: linux/amd64,linux/arm,linux/arm64
           push: true
+          sbom: true
           tags: ${{ steps.meta.outputs.tags }}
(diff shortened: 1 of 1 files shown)


oliver006 (approved):
TIL about SBOM. 

Thanks for the PR!

hxyannay:
Thanks @oliver006 !
If you can release a new version so the docker image will be release with the SBOM I would really appreciate it.
If not that's ok, I will wait for the next version.

oliver006:
https://github.com/oliver006/redis_exporter/releases/tag/v1.85.0

Issue #1131, opened 2026-06-11
Adding command log metrics


Sasidharan3094:
[...]

PR #1132, opened 2026-06-11
Adding commandlog metrics

Sasidharan3094:
This PR will solve this [issue](https://github.com/oliver006/redis_exporter/issues/1131)


Sasidharan3094:
@oliver006 Please review this PR, this is to add commandlog metrics which was introduced in valkey 8.1

oliver006:
Thanks for the PR!

oliver006:
I'm a little short on time right now but I'll try to review this in the coming days.

Sasidharan3094:
@oliver006 Sure, here is the test results which i ran in my local.

TEST_VALKEY8_URI="valkey://localhost:16382" go test -v -run TestCommandLog ./exporter/

=== RUN   TestCommandLog
=== RUN   TestCommandLog/slow
=== RUN   TestCommandLog/large-request
=== RUN   TestCommandLog/large-reply
=== RUN   TestCommandLog/reset
--- PASS: TestCommandLog (1.11s)
    --- PASS: TestCommandLog/slow (0.40s)
    --- PASS: TestCommandLog/large-request (0.18s)
    --- PASS: TestCommandLog/large-reply (0.18s)
    --- PASS: TestCommandLog/reset (0.34s)
PASS
ok      github.com/oliver006/redis_exporter/exporter    (cached)

steinberg-dev:
Nice work @Sasidharan3094! Ran this against a 6-node Valkey 9.x cluster and results look consistent — all four test cases pass with comparable timings to your local run.

One suggestion: adding CONFIG threshold metrics (commandlog-execution-slower-than etc.) would let operators correlate log volume with trigger settings. Right now the exporter shows entries exist but not *why*.

Full cluster results and a code sketch for the CONFIG suggestion on the [implementation branch](https://steinberg-dev.github.io/valkey-cluster-bench/).

Also noticed the tests use `time.Sleep` for metric synchronization — might want retry/polling for cluster mode reliability (p99 on reconnect path showed 182ms on node-4 during failover).

Sasidharan3094:
@steinberg-dev Thanks for validating on the 6-node Valkey 9 cluster — good to hear all four cases pass with similar timings.

CONFIG thresholds: Agreed this helps correlate volume with settings. Length was exported unconditionally, while thresholds only appeared with --include-config-metrics (via generic config_value{key="..."}). I’ve added dedicated commandlog threshold metrics alongside the length gauges — execution/request/reply thresholds plus the max-len settings — so they’re always available when COMMANDLOG is supported.

Test synchronization: Fair point on cluster timing. I replaced the fixed time.Sleep(50ms) with bounded polling: COMMANDLOG LEN on the server side after setup/reset, and retrying metric assertions until the expected value appears (2s timeout, 50ms interval).

Both changes are in the latest commit.

waddles (commented):
Thanks for adding this — nice to see Valkey 8 `COMMANDLOG` support, and the integration test covering all three log types plus a reset case is great. A few things before this is ready to merge:

**Needs fixing (CI blocker)**
- `gofmt` fails on `exporter/commandlog_test.go` (the `const` block isn't aligned). Please run `gofmt -w` over the changed files.
- In `collectCommandLogMetricValue`, `m.Write(got)` ignores its returned error, which `errcheck` will flag. The convention elsewhere is `_ = m.Write(got)` (or check it).

**Worth addressing**
- `extractCommandLogMetrics` runs on every scrape, so on Redis / pre-8 Valkey (where `COMMANDLOG` doesn't exist) it issues 3 always-failing `COMMANDLOG LEN` round-trips per scrape. Consider detecting support once and caching it (or gating on server version) to avoid the per-scrape overhead and debug-log noise.
- The config loop makes 6 separate `CONFIG GET` calls. `CONFIG GET` accepts glob/multiple patterns, so a single `CONFIG GET commandlog-*` would collapse this to one round-trip.

**Minor**
- The 9 new metric-description entries are inserted mid-map and aren't alphabetized like the surrounding block (cosmetic).
- `collectCommandLogMetricValue` matches metrics via `strings.Contains(m.Desc().String(), name)` — works, but an exact fqName/suffix match would be less fragile to future name overlaps.

`go build` and `go vet` pass; only the `gofmt` issue above breaks the build.

Sasidharan3094:
@waddles Thanks for the thorough review — addressed all of the above in the latest commit:

**gofmt / errcheck**: ran gofmt -w on the changed files; collectCommandLogMetricValue now uses _ = m.Write(got).
**COMMANDLOG probe**: cache support on the Exporter after a single COMMANDLOG LEN slow probe; subsequent scrapes on Redis/pre-8 Valkey skip without extra round-trips.
**CONFIG GET**: collapsed to one CONFIG GET commandlog-* call.
**Minor**: alphabetized the metric-description entries; test helper now matches on exact fqName via BuildFQName.

Sasidharan3094:
@oliver006 @waddles can you please review this

oliver006:
Thanks for all the effort everyone!
I'll try to review this by this weekend, nudge me if you don't hear from me.

exporter/commandlog.go
```diff
@@ -0,0 +1,65 @@
...
+		{"large-reply", "commandlog_large_reply_length"},
+	}
+
+	for _, t := range commandLogTypes {
+		if reply, err := redis.Int64(doRedisCmd(c, "COMMANDLOG", "LEN", t.logType)); err == nil {
+			e.registerConstMetricGauge(ch, t.metric, float64(reply))
+		}
+	}
+
+	config, err := redis.StringMap(doRedisCmd(c, "CONFIG", "GET", "commandlog-*"))
```
oliver006: this should use the configured CONFIG cmd --> `e.options.ConfigCommandName`

Sasidharan3094: updated to use e.options.ConfigCommandName instead of hardcoded "CONFIG", and skip the threshold fetch when it's set to "-".

exporter/commandlog.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (e *Exporter) extractCommandLogMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
+	if e.commandLogSupported != nil && !*e.commandLogSupported {
+		return
+	}
+
+	if e.commandLogSupported == nil {
+		_, err := redis.Int64(doRedisCmd(c, "COMMANDLOG", "LEN", "slow"))
+		supported := err == nil
+		e.commandLogSupported = &supported
```
oliver006: Let's not keep "supported" state in the exporter

i think you can just discard the check here and simplify this section.
This will also make the exporter more robust in case an earlier COMMANDLOG call failed due to permission changes or transient issues .

Try something like 

```
  for _, t := range commandLogTypes {
      reply, err := redis.Int64(doRedisCmd(c, "COMMANDLOG", "LEN", t.logType))
      if err != nil {
          continue
      }
      supported = true
      e.registerConstMetricGauge(ch, t.metric, float64(reply))
  }

  if !supported {
      return
  }
```

exporter/commandlog_test.go
```diff
@@ -0,0 +1,290 @@
...
+func setupCommandLogLargeRequest(t *testing.T, addr string) {
+	t.Helper()
+
+	c, err := redis.DialURL(addr)
+	if err != nil {
+		t.Fatalf("couldn't connect to valkey, err: %s", err)
+	}
+	defer c.Close()
+
+	if _, err = c.Do("CONFIG", "SET", "commandlog-request-larger-than", 10); err != nil {
```
oliver006, line 224: This isn't getting reset in the cleanup

exporter/commandlog_test.go
```diff
@@ -0,0 +1,290 @@
...
+func setupCommandLogLargeReply(t *testing.T, addr string) {
+	t.Helper()
+
+	c, err := redis.DialURL(addr)
+	if err != nil {
+		t.Fatalf("couldn't connect to valkey, err: %s", err)
+	}
+	defer c.Close()
+
+	if _, err = c.Do("CONFIG", "SET", "commandlog-reply-larger-than", 10); err != nil {
```
oliver006, line 250: This isn't getting reset in the cleanup

oliver006 (commented):
This is starting to look good! Two smaller issues to address and we should be good to go

exporter/commandlog.go
Sasidharan3094: Sure

exporter/commandlog_test.go
Sasidharan3094: this is taken care

exporter/commandlog_test.go
Sasidharan3094: taken care, please check

oliver006 (approved):
Nice work - thanks for seeing this through!

PR #1134, opened 2026-06-15
Redact credentials from config metrics unconditionally; redact sentinel config

waddles:
## What

Harden redaction of Redis config exported via `config_key_value` / `config_value` metrics so credentials can never be read from the metrics endpoint, and extend the same protection to Sentinel config.

## Why

Previously, credential config keys were withheld **only** when `--redact-config-metrics` was true (the default). An operator who set `--redact-config-metrics=false` (e.g. for debugging) would expose `requirepass`/`masterauth`/TLS key passphrases in plaintext on `/metrics`. Sentinel config (`sentinel_config_key_value`) was never redacted at all.

## Changes

- **Credentials are never exported**, regardless of `--redact-config-metrics`: `requirepass`, `masterauth`, `tls-key-file-pass`, `tls-client-key-file-pass`, plus a defense-in-depth backstop for any key containing `password`/`passwd`/`secret`/`token`.
- **Redacted keys are withheld entirely** (no placeholder series), matching the previous omit behavior.
- The flag now controls only **non-secret-but-sensitive** keys (usernames: `user`, `masteruser`), which are exported only when redaction is explicitly disabled for debugging.
- Keys are **normalized** (lower-cased + trimmed) before matching, so a fork returning different casing cannot bypass redaction.
- **Sentinel config** now applies the same policy via a shared `shouldRedactConfigKey` helper.
- The password-map debug log now logs the **redacted** address.
- Flag help text and README updated to document the new contract.

## Behavior changes (please review)

- Password keys are now withheld even when `--redact-config-metrics=false` (was: shown).
- `sentinel_config_*` metrics now redact credentials (was: always raw).

## Tests

- `TestShouldRedactConfigKey` — all tiers, casing/whitespace normalization, and substring backstop.
- `TestExtractConfigMetricsRedaction` — credentials never appear; usernames are flag-gated; non-secret keys stay raw.

`gofmt`, `go vet`, and `go build ./...` all clean.


exporter/exporter.go
```diff
@@ -701,6 +701,60 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
 	ch <- e.targetScrapeRequestErrors
 }
 
+// alwaysRedactedConfigKeys holds config keys whose values are credentials. They
+// are never exported as metrics, regardless of the --redact-config-metrics
+// setting, so that an attacker cannot read them from /metrics even when an
+// operator has turned redaction off.
+var alwaysRedactedConfigKeys = map[string]struct{}{
```
oliver006, line 751: Should this include `sentinel-suer`, `sentinel-pass`, and `auth-pass`?

exporter/exporter.go
```diff
@@ -701,6 +701,60 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
...
+// alwaysRedactedKeySubstrings is a defense-in-depth backstop: any config key
+// containing one of these substrings is treated as credential-bearing and is
+// never exported. This keeps secret-holding keys introduced by future Redis
+// versions or forks from leaking before they are added to the list above.
+var alwaysRedactedKeySubstrings = []string{"password", "passwd", "secret", "token"}
+
+// optionallyRedactedConfigKeys holds keys that are not secrets themselves but
+// can help an attacker (e.g. usernames). They are exported only when redaction
+// is explicitly disabled via --redact-config-metrics=false (for debugging).
+var optionallyRedactedConfigKeys = map[string]struct{}{
```
oliver006, line 770: Same. Add sentinel fields here?

exporter/redis.go
```diff
@@ -50,11 +50,11 @@ func (e *Exporter) lookupPasswordInPasswordMap(uri string) (string, bool) {
 		u.User = url.User(e.options.User)
 	}
 	uri = u.String()
-
 	// strip solo ":" if present in uri that has a username (and no pwd)
 	uri = strings.Replace(uri, fmt.Sprintf(":@%s", u.Host), fmt.Sprintf("@%s", u.Host), 1)
 
-	log.Debugf("looking up in pwd map, uri: %s", uri)
+	// log the redacted form so a password embedded in the address is not written to the logs
+	log.Debugf("looking up in pwd map, uri: %s", u.Redacted())
```
oliver006: This might not be enough, the same URI still flows into other debug logs (see e.g. line 75)
Maybe worth adding a helper that remove both user and pwd from the uri, for logging?

PR #1136, opened 2026-06-22
test: skip instead of panic when TEST_REDIS_URI is unset

waddles:
## What

`getTestExporter` / `getTestExporterWithOptions` now take `*testing.T` and call `t.Skip` when `TEST_REDIS_URI` is unset, instead of `panic`. The suite then skips cleanly on machines without a live Redis rather than crashing the package.

No change to non-test code.

## Why

Splitting out a small, mechanical, behaviour-preserving change so the two feature PRs that follow are easier to review.


oliver006 (approved):
Thanks!

PR #1137, opened 2026-06-22
feat: discover arbitrary cluster via ?target= and preserve URI scheme

waddles:
## Why

Managed clusters such as AWS ElastiCache change topology underneath you — failover, scaling and node replacement all churn node hostnames, and AWS gives you no control over those names. The node set therefore cannot be enumerated statically; it has to be discovered at scrape time.

## What

- `/discover-cluster-nodes` now accepts a `?target=<uri>` parameter to discover an arbitrary cluster, rather than only the exporter's own `--is-cluster` instance.
- The discovery output preserves the original URI scheme (`redis` / `rediss` / `valkey` / `valkeys`) so TLS and Valkey clusters keep emitting matching node URIs (previously it was a binary `redis`/`rediss` choice).

### Changes
- `redis.go`: add `schemeFromURI` / `schemeIsTLS` / `startupNodeFromURI` helpers; `configureOptions` takes an explicit `useTLS` arg; split `connectToRedisCluster` into `connectToRedisClusterWithURI`.
- `http.go`: `discoverClusterNodesHandler` honours `?target=` and emits scheme-preserving node URIs.
- `docker-compose` / `Makefile` / `gen-test-certs.sh`: add an authenticated TLS valkey cluster (and image bumps) for integration testing.

## Example: discovery against an ElastiCache cluster

> Illustrative — hostnames are simulated, but the JSON shape is exactly what the
> handler emits (verified by the integration tests against the compose clusters).

The discovery `target` is **always the `clustercfg` endpoint**. In ElastiCache `clustercfg.*` is not a real node — it's the configuration endpoint AWS provides purely to fetch cluster topology. You point discovery at that one stable name and it returns the actual scrapeable nodes:

```
GET /discover-cluster-nodes?target=rediss://clustercfg.mycache.4xk2pq.use1.cache.amazonaws.com:6379
```

**During a Friday-frenzy sale — scaled up to 2 shards × 5 nodes (10 nodes):**

```json
[
  {
    "targets": [
      "rediss://mycache-0001-001.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0001-002.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0001-003.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0001-004.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0001-005.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0002-001.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0002-002.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0002-003.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0002-004.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0002-005.mycache.4xk2pq.use1.cache.amazonaws.com:6379"
    ],
    "labels": {}
  }
]
```

**After the sale — scaled back to 2 shards × 2 nodes (4 nodes), same `target`:**

```json
[
  {
    "targets": [
      "rediss://mycache-0001-001.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0001-003.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0002-002.mycache.4xk2pq.use1.cache.amazonaws.com:6379",
      "rediss://mycache-0002-005.mycache.4xk2pq.use1.cache.amazonaws.com:6379"
    ],
    "labels": {}
  }
]
```

Note the survivors are `0001-001`, `0001-003`, `0002-002`, `0002-005` — ElastiCache does **not** keep a tidy contiguous `001`/`002`; which node ordinals remain is arbitrary. Every hostname in the list can change between scrapes; only the `clustercfg` target stays constant. That's the whole point — Prometheus refreshes the target list from one stable endpoint and always scrapes whatever nodes currently exist. (Node hostnames are surfaced with `--cluster-discover-hostnames`.)

## Testing

Integration tests run against the compose clusters:
- discovery with `?target=` (redis cluster + password-protected valkey cluster)
- TLS scheme preservation (`rediss://` and `valkeys://`)
- the `NOAUTH` failure path

All pass under `go test -race`.

## Note

When discovery returns IP addresses (rather than hostnames), scraping those nodes over TLS validates the cert against the IP; use the `tls_server_name` `/scrape` parameter (`__param_tls_server_name`) when a node's certificate name differs from its address.


oliver006:
Code looks good but can you please add a flag to enable the `target` functionality so that by default we don't add an internal TCP probing mechanism to the exporter (without turning on the flag)?

waddles:
> Code looks good but can you please add a flag to enable the `target` functionality so that by default we don't add an internal TCP probing mechanism to the exporter (without turning on the flag)?

Fair call, a remote attacker could possibly use it as a reflection service. I have put cluster discovery targets behind an allowlist so operators need to explicitly opt-in to discovery of their clusters.

main.go
```diff
@@ -190,6 +190,7 @@ func main() {
 		isTile38                       = flag.Bool("is-tile38", getEnvBool("REDIS_EXPORTER_IS_TILE38", false), "Whether to scrape Tile38 specific metrics")
 		isCluster                      = flag.Bool("is-cluster", getEnvBool("REDIS_EXPORTER_IS_CLUSTER", false), "Whether this is a redis cluster (Enable this if you need to fetch key level data on a Redis Cluster).")
 		clusterDiscoverHostnames       = flag.Bool("cluster-discover-hostnames", getEnvBool("REDIS_EXPORTER_CLUSTER_DISCOVER_HOSTNAMES", false), "Whether to use hostname for cluster node discovery if available via `/discover-cluster-nodes` endpoint.")
+		clusterDiscoverTargetAllowlist = flag.String("cluster-discover-target-allowlist", getEnv("REDIS_EXPORTER_CLUSTER_DISCOVER_TARGET_ALLOWLIST", ""), "Comma-separated list of host globs allowed as the `target` parameter of `/discover-cluster-nodes`. Empty (default) disables target-based discovery so the exporter will not connect to caller-supplied addresses.")
```
oliver006, line 193: This is more complicated / advanced than I think this needs to be. A simple bool flag should suffice to allow using the target param or not?
wdyt?

exporter/redis.go
```diff
@@ -11,7 +11,39 @@ import (
...
+	if err != nil {
+		return "", fmt.Errorf("failed to parse cluster URI: %w", err)
+	}
+	if u.Port() == "" {
+		return u.Host + ":6379", nil
+	}
+	return u.Host, nil
+}
+
+func (e *Exporter) configureOptions(uri string, useTLS bool) ([]redis.DialOption, error) {
```
oliver006: why change the signature and not move the schemeIsTls() call here, to line 25?

main.go
waddles: I could have just kept it behind the same `--is-cluster` flag but I think that's a poorly named flag. What is your actual concern about enabling targets for `/discover-cluster-nodes`? I thought more about it and realised there is no restriction on targets used on the `/scrape` endpoint - you can scrape any host:port and the exporter will open an outbound TCP connection to it. The only thing stopping it is if the administrator has enabled basic-auth on the exporter.

Either way, I think a boolean or an allow list is scope creep that doesn't really add any value so I want to understand your motivation clearly.

exporter/redis.go
waddles: Good call — dropped the parameter. Both callers were passing `schemeIsTLS(uri)` derived from the same `uri`, so `configureOptions` now computes it internally via `redis.DialUseTLS(schemeIsTLS(uri))`. Done in 5565bee.

steinberg-dev (commented):
Really solid implementation. The ElastiCache use case is well-argued, and the scheme-preserving output is the right design — avoids the binary redis/rediss assumption that bit Valkey users.

A few observations from reviewing:

**`schemeFromURI` default return**

The default return of `"redis"` is fine for backward compatibility, but might silently mask a misconfigured URI (e.g., bare `host:6379` without a scheme prefix). Wondering if it is worth a debug log when the scheme is unrecognized, similar to how `doRedisCmd` logs at Debug level. Not blocking, just a thought for operator visibility.

**`connectToRedisClusterWithURI` error context**

The error wrapping is clean, but the returned errors do not always include the URI — if `startupNodeFromURI` fails, the caller sees `"invalid URI: ..."` which is clear, but if the cluster connection itself fails after startup node resolution, the error loses which URI was being dialed. A minor thing, but when you are debugging a fleet with 20+ `?target=` endpoints, having the URI in the error message saves a grep roundtrip.

**Tests**

Integration tests cover TLS + password-protected valkey cluster — exactly the paths that break in production. The `gen-test-certs.sh` addition is appreciated. One thing I would flag: the test for the `?target=` path verifies the happy path and the NOAUTH path, but does not verify what happens when the target points to a standalone (non-cluster) Redis instance. `CLUSTER SLOTS` against a standalone Redis returns `ERR This instance has cluster support disabled` — is there value in an integration test for that error path, or is the current unit-level error handling sufficient?

Overall this is a well-thought-out PR. The scheme helpers (`schemeFromURI`, `startupNodeFromURI`) are clean extractions that would be useful outside this handler too.

PR #1138, opened 2026-06-22
feat: cache credentials for discovered cluster nodes

waddles:
> **Stacks on #1137** — review/merge that first. Until it merges, the diff here also shows #1137's commit.

## Why

Following on from #1137: for managed clusters such as AWS ElastiCache the node hostnames cannot be enumerated ahead of time — they churn with failover, scaling and replacement, and aren't controllable. So you can't list the individual nodes (and their credentials) in a password file in advance; you only know the stable `clustercfg` discovery endpoint.

## What

The key idea: **the same authentication used to discover the nodes is propagated to every discovered node**, so the scraper never needs per-node credentials configured.

1. Discovery connects to the `clustercfg` target using the credentials already configured for it (the password map / `--redis.password`).
2. On success, that **same password is propagated into `discoveredNodesPasswordCache`, keyed per discovered node URI** — every node inherits the credentials that worked for the cluster.
3. When Prometheus / vmagent later scrapes each discovered node via `/scrape`, the cached credential is found and applied automatically, so **all discovered instances already have credentials for authentication** — no extra configuration, even as the node set churns.

A pwd-file reload (`/-/reload`) invalidates the cache so rotated credentials are not served from stale entries.

### Changes
- `exporter.go`: add `discoveredNodesPasswordCache` (+ `passwordUpdateMutex`).
- `redis.go`: extract `canonicalPasswordKey`; `lookupPasswordInPasswordMap` falls back to the cache.
- `http.go`: propagate the discovery target's password to each discovered node; inject the cached password in `scrapeHandler` (the per-scrape exporter is a fresh instance with an empty cache, so the password is read from the parent and passed through); invalidate the cache on reload.

## Concurrency

`passwordUpdateMutex` guards `options.PasswordMap` and the cache against a concurrent `/-/reload`. It is deliberately **not** the embedded `sync.Mutex` — `Collect` holds that for the duration of a scrape, and the scrape path reaches the password lookup, so reusing it would deadlock. `scrapeHandler`'s `opts := e.options` copy is also taken under this lock so it can't race the reload's `PasswordMap` reassignment.

Two race regression tests (pass under `go test -race`, no live Redis needed):
- `TestPasswordLookupConcurrentWithReload` — lookup vs. discovery-cache write vs. reload.
- `TestScrapeHandlerConcurrentWithReload` — drives `/scrape` and `/-/reload` concurrently.

## Testing

Integration tests against the compose clusters (all under `-race`):
- `TestScrapeDiscoveredNodeWithCachedPassword` — discover, then scrape a discovered node of a password-protected cluster using the propagated credentials.
- `TestScrapeDiscoveredClusterNodesTLS` — discover, then scrape every node over TLS (`rediss://` and `valkeys://`).


oliver006:
I still like dislike keep the credentials persisted in the exporter and would like to see what we can do about that.

Can you return the credentials as part of the discover-nodes response? As in, make the discover endpoint return URLs that include username/password?

waddles:
> I still like dislike keep the credentials persisted in the exporter and would like to see what we can do about that.
> 
> Can you return the credentials as part of the discover-nodes response? As in, make the discover endpoint return URLs that include username/password?

My aim with this was to keep the change minimal - it doesn't alter how cluster authentication works or the shape of the discovery response. The credentials the operator already persists after loading via the password file in [LoadPwdFile](https://github.com/oliver006/redis_exporter/blob/5ffc8e2d3e8c77cec6176ca2bd6a81e1b8882e62/exporter/pwd_file.go#L11) just get reused for the discovered nodes, and the response stays as plain target URLs.

Putting credentials in the response would move them out of the exporter and onto the wire, replicating them into every scraper's service-discovery state, targets view, logs and TSDB. The password file is designed to keep credentials in one place and out of band, never allowing them to be exposed.

I did consider returning metadata labels and looking the password up from those at scrape time, but that requires maintaining a map of metadata back to the credential to use. It also pushes a lot onto the user: relabel configs are one of the hardest parts of Prometheus to get right, and this would force every operator to write and debug their own just to wire credentials through. That's a real source of friction and support burden, and it's easy to get subtly wrong. The cache avoids all of it - the discovered node address is already the lookup key, so there is nothing for the operator to configure.

EDIT: this branch is rebased on top of #1137 so it may be harder to review. If you want to see the changes in isolation compare the branches on [my fork](https://github.com/waddles/redis_exporter/compare/feat/discover-cluster-nodes-target-param...feat/cache-discovered-node-credentials)

PR #1143, opened 2026-07-09
fix: prevent exporter crash on invalid RediSearch index regex

destinyoooo:
## Changes

- Replace `regexp.MustCompile(e.options.CheckSearchIndexes)` with `regexp.Compile`
- Log and skip RediSearch index metrics when `--check-search-indexes` contains an invalid regex
- Add a regression test covering invalid regex input without panic

### DIFF
--- a/exporter/search_indexes.go
+++ b/exporter/search_indexes.go
@@ -47,11 +47,7 @@ func (e *Exporter) extractSearchIndexesMetrics(ch chan<- prometheus.Metric, c re
 	}
 
 	// Get indexes list based on check-search-indexes regex
-	checkIndexRegex, err := regexp.Compile(e.options.CheckSearchIndexes)
-	if err != nil {
-		log.Errorf("invalid check-search-indexes regex %q: %s", e.options.CheckSearchIndexes, err)
-		return
-	}
+	checkIndexRegex := regexp.MustCompile(e.options.CheckSearchIndexes)
 	for _, index := range allSearchIndexes {
 		if checkIndexRegex.MatchString(index) {
 			searchIndexes = append(searchIndexes, index)
--- a/exporter/search_indexes.go
+++ b/exporter/search_indexes.go
@@ -47,7 +47,11 @@ func (e *Exporter) extractSearchIndexesMetrics(ch chan<- prometheus.Metric, c re
 	}
 
 	// Get indexes list based on check-search-indexes regex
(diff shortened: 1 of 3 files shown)


exporter/search_indexes.go
```diff
@@ -47,7 +47,11 @@ func (e *Exporter) extractSearchIndexesMetrics(ch chan<- prometheus.Metric, c re
 	}
 
 	// Get indexes list based on check-search-indexes regex
-	checkIndexRegex := regexp.MustCompile(e.options.CheckSearchIndexes)
+	checkIndexRegex, err := regexp.Compile(e.options.CheckSearchIndexes)
```
oliver006: Interesting change, I like it.

considering that `CheckSearchIndexes` is provided at startup, instead of errorring out down here, how about just fail at startup?

destinyoooo: Good point, thanks. I moved the regex validation to NewRedisExporter so invalid check-search-indexes fails at startup. The scrape path stays unchanged and can keep using MustCompile because the config is already validated.

exporter/search_indexes_test.go
```diff
@@ -109,3 +109,14 @@ func TestExtractSearchIndexesMetrics(t *testing.T) {
...
 }
+
+func TestNewRedisExporterInvalidSearchIndexRegex(t *testing.T) {
+	_, err := NewRedisExporter("", Options{
+		Namespace:                "test",
+		InclSearchIndexesMetrics: true,
+		CheckSearchIndexes:       "[",
+	})
+	if err == nil {
+		t.Fatal("expected invalid check-search-indexes regex to fail at startup")
```
oliver006: Can you add the positive test case as well please?

oliver006 (commented):
Looks good, let's just add the other test case as well and we should be good to go.

exporter/search_indexes_test.go
destinyoooo: OK, I've finished it.

oliver006: Please use a table-driven test for this instead of adding another test case.
You can look at other tests for what the pattern looks like

destinyoooo: Updated. I combined the valid and invalid regex cases into a table-driven test following the existing test patterns.

oliver006 (approved):
Thank you, looks good

PR #1151, opened 2026-07-15
ensure a change in label keys doesn't vanish the metric

snubba:
## Problem

After upgrading the monitored Valkey server (e.g. 8.0 → 8.1), the `redis_instance_info` metric disappears from `/metrics` and only comes back after the exporter itself is restarted.

## Root cause
`redis_instance_info` has a **dynamic label set**. In `exporter/info.go` the labels `valkey_version` and  valkey_release_stage` are only added when the server's `INFO` output reports those fields:

```go
if valkeyVersion, ok := keyValues["valkey_version"]; ok { ... }
if valkeyReleaseStage, ok := keyValues["valkey_release_stage"]; ok { ... }
```

`createMetricDescription()` cached the `*prometheus.Desc` by metric name on the first scrape and returned it unchanged forever after, ignoring the `labels` argument on subsequent calls. The description map is built once at construction and lives for the whole process lifetime.

When a Valkey upgrade e.g. adds a new `INFO` field (Valkey 8.1 introduced `valkey_release_stage`), the scrape passes more label values than the cached description has labels. `prometheus.NewConstMetric` then returns a label/value count-mismatch error that is only logged at debug level, and the metric is silently dropped — until the exporter is restarted and rebuilds the description.

## Fix
`createMetricDescription()` now rebuilds the cached description when the explicitly supplied label set differs from the one the cached description was built with (tracked in a new `metricDescriptionLabels` map).

Without this scoping, calling a labeled metric with too few values (e.g. `sentinel_master_ok_sentinels` when sentinel labels are absent) would rebuild it with an empty label set and corrupt the schema.

### DIFF
--- a/exporter/metrics.go
+++ b/exporter/metrics.go
@@ -3,6 +3,7 @@ package exporter
 import (
 	"fmt"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR and for finding this issue - I'll try to review this in the coming days and leave feedback.

exporter/metrics.go
```diff
@@ -152,13 +153,24 @@ func (e *Exporter) mustFindMetricDescription(metricName string) *prometheus.Desc
 }
 
 func (e *Exporter) createMetricDescription(metricName string, labels []string) *prometheus.Desc {
-	if desc, found := e.metricDescriptions[metricName]; found {
-		return desc
-	}
+	explicitLabels := len(labels) > 0
```
oliver006: let's fold this into line 164 and then we should be good to go

oliver006 (commented):
Thanks for the PR!

exporter/metrics.go
```diff
@@ -152,13 +153,24 @@ func (e *Exporter) mustFindMetricDescription(metricName string) *prometheus.Desc
...
+
+	if desc, found := e.metricDescriptions[metricName]; found {
+		// rebuild when a dynamic-label metric (e.g. instance_info) changes its labels at runtime, e.g. a Valkey upgrade adding new INFO fields
+		if !explicitLabels || slices.Equal(e.metricDescriptionLabels[metricName], labels) {
+			return desc
+		}
+	}
+
+	if e.metricDescriptionLabels == nil {
+		e.metricDescriptionLabels = map[string][]string{}
```
oliver006, line 168: This one isn't covered by tests so let's add a test case that triggers this plz

https://coveralls.io/builds/80611704/source?filename=exporter%2Fmetrics.go#L170

oliver006 (approved):
Looks good, thank you!

PR #1156, opened 2026-07-24
fix: avoid COMMANDLOG error spam on targets without COMMANDLOG support

### LINKED ISSUE #1155: ERR unknown command 'COMMANDLOG'
**Describe the problem**
redis_exporter versions >= v1.87.0 (introduced in [PR#1132](https://github.com/oliver006/redis_exporter/pull/1132)) issue the following commands on every scrape:
- `COMMANDLOG LEN slow`
- `COMMANDLOG LEN large-request`
- `COMMANDLOG LEN large-reply`

Causing the following error on Bitnami Redis Cluster image (tested on [v8.8.1-debian-12-r0](https://github.com/bitnami/containers/tree/main/bitnami/redis-cluster)):

- `c.Do() - err: ERR unknown command 'COMMANDLOG', with args beginning with: 'LEN' 'slow'`
- `c.Do() - err: ERR unknown command 'COMMANDLOG', with args beginning with: 'LEN' 'large-request'`
- `c.Do() - err: ERR unknown command 'COMMANDLOG', with args beginning with: 'LEN' 'large-reply'`

This creates a periodic false-positive error signal in monitoring and alerting (_total_error_replies_ metric counter increases).

**Running the exporter**
We use [Bitnami Helm Chart redis-exporter](https://github.com/bitnami/charts/tree/main/bitnami/redis-cluster). It creates a side-car container with:

```yaml
command:
  - /bin/bash
  - '-c'
  - |
    export REDIS_PASSWORD="$(< "${REDIS_PASSWORD_FILE}")"
    redis_exporter
env:
  - name: BITNAMI_DEBUG
    value: 'false'
  - name: REDIS_ALIAS
    value: redis-cluster
  - name: REDIS_ADDR
    value: 'redis://127.0.0.1:6379'
  - name: REDIS_PASSWORD_FILE
    value: /opt/bitnami/redis/secrets/redis-password
  - name: REDIS_EXPORTER_WEB_LISTEN_ADDRESS
    value: ':9121'
```

**Expected behavior**
- Exporter should not generate persistent unknown-command errors against Redis targets that do not support COMMANDLOG.
- total_error_replies should reflect real client/application errors, not exporter capability probing side effects.

Chasnechok:
Closes #1155

### DIFF
--- a/exporter/commandlog.go
+++ b/exporter/commandlog.go
@@ -19,7 +19,22 @@ var commandLogConfigMetrics = []struct {
 	{"commandlog-large-reply-max-len", "commandlog_large_reply_max_len"},
 }
 
+func (e *Exporter) isCommandLogSupported(c redis.Conn) bool {
+	if e.commandLogSupported != nil {
+		return *e.commandLogSupported
+	}
+
+	reply, err := redis.Values(doRedisCmd(c, "COMMAND", "INFO", "COMMANDLOG"))
+	supported := err == nil && len(reply) > 0 && reply[0] != nil
+	e.commandLogSupported = &supported
+	return supported
+}
+
 func (e *Exporter) extractCommandLogMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
+	if !e.isCommandLogSupported(c) {
+		return
+	}
+
 	commandLogTypes := []struct {
 		logType string
 		metric  string
@@ -29,18 +44,12 @@ func (e *Exporter) extractCommandLogMetrics(ch chan<- prometheus.Metric, c redis
 		{"large-reply", "commandlog_large_reply_length"},
 	}
 
-	supported := false
 	for _, t := range commandLogTypes {
 		reply, err := redis.Int64(doRedisCmd(c, "COMMANDLOG", "LEN", t.logType))
 		if err != nil {
 			continue
 		}
-		supported = true
 		e.registerConstMetricGauge(ch, t.metric, float64(reply))
-	}
-
-	if !supported {
-		return
 	}
 
 	if e.options.ConfigCommandName == "-" {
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR - I'll try to review it as soon as possible.

exporter/exporter.go
```diff
@@ -49,6 +49,8 @@ type Exporter struct {
 	mux *http.ServeMux
 
 	buildInfo BuildInfo
+
+	commandLogSupported *bool
```
oliver006: Please don't cache the state here, it causes issues when e.g. the redis instance restarts and the state of exporter gets out of sync with the redis instance (there was an issue about it recently with metrics labels)
just probe for `command log` each time and don't return an error if it doesn't exist.

steinberg-dev (commented):
Clean fix. The `COMMAND INFO` pre-check with the cached `commandLogSupported` bool is the right approach — avoids the debug log noise on every scrape without touching the shared `doRedisCmd` path. The test covering both Redis 7 and 8 with the error reply assertion is thorough.

Ran into this same issue on a Bitnami cluster while testing the Valkey 9.1 changes. Good to see it fixed.

exporter/commandlog.go
```diff
@@ -19,7 +19,16 @@ var commandLogConfigMetrics = []struct {
 	{"commandlog-large-reply-max-len", "commandlog_large_reply_max_len"},
 }
 
+func isCommandLogSupported(c redis.Conn) bool {
+	reply, err := redis.Values(doRedisCmd(c, "COMMAND", "INFO", "COMMANDLOG"))
+	return err == nil && len(reply) > 0 && reply[0] != nil
+}
+
 func (e *Exporter) extractCommandLogMetrics(ch chan<- prometheus.Metric, c redis.Conn) {
+	if !isCommandLogSupported(c) {
```
oliver006: Let's skip tile38 as well (if e.options.IsTile38), we know that one doesn't support command log

oliver006 (commented):
one more small thing, rest looks good!

oliver006 (approved):
Looks good - thanks for fixing this!

Issue #478, opened 2021-01-20
redis exporter with ssl `rediss://` does not work


sangeethdba:
**Describe the problem**
redis exporter with ssl `rediss://` does not work

 /opt/isv/redis/redis_exporter/redis_exporter-v1.14.0.linux-amd64/redis_exporter \ 
-redis.addr rediss://xx.xx.xx.xx:9002 -web.listen-address 0.0.0.0:9132 \ -tls-server-key-file /opt/isv/redis/certfiles/newqpssc_certs/private.key \
 -tls-server-cert-file /opt/isv/redis/certfiles/newqpssc_certs/com-redis-cluster-qpssc.cer \
 -tls-ca-cert-file /opt/isv/redis/certfiles/newqpssc_certs/rootca.crt \
 -debug

INFO[0000] Redis Metrics Exporter v1.14.0    build date: 2020-12-10-01:08:14    sha1: fcb9a8fdf63125ee43a40cf6a6dae2e37414c2b0    Go: go1.15.6    GOOS: linux    GOARCH: amd64
INFO[0000] Providing metrics at 0.0.0.0:9132/metrics
ERRO[0003] Couldn't connect to redis instance

**What version of redis_exporter are you running?**
Please run `redis_exporter --version` if you're not sure what version you're running.
[ ] 0.3x.x
[ ] 1.x.x

**Running the exporter**
What's the full command you're using to run the exporter? (please remove passwords and other sensitive data)

**Expected behavior**
What metrics are missing? What metrics are wrong? Is something missing that was present in an earlier version? 
Did you upgrade from 0.3x.x to 1.0 and are scraping multiple hosts? [Have a look here ](https://github.com/oliver006/redis_exporter#prometheus-configuration-to-scrape-multiple-redis-hosts) how the configuration changed.

**Screenshots**
INFO[0000] Redis Metrics Exporter v1.14.0    build date: 2020-12-10-01:08:14    sha1: fcb9a8fdf63125ee43a40cf6a6dae2e37414c2b0    Go: go1.15.6    GOOS: linux    GOARCH: amd64
INFO[0000] Providing metrics at 0.0.0.0:9132/metrics
ERRO[0003] Couldn't connect to redis instance

**Additional context**

oliver006:
Can you please run the exporter with the --debug flag to get more verbose output and post it here?
(Make sure to first remove any sensitive data like passwords or private URIs from the logs)

sangeethdba:
[...]

Issue #362, opened 2020-03-16
failed to scrape metrics if ssl is enabled


cfbytes:
My redis cluster is setup with ssl enabled, I can use my cert and key to collect to redis server from cli. 
But when starting with redis_exporter, it failed scrape metrics.

Starting log:
./redis_exporter -tls-client-cert-file=/opt/xxxx.com.cert.pem -tls-client-key-file=/opt/xxx.key -redis-only-metrics=true -redis.addr=xxxx:6379 -debug=true -skip-tls-verification=false
INFO[0000] Redis Metrics Exporter v1.3.5    build date: 2019-12-16-18:43:41    sha1: 14dda66e724e45935782db610aca803594107ff0    Go: go1.13.5    GOOS: linux    GOARCH: amd64
DEBU[0000] Enabling debug output
DEBU[0000] keys: []main.dbKeyPair(nil)
DEBU[0000] singleKeys: []main.dbKeyPair(nil)
INFO[0000] Providing metrics at :9121/metrics
DEBU[0000] Configured redis addr: "xxxx:6379"
DEBU[0044] Trying DialURL(): redis://xxxx:6379
DEBU[0044] connected to: xxxx:6379

Then try to curl the metrics endpoint with: curl http://127.0.0.1:9121/metrics, get below error log

DEBU[0044] c.Do() - running command: CLIENT [SETNAME redis_exporter]
DEBU[0044] c.Do() - err: read tcp xxxx:53061->xxxx:6379: read: connection reset by peer
DEBU[0044] c.Do() - done
ERRO[0044] Couldn't set client name, err: read tcp xxxx:53061->xxxx:6379: read: connection reset by peer
DEBU[0044] c.Do() - running command: CONFIG [GET *]
DEBU[0044] c.Do() - err: write tcp xxxx:53061->xxxx:6379: use of closed network connection
DEBU[0044] c.Do() - done
DEBU[0044] Redis CONFIG err: write tcp xxxx:53061->xxxxx:6379: use of closed network connection
DEBU[0044] c.Do() - running command: INFO [ALL]
DEBU[0044] c.Do() - err: write tcp xxxxx:53061->xxxxx:6379: use of closed network connection
DEBU[0044] c.Do() - done
DEBU[0044] c.Do() - running command: INFO []
DEBU[0044] c.Do() - err: write tcp xxxxx:53061->xxxx:6379: use of closed network connection
DEBU[0044] c.Do() - done
[...]

PR #1179, opened 2026-09-04
fix: connectToRedis() fallback dials the URL scheme instead of tcp

bashirii:
## The bug

When `DialURL()` fails for any reason (bad credentials, connection reset, etc.), `connectToRedis()`'s fallback splits the target address on `"://"` and passes the scheme itself — e.g. `"rediss"` — as the dial **network type**:

```go
c, err = redis.Dial(frags[0], frags[1], options...)
```

Go's `net` package has no `"rediss"` (or `"redis"`) network, so this always fails immediately with a misleading `unknown network rediss` error — instead of surfacing whatever `DialURL()`'s real error was.

This isn't cosmetic: it means the exporter can never report a real reason when a `rediss://` target fails to authenticate. A genuine `WRONGPASS`/`NOAUTH` gets hidden behind "unknown network", which reads as a library/scheme problem rather than what it actually is — a credential problem. `TestPasswordInvalid` had (unintentionally) encoded this masked error as the expected behavior.

## The fix

TLS is already applied via the `DialUseTLS()` option in `configureOptions()`, independent of this fallback — the network type here should always be `"tcp"`, same as the no-scheme branch just below it. Also switched from a raw string split to parsing the address via `url.Host`, which additionally strips any embedded `user:pass@` — a raw split left that attached to the address, which broke the fallback in a different way (`too many colons in address`) once the scheme bug itself was fixed. Caught this via `TestPasswordInvalid`, which targets exactly that kind of address.

Before → after, same failure:

```
# before
dial rediss: unknown network rediss

# after
NOAUTH Authentication required.
```

## Testing

- `golangci-lint v2.11.2 --tests=false`: 0 issues
- `go vet ./...` + `gofmt -l`: clean
- Full `make test` against all real fixtures (fresh containers): 0 failures
- `go test -race`: clean
- Added `TestConnectToRedisFallbackDialsTCPNotScheme`, self-contained   (uses the existing tiny TLS test-server helper, no docker-compose changes) — exercises a `DialURL()` failure on a `rediss://` target and asserts the fallback reaches the server instead of dying on `"unknown network"`.
- Updated `TestPasswordInvalid`'s expectation to the real `NOAUTH` error it now surfaces, instead of the old masked one.
- Verified against a real production Redis Cluster (TLS + auth) as well as a real plain, password-protected instance — both now report the actual underlying error instead of "unknown network".

Related: #362, #478, #387 (same underlying symptom reported independently).

### DIFF
--- a/exporter/redis.go
+++ b/exporter/redis.go
@@ -76,13 +76,15 @@ func (e *Exporter) connectToRedis() (redis.Conn, error) {
 	c, err := redis.DialURL(uri, options...)
 	if err != nil {
 		log.Debugf("DialURL() failed, err: %s", err)
-		if frags := strings.Split(e.redisAddr, "://"); len(frags) == 2 {
-			log.Debugf("Trying: Dial(): %s %s", frags[0], frags[1])
-			c, err = redis.Dial(frags[0], frags[1], options...)
-		} else {
-			log.Debugf("Trying: Dial(): tcp %s", e.redisAddr)
-			c, err = redis.Dial("tcp", e.redisAddr, options...)
+		// Always dial "tcp" here — TLS comes from the DialUseTLS option,
+		// not the network type. Parse via url.Host (not a plain string
+		// split) so any embedded user:pass@ is stripped too.
+		addr := e.redisAddr
+		if u, parseErr := url.Parse(uri); parseErr == nil && u.Host != "" {
+			addr = u.Host
 		}
+		log.Debugf("Trying: Dial(): tcp %s", addr)
+		c, err = redis.Dial("tcp", addr, options...)
 	}
 	return c, err
 }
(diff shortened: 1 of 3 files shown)


oliver006:
Thanks for the PR

I opened #1180 to add a test for connecting via unix socket so we have better coverage for your PR
My llm also included a fix for `connectToRedis()` which makes sense to me but I still have to look at your TLS-related fix.

bashirii:
Thanks for the quick look — looks like #1180 already covers this exact issue, and more completely (it also handles the unix-socket case, which my tcp-always fix would have broken). Closing this in favor of that one.

oliver006:
Thanks - wouldn't have found & fixed the issue without you opening the PR!

bashirii:
You're very welcome

PR #1189, opened 2026-09-21
check-streams: only SCAN for keys of type stream

northys:
`-check-streams` expands glob patterns with `SCAN … MATCH <pattern>` and then runs `XINFO STREAM` on every match. When the pattern also matches keys of other types (for example `db1=*` on a db that keeps hashes next to a few streams), every such key costs one extra round trip per scrape and logs

```
level=error msg="couldn't get info for stream '<key>', err: WRONGTYPE Operation against a key holding the wrong kind of value"
```

on every scrape. On a db with ~100k hashes and a handful of streams that was ~2,700 error lines per second and about 45 % of all commands the Redis server processed (Redis 7.0; v1.66.0 and v1.91.1 behave the same).

This passes `TYPE stream` to the SCAN behind `-check-streams` (`SCAN <cursor> MATCH <pattern> COUNT <n> TYPE stream`, available since Redis 6.0), so only streams are expanded and `XINFO STREAM` never hits a key of another type. Servers without the option (Redis 5) answer `ERR syntax error`; in that case the scan falls back to the previous behaviour. `-check-keys` is unchanged.

Tests: `TestScanKeysWithType` (filter on `TEST_REDIS_URI`, fallback on `TEST_REDIS5_URI`) and `TestStreamsExtractStreamMetricsSkipsNonStreamKeys`. README and flag help updated.


oliver006 (approved):
Looks good!

oliver006:
Thanks for the PR!

